PR #26, opened 2015-09-03
Build the actual list of ranks to receive work from in dcp2

awellington:
The list of ranks sending us work is not necessarily sequential, treating it as such leads to hung processes awaiting messages that will never be sent. This is especially prevalent as the number of ranks increases.


adammoody:
Thanks, Andrew!  Nice catch.

PR #41, opened 2017-02-10
Sikich1/umask

dsikich:
Added support for umask in dchmod. So, for things like -r, +r, then it will take into account the umask to set the permissions.

### DIFF
--- a/src/dchmod/dchmod.c
+++ b/src/dchmod/dchmod.c
@@ -718,10 +718,12 @@ static void set_symbolic_bits(const struct perms* p, mfu_filetype type, mode_t* 
         mode_t old_mask = umask(S_IWGRP | S_IWOTH);
         umask(old_mask);
 
+        /* save the mode after parsing before masking off bits */
+        mode_t assume_all_mode = *mode;
+
         /* only change permission bits that umask allows */
         *mode &= ~old_mask;
 
-        /* handle mode different if it is a +r,-r, etc */
         if (p->plus) {
             /* now OR this with the old mode to not turn off any
             * bits that were previously turn on */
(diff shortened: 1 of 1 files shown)


src/dchmod/dchmod.c
```diff
@@ -708,6 +711,29 @@ static void set_symbolic_bits(const struct perms* p, mfu_filetype type, mode_t*
...
+        /* handle mode different if it is a +r,-r, etc */
+        if (p->plus) {
+            /* now OR this with the old mode to not turn off any
+            * bits that were previously turn on */
+            *mode |= old_mode;
+        } else {
+            /* in the case that it was a -r,-w, etc we have to
+             * turn the bits back on that shouldn't be affected
+             * by umask */
+            *mode |= old_mask;
```
adammoody: Something may be off here.  If I have an existing file with rwxr-xr-x, and then the user does a "-w" with a umask of 077, I think the code here would end up leaving the file with something like r-xrwxrwx when it should be r-xr-xr-x.  Is that right?

I think there could be some similar problem cases with line 728, too.

I think we may need another mask operation in here on old_mode with old_mask somehow.

adammoody:
I haven't fully thought through you're latest push, but I think we may be able to drop the if/else case with something like the following:

/* get set of bits in current mode that we won't change because of umask */
old_mode &= old_mask;

/* mask out any bits of new mode that we shouldn't change due to umask */
*mode &= ~old_mask;

/* merge in bits from previous mode that had been masked */
*mode |= old_mode;

Can you try this and see if that works?

adammoody:
As one more change, let's call umask once in main() or some place like that and store the mask value in a global variable.  That way we don't repeatedly call it for every file.

PR #42, opened 2017-02-15
dcmp --sync option

dsikich:
1. Added a sync option to dcmp.c
2. Reused some code to create a dcmp_only_dst function in order to check which files were only in the dst list and copy them to the remove list. 

* note: for 2 I did that because in dcmp_strmap_compare it is looping over only the src map, and checking for files only in the src directory. So, in order to get the files that need to be deleted in the dst list I looped over the dst map. I reused some code in dcmp_strmap_check_dst, but I moved some of it to another function because it did a lot more than I thought I needed. Do you think there is a better way to get which files need to be removed?


adammoody:
@dsikich This looks good to me.  I was surprised to see that the strmap_compare only loops over the source map, too.  I didn't realize that.  I guess it must not report the number of files in the destination that don't exist in the source?

dsikich:
@adammoody Okay, thank you. I was looking through it again in totalview, and it does find the number of files in the destination that don't exist in the source, but it does so after the call to dcmp_strmap_compare in main. It is on the call to dcmp_outputs_write that it finds this info, and prints it to the console. It is in dcmp_output_write. I could move all of the remove/sync logic to after the dcmp_stramp_compare function completes?

adammoody:
There are two things here I'm thinking about.

First, in the case that we hit an error when trying to compare the content of two files, we return -1.  We then use that return value in the segmented scan, which is really set up to expect 0 or 1.  I think it may still do the right thing with -1, but I think we should update the code to check the return value and use 1 in case 0 is not returned.

Second, a bigger issue is that I think we may not do the right thing if the source and destination files are different, but we hit an error when opening/seeking.  In this case, we probably will detect the difference, but we'll have bailed out from the function when we should have copied over the destination file with the source data.  So we may report the files as being different, but the sync won't have fixed them.  If this is true, we should probably make sure the tool really reports an error here.

I'll work on these today.

PR #45, opened 2017-03-22
test : add test case for dcp against FIEMAP

KnightKu:
Signed-off-by: Gu Zheng <gzheng@ddn.com>

### DIFF
--- a/test/tests/test_dcp/checkfiemap.c
+++ b/test/tests/test_dcp/checkfiemap.c
@@ -1,0 +1,160 @@
+/* GPL HEADER START
+ *
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 only,
+ * as published by the Free Software Foundation.
+ *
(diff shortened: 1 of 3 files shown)


adammoody:
@dsikich Since you're familiar with dcmp and the testing infrastructure, can you review this PR?

dsikich:
@adammoody Yes, I can do that. Do you also want me to review the additions they made to the test_dcmp script? There is also a pull request for that.

adammoody:
Yes, that'd be helpful too.  Thanks!

dsikich:
I actually checked on the test_octal.py script I added to nose2, and I forgot to pass in the variables to the script as well. I'm sure there are many ways to do this, but below I added the first way that worked for me.
 
Also, the dchmod_path is the only variable used in the actual bash script, the mpifu_path variable I just used to make it easier to change the call path to the script

#!/usr/bin/env python2
import subprocess
 
# chane your path here as necessary  
mpifu_path  = "~/mpifileutils/test/tests/test_dchmod/test_octal.sh"
dchmod_path = "~/mpifileutils/install/bin/dchmod"
 
def test_octal():
p = subprocess.Popen(["%s %s" % (mpifu_path, dchmod_path)], shell=True, executable="/bin/bash").communicate()

PR #46, opened 2017-03-23
Re-work dstripe to restripe files found during path walking

dinatale2:
dstripe is now able to accept a list of paths as input. It will walk the paths provided and generate a list of files to restripe. The `-s` (stripe size) and `-c` (stripe count) can now be used to filter down the list of files found during the walk. Files that are left over are restripe candidates. 

Also replaced the old lustre api calls with the much cleaner llapi_layout.

TODO: Need to update the docs. I'll update this pull request with that soon.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,4 +1,5 @@
 Bringhurst, Jon <jonb@lanl.gov>
+Di Natale, Giuseppe <dinatale2@llnl.gov>
 Fuller, Douglas <dfuller@redhat.com>
 Groff, Todd <groffta@ornl.gov>
 LaFon, Jharrod <jlafon@lanl.gov>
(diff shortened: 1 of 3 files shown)


dinatale2:
Updated to include documentation on dstripe's new behavior.

adammoody:
This looks good to me so far.  Clever method to generate a global random suffix that can be used :-)

My only comment so far is that it would help to add some more comments to the new code sections.

dinatale2:
@adammoody Went ahead and updated with comments. Also realized I wasn't freeing the path list, so now I'm cleaning that up correctly.

dinatale2:
@adammoody I went ahead an placed myself in the AUTHORS file as well.

PR #47, opened 2017-03-30
Add min file size filter to dstripe

dinatale2:
Introduce a -m/--minsize option to allow users
to filter out files that are below a certain
file size. Users will most likely want to restripe
larger files.


adammoody:
Looks good to me.  Thanks, @dinatale2.

PR #48, opened 2017-03-30
dstripe's report option should print file size

dinatale2:
Enable dstripe to print a file's size when using the
--report/-r option.

@adammoody I think the report option might be producing too much text. Do you have a preferred format to print a files info? I can incorporate it into this PR.

### DIFF
--- a/doc/markdown/dstripe.1.md
+++ b/doc/markdown/dstripe.1.md
@@ -28,7 +28,7 @@ dstripe enables one to restripe a file across the underlying storage devices. On
 :	The minimum size a file must be to be a candidate for restriping. It is possible to use units like "MB" and "GB" after the number, which should be immediately follow the number without spaces (ex. 2MB). The default minimum file size is 0MB.
 
 -r, \--report
-:	Display the stripe count and stripe size of all files found in PATH. No restriping is performed when using this option.
+:	Display the file size, stripe count, and stripe size of all files found in PATH. No restriping is performed when using this option.
 
 -v, \--verbose
 : 	Run in verbose mode.
--- a/src/dstripe/dstripe.c
+++ b/src/dstripe/dstripe.c
@@ -50,7 +50,7 @@ static void print_usage(void)
     printf("  -c, --count            - stripe count (default -1)\n");
     printf("  -s, --size             - stripe size in bytes (default 1MB)\n");
     printf("  -m, --minsize          - minimum file size (default 0MB)\n");
-    printf("  -r, --report           - input file stripe info\n");
+    printf("  -r, --report           - display file size and stripe info\n");
     printf("  -v, --verbose          - verbose output\n");
     printf("  -h, --help             - print usage\n");
     printf("\n");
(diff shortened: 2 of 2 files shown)


adammoody:
Don't know how this would look, but we could print the stripe count, stripe size, and file size in fixed length columns followed by the file name.  We could also sort the file names to print the list alphabetically.

I think we'll also want to support filtering here, too.  Does the --minsize filter apply?

dinatale2:
I can add the minsize filter since it got merged. I'll also incorporate the other filters.

dinatale2:
@adammoody I went ahead and refreshed this. My `printf` statements aren't the cleanest, but it demonstrates sample output from `--report` as a table. I think printing it as a table is a step in the right direction.

I did see a function which would take bytes and convert them to a more user friendly format (i.e. 1024 bytes generates a string `1K`). Did I miss it by chance?

adammoody:
Great.  We have a routine that helps with printing formatted bytes.  You pass it the number of bytes (integer) and it returns a double and a string giving the units.  You can see an example of it in action here:
https://github.com/hpc/mpifileutils/blob/master/src/dcp/dcp.c#L135

dinatale2:
@adammoody Went ahead and refreshed this. Do you mind running `dstripe -r` and providing feedback on it's output?

adammoody:
Since the size, count, and stripe size should work well for fixed-length fields, but the path is variable length, let's order the fields this way:

Stripe Count, Stripe Size, File Size, File Path

This way we can let the file path run out to the right as far as needed.

I happened to run a test on some 0-byte files, and it's giving the following for that:
File Path                                          Size    Cnt Str Size
---------                                          ----    --- --------
/path/to/file   0.000 B 1152921504606846978  1.00 EB

Looks like we need to fix up something in that case.

dinatale2:
@adammoody How did you make the file? I've never seen something like that happen before.

adammoody:
I think I did something like:
mkdir dir
touch dir/file1
touch dir/file2
dstripe -r dir

dinatale2:
I moved the path to the end. I still can't reproduce the issue you were seeing by just touching a file.

adammoody:
Ah, yes a touch shows reasonable results for me, too.  I must have done something unusual here.  Interestingly, if I run lfs getstripe on the files, it reports that the file "has no stripe info".  I have a couple guesses.  Let me see what I can find.

adammoody:
Ok, found it.  I had created these files while debugging dcp under totalview.  I had let dcp create the files with mknod() and then stopped dcp before it opened or wrote any data to the files.  Apparently, lustre doesn't assign any stripe information to files in this state.

In a second test, if I let dcp run a bit further to the the point where it opens the file for writing (but still hasn't written data), then lustre assigns striping info.

dinatale2:
Ah, ok. I think that's a bit of an edge case. I don't see a way to figure out if the file has no striping information using `llapi_layout`. Solving that seems like it's going to require some more research.

Also, are the column headings fine? Have a preference?

adammoody:
Yep, I think it's unlikely that a normal user will see this anyway.  Let's consider this "no stripe info" case to be low priority.

The field names look good to me.  To get things to line up more consistently, let's change the floating point fields widths to always have 2 digits after the decimal point.  So something like "XXX.XX UB".

dinatale2:
Ok, I went ahead and updated it. Let me know what you think.

adammoody:
Thanks @dinatale2.  Looks good.  Go ahead and merge.

PR #54, opened 2017-04-18
Correct random suffix generation in dstripe

dinatale2:
When generating a random suffix in dstripe, check
the full file list for potential collisions.


dinatale2:
Actually, this turns out to be unnecessary. Not sure why I thought there was an issue. Closing.

PR #66, opened 2017-06-07
added dsh print default

dsikich:
I just added a print default of 25 to the print_sums function, and an extra print line that says how many of the total number of items are being printed/displayed.


experimental/dsh/dsh.c
```diff
@@ -559,6 +560,7 @@ static void print_sums(mfu_path* origpath, uint64_t count, uint64_t allmax, uint
             //printf("%6.2f %2s %*llu %s\n", agg_size_tmp, agg_size_units, digits, (unsigned long long) count, name);
             printf("%6.2f %2s %6.2f %1s %s\n", agg_size_tmp, agg_size_units, count_tmp, count_units, name);
         }
+        printf("\n\n(printed top %d of a total of%6.2f items)\n\n", print_default, allsum_tmp);
```
adammoody: Let's drop one of the two leading newlines in the message here.

Let's use an int rather than a float for the total count.  Does allcount works?  Also, I think we need a space between "of" and the number.

Let's replace "items" with "lines" in the message.  We use "Items" to mean something else at the top.  Using a different word here will help to avoid confusion.

PR #72, opened 2017-06-13
dsh-print-updates

dsikich:
Summary of updates:
1) bump print_default from 25 to 100
2) print the whole list if print_default == -1, otherwise print minimum of print_default and allcount
3) print the value of allcount in "top 100 of ..." message instead of allsum_count, and print this as an unsigned long long

### DIFF
--- a/experimental/dsh/dsh.c
+++ b/experimental/dsh/dsh.c
@@ -516,6 +516,13 @@ static void print_sums(mfu_path* origpath, uint64_t count, uint64_t allmax, uint
     if (digits < 5) {
         digits = 5;
     }
+    
+    /* get the minimum value of print default and allcount to print 
+     * So, if print default is less than allcount and greater than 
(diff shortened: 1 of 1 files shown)


experimental/dsh/dsh.c
```diff
@@ -516,6 +516,13 @@ static void print_sums(mfu_path* origpath, uint64_t count, uint64_t allmax, uint
     if (digits < 5) {
         digits = 5;
     }
+    
+    /* get the minimum value of print default and allcount to print 
+     * So, if print default is less than allcount and greater than 
+     * zero print that, but if not, then print what is in the list */
+    if (print_default < allcount && print_default >= 0) {
```
adammoody: We'll want to keep the value of allcount, so let's define a new variable here:

uint64_t print_count = allcount;
if (print_default ...) {
  print_count = print_default;
}

Then update the loop to use print_count and change the printf to print print_count and allcount.

PR #73, opened 2017-06-15
Handle non-Lustre files gracefully in dstripe

dinatale2:
While filtering a list of files in dstripe, if a file
is not part of a lustre file system, simply skip it
for now.

I figure skipping the file would be appropriate since
we can't obtain lustre striping information and it clearly
states in the man pages that this tool only works for
lustre files. If it doesn't show up in the report, then
it should be assumed it's not a candidate for restripe.
Thoughts on that?


adammoody:
Thanks, @dinatale2.  Sounds reasonable to me.  In keeping with the (verbose) coding style, please update the if/continue blocks to have braces, then go ahead and merge in.

PR #76, opened 2017-06-15
Freeing sync lists

dsikich:
1. I updated this to check for NULL in dcmp_strmap_compare before freeing the lists when sync is not being used.
2. I took out the -o option in the dcmp man pages since it doesn't seem to be working or we don't know how to use it
3. ran pandoc on markdown pages for tools


src/dcmp/dcmp.c
```diff
@@ -1054,8 +1054,8 @@ static void dcmp_strmap_compare(mfu_flist src_list,
     mfu_flist dst_compare_list = mfu_flist_subset(dst_list);
   
     /* remove and copy lists for sync option */ 
-    mfu_flist dst_remove_list = MFU_FLIST_NULL;
-    mfu_flist src_cp_list     = MFU_FLIST_NULL; 
+    mfu_flist dst_remove_list = NULL;
+    mfu_flist src_cp_list     = NULL;
```
adammoody: Let's keep with setting these to MFU_FLIST_NULL, but check for that instead of NULL in the if/free blocks below.

PR #79, opened 2017-07-13
Rename website from fileutils -> mpifileutils

gonsie:
Updated the repo name (as many places as I could find).


bringhurst:
Hey @gonsie, do you know if those mailing lists are still correct?

bringhurst:
Also, we should probably remove the google analytics tracking code, since it's setup for the old fileutils.io.

PR #87, opened 2017-08-02
dbz2 added for parallel compression---Signed-off-by Ahana Roy Choudhu…

ahanagemini:
…Ahana Roy Choudhury, ahanagemini@gmail.com
Code for parallel compression is in experimental/dbz2. Please feel free to contact me for any issues related to the pull request.


adammoody:
Thanks, @ahanagemini!  I haven't had time to read this in detail yet, but I like the idea and appreciate all of the work you've put in for this.  In a quick skim so far, it would help to apply our astyle options to the source files and push another commit.  I think the following commands will work:
# download astyle tarball from sourceforge
https://sourceforge.net/projects/astyle/

# build astyle binary
tar -zxf astyle_3.0.1_linux.tar.gz
pushd astyle/src
  make -f ../build/gcc/Makefile
popd

# apply astyle to foo.c file
astyle/src/bin/astyle --options=astyle.options foo.c

adammoody:
Thanks, @ahanagemini  Another request.  We have some POSIX I/O wrappers to help with errors and EINTR which sometimes show up under debuggers:

https://github.com/hpc/mpifileutils/blob/master/src/common/mfu_io.h

Would you please replace the POSIX I/O call with the mfu wrapper where a wrapper exists?

adammoody:
Thanks for your work, @ahanagemini.  Let's also fix up the error cases.  Instead of calling exit() or MPI_Finalize() during an error, would you please change the code to call MPI_Abort instead, e.g.,

MPI_Abort(MPI_COMM_WORLD, 1);

Some MPI libraries hang on MPI_Finalize until all ranks call it, so if just one rank detects an error, the job would hang.  Calling exit() generally behaves better, but that depends on the good the job launcher is.

adammoody:
Also, we have some wrappers for allocating and freeing memory.  These are defined in src/common/mfu_util.h.  These wrappers abort if the allocation fails, so it's not necessary to check whether the returned pointer is NULL, which cleans up code calling malloc.

Would you please replace malloc() with MFU_MALLOC() and replace calls to free(foo) with mfu_free(&foo)?  Then go ahead and remove the checks for NULL on malloc calls.

Note in the free call, you pass the address of the pointer variable.  This is because the free call sets the pointer back to NULL after freeing the memory.

PR #88, opened 2017-08-25
adding code to track if read error was on src or dest side

dsikich:
* added code to dcmp_strmap_compare_data & dcmp_compare_data to track whether or not the open file/reading errors were on the src or dest side


adammoody:
Looking at this code again, let's move the MFU_LOG message about src/dest inside of the function.  This way we can also report the type of operation that failed, e.g., open vs read vs lseek.  We'll still return -1 from the function and print the second error message from the calling routine.

PR #98, opened 2017-12-18
dcp: add pure copy mode support

KnightKu:
Add pure mode support pure mode, e.g. if src ends with trailing slash, don't create additional
directory level at the destination, just copy sub dirs/files under src, similar to "rsync /foo/bar/ /foo1/".


dsikich:
@KnightKu Thanks for catching this!

PR #101, opened 2018-01-19
Dcp minor improve

KnightKu:


dsikich:
@KnightKu Thanks for improving the error checking in dcp!

PR #103, opened 2018-02-12
dcp: rework the data copy routine with libcircle

KnightKu:
Rework the data copy routinue with libcircle, so that we can trace
the progress timely, like dwalk.

[2018-02-12T11:06:41] [0] [../../../src/common/mfu_flist_copy.c:1508] Copying data.
2018-02-12 11:06:51: files copied 13.45%..[844/6274]...
2018-02-12 11:07:02: files copied 28.36%..[1779/6274]...
2018-02-12 11:07:12: files copied 46.21%..[2899/6274]...
2018-02-12 11:07:22: files copied 69.51%..[4361/6274]...
2018-02-12 11:07:32: files copied 82.07%..[5149/6274]...
2018-02-12 11:07:41: files copied 100.00%..[6274/6274]...

### DIFF
--- a/src/common/mfu_flist_copy.c
+++ b/src/common/mfu_flist_copy.c
@@ -97,6 +97,14 @@ typedef struct {
     int   fd;   /* file descriptor */
 } mfu_copy_file_cache_t;
 
+typedef struct {
+    char *dest_path;
+    char *name;
+    uint64_t offset;
+    uint64_t len;
+    uint64_t size;
+} mfu_cp_data_t;
+
 /****************************************
  * Define globals
  ***************************************/
@@ -107,6 +115,13 @@ static mfu_copy_stats_t mfu_copy_stats;
 /** Cache most recent open file descriptor to avoid opening / closing the same file */
 static mfu_copy_file_cache_t mfu_copy_src_cache;
 static mfu_copy_file_cache_t mfu_copy_dst_cache;
+
+static int CUR_NUMPATHS;
+static mfu_param_path* CUR_PATHS;
+static mfu_param_path* CUR_DESTPATH;
+static mfu_copy_opts_t* CUR_OPTS;
+static mfu_file_chunk* CUR_CHUNK;
+static uint64_t CUR_TOTAL_FILES;
 
 int mfu_input_flist_skip(const char* name, void *args)
 {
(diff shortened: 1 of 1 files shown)


dsikich:
@KnightKu there is another version of dcp that does the copy as it is walking the directory with libcircle (dcp1). It looks like this code would fit better with that version. The purpose of this version is to be able to do the libcircle walk, store it in an flist, and then operate on the flist. I don't think this implementation with libcircle fits this dcp version.

KnightKu:
Hi @dsikich , yeah, I saw the dcp1 and looked into the libcircle code, I think there are two benefits with this change, first one is we can use the internal timer to trace the progress timely, and the other one is libcircle can auto-balance each worker thread's workload to improve the performance, right?

dsikich:
@KnightKu I agree, I think being able to trace progress will be great! In terms of the auto-balance I am not aware of how it will help performance because the bytes are evenly distributed among the processes already in the mfu_file_chunk_list_alloc function (mfu_flist_chunk.c).

@adammoody do you know if there is a benefit to the libcircle implementation over using MPI calls?

adammoody:
@KnightKu, The main reason we implemented this second dcp tool w/o libcircle is to prevent consecutive regions of files from being randomly assigned to processes.  This has the effect of reducing file byte-range lock contention within lustre.  We found this to be helpful when using a lot of processes to copy a single big file (100GB+).  Adding libcircle in there would cause that problem to come back.

Having said that, it would be nice to expose this an option to the user.  We could default to the pre-assigned and static method we have now and allow the user to switch to this libcircle version, or perhaps we could chose dynamically based on the stats we gather from the file list.

Could you modify this to provide both methods instead of replacing the one that was there?

Also, it'd be useful to report bytes copied, too, in case the user is moving a single large file.

adammoody:
As for reporting progress in the existing static method, I think we could use an ibcast / allreduce method like in dcmp.  If we can come up with a function for this, we could probably use it in both tools.

KnightKu:
Hi @adammoody , thanks very much for your explanation.
Yeah, can't agree more, will rework it as you suggested soon.

adammoody:
@KnightKu, I wanted to circle back and check on this.  Where did we leave things here?

adammoody:
@KnightKu , are you up for refreshing the dcp progress message?  This is something that would be good to include in a future release.

PR #106, opened 2018-03-01
Correct detection of Lustre API

dinatale2:
Correct lx_find_lustre.m4 to properly check for
header files. Also check that 'llapi_layout_alloc' is
a function in the Lustre API because dstripe currently
depends on it.

### DIFF
--- a/m4/lx_find_lustre.m4
+++ b/m4/lx_find_lustre.m4
@@ -1,26 +1,45 @@
-# looks for lustre header files
-# if found, defines LUSTRE_SUPPORT=1 in config.h
+dnl #
+dnl # Determines if Lustre API is installed.
+dnl # Checks if Lustre API has llapi_layout.
+dnl #
 AC_DEFUN([X_AC_LUSTRE], [
-  AC_MSG_CHECKING([whether to enable Lustre support])
-  AC_ARG_ENABLE([lustre],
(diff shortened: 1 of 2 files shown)


dsikich:
@dinatale2 Is there a version of Lustre that always supports this llapi_layout?

dinatale2:
@dsikich According to `git blame` on the Lustre code base, some version of Lustre 2.6 started supporting `llapi_layout`.

dsikich:
@dinatale2 okay, thanks. I think this is good. It might be nice to have a message to say which version of Lustre someone needs to run dstripe. Unless, is it possible to have a Lustre version < 2.6 and still use the llapi_layout?

dinatale2:
@dsikich Sure, I can try and improve the messaging. I also have an idea for a tweak to this that I would like to try. Stay tuned.

dinatale2:
@dsikich I updated the patch. The tweak I wanted to introduce to ` m4/lx_find_lustre.m4` didn't make sense. I did improve the messaging in `dstripe` to indicate it's current Lustre version dependent (2.7 is when `llapi_layout` was officially introduced).

dsikich:
@dinatale2 okay, great. I think that will help let people know what version of Lustre definitely supports it right away.

PR #107, opened 2018-03-09
Introduce Lustre 2.5 support

dinatale2:
Lustre 2.5 is still a widely used version of Lustre. Introduce
support for Lustre 2.5 APIs since llapi_layouts were not introduced
until 2.7.

This is currently untested. I am attempting to find a way to test it.


dsikich:
@dinatale2 I tested dstripe with and without the m4 check for the llapi ... I think we can go ahead and merge this in.

PR #111, opened 2018-03-26
Adds dfmake which can be used to create file/directory trees

adammoody:


adammoody:
@levatin1, here's the PR for dfmake.  You should be able to push updates to the filemaker branch to update this.

PR #112, opened 2018-04-06
fix initialization of do_sync and sparse variables in dcp

dsikich:
The do_sync variable was not being initialized to zero by default,
which would cause undefined behavior. Also, the sparse variable
was being initialized to zero by default, but it should be a
boolean type based on how the struct was specified.

### DIFF
--- a/src/dcp/dcp.c
+++ b/src/dcp/dcp.c
@@ -26,6 +26,7 @@
 #include <inttypes.h>
 #include <unistd.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


KnightKu:
Hi @dinatale2, can we merge this patch now? we hit the same problem.

dinatale2:
@KnightKu Merged!

KnightKu:
@dinatale2 cool, thanks.

PR #123, opened 2018-05-28
drm: add traceless mode support

KnightKu:
Add traceless mode support to drm (-T/--traceless), if the option is enabled,
unlink file/dir/link will not change its parent dir's timestamp (atiem/mtime),
it is useful if some guy want to remove the extra files under dst dir after
a dcp, but not change the timestamp of the dst filesystem tree, in order to
keep the same as src.

### DIFF
--- a/src/common/mfu_flist_remove.c
+++ b/src/common/mfu_flist_remove.c
@@ -40,6 +40,7 @@
 #include "libcircle.h"
 #include "dtcmp.h"
 #include "mfu.h"
+#include "mfu_flist_internal.h"
 
 /*****************************
  * Global functions used by remove routines
@@ -576,7 +577,7 @@ static void remove_libcircle(mfu_flist list, uint64_t* rmcount)
 /* removes list of items, sets write bits on directories from
  * top-to-bottom, then removes items one level at a time starting
  * from the deepest */
-void mfu_flist_unlink(mfu_flist flist)
+void mfu_flist_unlink(mfu_flist flist, bool traceless)
 {
     int level;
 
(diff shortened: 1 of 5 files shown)


adammoody:
When deleting lots of files in a single parent, it seems like this will require a lot of extra stats and setting the timestamps.

Actually, now that I think about it, we could end up with race conditions because two different processes might be removing items from a parent directory at the same time, so the stat of one could get the wrong time values.

To solve both problems, it seems like we'd want to capture the timestamps once before removing files, delete everything, and then go back and update the timestamps.

src/common/mfu_flist_remove.c
```diff
@@ -81,6 +117,14 @@ static void remove_type(char type, const char* name)
                  );
     }
 
+    if (traceless && !status) {
+        if(utimensat(AT_FDCWD, pdir, times, AT_SYMLINK_NOFOLLOW) != 0) {
+            MFU_LOG(MFU_LOG_ERR, "Failed to changeback timestamps on %s utime() errno=%d %s",
+                pdir, errno, strerror(errno));
+        }
+
+        mfu_free(&pdir);
```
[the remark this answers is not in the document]
KnightKu: Hi @adammoody, thanks for the catch, will fix it.

KnightKu:
Hi @adammoody , thanks for the comment.

yeah, this race condition is possible.

>  To solve both problems, it seems like we'd want to capture the timestamps once before removing files, delete everything, and then go back and update the timestamps.

I am wondering whether check the timestamp of pdir when we setback it can resolve this problem, e.g. if the mtime is older than the mtime we keep (means the timestamp we stored is a race one), just drop it, don't update it.

KnightKu:
Hi @adammoody , could you please review the updated one? thanks.

KnightKu:
ping @adammoody

src/common/mfu_flist_remove.c
```diff
@@ -588,6 +589,54 @@ void mfu_flist_unlink(mfu_flist flist)
...
+        /* get user data if needed */
+        if (file_list->have_users == 0) {
+            mfu_flist_usrgrp_get_users(pstatlist);
+        }
+
+        /* get groups data if needed */
+        if (file_list->have_groups == 0) {
+            mfu_flist_usrgrp_get_groups(pstatlist);
+        }
+
```
adammoody: We've got a new mfu_flist_set_detail(flist, 1) function that we can use here to replace lines 602-615.

adammoody:
This should avoid the race condition.

It looks like this will try to set timestamps on directories that may have been deleted.  Does it still function in those cases?

It'd be good to develop it further.  Ideally, to avoid imposing extra load on the file system, it'd be good to find a method that only sets a timestamp once for each parent directory, and one that does not try to set timestamps on deleted items.  We could probably use a mfu_flist_map() to help with this.  Could you take a look at that?

adammoody:
Oh, we could use DTCMP_Rank_strings to identify a leader for each parent directory.

https://github.com/LLNL/dtcmp/blob/817a22e187c1413fbc61ca124b5ab3748cc90b58/src/dtcmp.h#L665

The remap alone would not do the trick since it won't remove duplicate entries.

adammoody:
And let's change that to Rankv_strings, since Rank_strings assumes each process has the same count, which is not necessarily true:

https://github.com/LLNL/dtcmp/blob/817a22e187c1413fbc61ca124b5ab3748cc90b58/src/dtcmp.h#L714

src/common/mfu_flist_remove.c
```diff
@@ -588,6 +589,54 @@ void mfu_flist_unlink(mfu_flist flist)
...
+                continue;
+            }
+            /* insert item into output list */
+            mfu_flist_insert_stat(pstatlist, pdir, st.st_mode, &st);
+            mfu_free(&pdir);
+        }
+
+        /* compute global summary */
+        mfu_flist_summarize(pstatlist);
+    }
```
adammoody: Let's also add an MPI_Barrier() in here after summarize to be sure that all procs have executed their stat calls before moving on to deleting things.  It's not necessary since summarize includes allreduce calls.  However, it won't hurt much, and it's good practice to make this synchronization step explicit in our control flow.

KnightKu:
Hi @adammoody , thanks for your comments, will improve it as you suggested.

KnightKu:
Hi @adammoody , could you please review it again? thanks.

adammoody:
Hi @KnightKu , did you push the new changes?  Does this include the Rankv change?

KnightKu:
Hi @adammoody , I just use mfu_flist_spread to spread the items among ranks, is this way OK?

KnightKu:
ping

src/common/mfu_flist_remove.c
```diff
@@ -588,6 +589,81 @@ void mfu_flist_unlink(mfu_flist flist)
     int levels, minlevel;
     mfu_flist* lists;
     mfu_flist_array_by_depth(flist, &levels, &minlevel, &lists);
+    mfu_flist pstatlist;
+    uint64_t size = mfu_flist_size(flist);
+    char* strings[size];
```
adammoody: I don't know that this line is safe with all C compilers, so let's use a MFU_MALLOC here and a corresponding free later.

adammoody:
Thanks, @KnightKu.  This looks good.  It solves the stat storm on the directory names, so that will reduce pressure on the metadata server.  With more work, I think we could also avoid the access() that looks for deleted directories, but that's a bit tricky, so we can save that for later if it happens to become a problem.

I do have one suggestion to change the strings array to use MALLOC, which I think will be safer across C compilers.

src/common/mfu_flist_remove.c
```diff
@@ -588,6 +589,81 @@ void mfu_flist_unlink(mfu_flist flist)
...
+        for (idx = 0; idx < size; idx++) {
+            if(group_rank[idx] == 0) {
+                /* stat the item */
+                struct stat st;
+                char *pdir = strings[idx];
+                int status = mfu_lstat(pdir, &st);
+
+                if (status != 0) {
+                    MFU_LOG(MFU_LOG_ERR, "mfu_lstat(): %d", status);
+                    mfu_free(&pdir);
```
adammoody: Need to do this?  I think it would lead to a possible double free at line 652 where we free each value in the strings array.

adammoody:
Oh, since we're using -t in some tools to mean "text output", let's pick another option letter.  At some point, we may want a tool to support both --traceless and --text.

KnightKu:
Hi @adammoody , thanks very for your quick review, let me fix these issues.

KnightKu:
Hi @adammoody , updated it, thanks.

PR #127, opened 2018-05-30
Add seperate dsync tool extracted from dcmp

dsikich:
This is still very much a work in progress. There is a lot of
duplicated code that exists in both dsync and dcmp, so work needs
to be done to extract the common code and make it available to both
tools. The dsync tool is functional, but needs heavy refactoring.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -82,6 +82,7 @@ AC_CONFIG_FILES([Makefile                \
                  src/dbcast/Makefile     \
                  src/dchmod/Makefile     \
                  src/dcmp/Makefile       \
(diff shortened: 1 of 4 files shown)


adammoody:
Thanks, @dsikich.  Looks good.  Wow, there is a lot of code in that tool.  Hopefully, we can find a good set of calls to create some common functions.

adammoody:
Looks good to me if you want to merge this.  In that case, we can open another PR once you get into the refactoring.  We can also go ahead and drop the --sync option and related logic from dcmp to simplify that code a bit.

PR #128, opened 2018-05-30
mfu: write flist to text file

adammoody:
To satisfy request to write file list to text file.  The final version should allocate a smaller write buffer to not use too much memory.  Also, it would be nice to provide an interface to allow caller to customize fields and formats used when writing data.

### DIFF
--- a/src/common/mfu_flist.h
+++ b/src/common/mfu_flist.h
@@ -198,6 +198,12 @@ void mfu_flist_write_cache(
     mfu_flist flist
 );
 
+/* write file list to text file */
+void mfu_flist_write_text(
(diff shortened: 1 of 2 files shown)


KnightKu:
Cool, looking forward to this feature, it would be really useful to end users.

adammoody:
Thanks @dsikich.  Look good now?

Since there may be interest in getting this added sooner rather than later, I'll drop the WIP label and we can open issues to come back and think about an interface to express the fields/format to be written.  I'm thinking an printf() style string or something that the user can pass in where we'll fill in %s entries with formatted values.

Issue #129, opened 2018-05-31
Allow custom fields and format when printing file info


adammoody:
When printing file info to the screen or writing it to a text file, it would be nice to allow the user to specify the fields and formats they'd like to use.  This could be done with printf/strftime-style format string, e.g.

%F - full path of file
%f - basename of file

%m - mode in octal
%M - mode in long dash form -rwxr-x---

%s - size in bytes
%S - size as float with human-readable units

adammoody:
Perhaps reuse "find -printf" syntax

PR #130, opened 2018-06-11
make use of text output fucntion

KnightKu:
Text output feature has been landed, let's make use of it, change text
to the default output format, and another option '-r/--raw' is introduced to
enable original native format.

### DIFF
--- a/experimental/dsh/dsh.c
+++ b/experimental/dsh/dsh.c
@@ -1756,6 +1756,7 @@ int main(int argc, char** argv)
     char* inputname  = NULL;
     char* outputname = NULL;
     int walk = 0;
+    int text = 0;
     /* set print default to 25 for now */
     int print_default = 100;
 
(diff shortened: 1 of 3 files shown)


adammoody:
Thanks @KnightKu.  I'd like to consider our tool interface a bit on this one.  For the text output, we should develop a way for the user to specify the fields they want printed, and perhaps the format of each field.  We'll need to define a command line syntax for that.  Also, I'd like to update our binary format to handle variable length paths, instead of writing fixed length records based on the longest path as it does now, which tends to generate very large output files when writing a lot of records with even a single long path.  So we may end up with more than one binary format.  Finally, we have some existing tools that expect the tools to output binary format when using -o, and we'd break them with a change like this.  That might be ok, but we should be careful about it.

I can imagine this might be useful while we work to hash all of that out, though.  Until then let's keep the binary format as the default with -o so we don't break existing workflows, and let's define a new (hidden) option to write text output.

adammoody:
Related issue to define text fields/formats:  https://github.com/hpc/mpifileutils/issues/129

experimental/dsh/dsh.c
```diff
@@ -1767,13 +1768,14 @@ int main(int argc, char** argv)
...
         {"verbose",  0, 0, 'v'},
+        {"text",     0, 0, 't'},
         {0, 0, 0, 0}
     };
 
     int usage = 0;
     while (1) {
         int c = getopt_long(
-                    argc, argv, "f:i:o:lhv",
+                    argc, argv, "f:i:o:lhvrt",
```
adammoody: Thanks @KnightKu.  Looks good, but let's drop the "r" here.  At least I think you mean to drop it.

KnightKu: yeah, true, forgot it here, let me fix it.

KnightKu:
fixed it, thanks.

adammoody:
Thanks, @KnightKu!

Issue #36, opened 2017-01-26
Check whether umask is needed in dchmod


adammoody:
I think chmod takes umask into account in some cases.  If possible we should try to emulate that.

dsikich:
Hi Adam,

I haven’t been able to find any information about chmod taking into account umask. The man page does not mention it. From what I have read so far I gathered that umask is used to set the permissions on newly created files and/or directories. I can’t think of a reason the umask would be taken into account? I did read something about the userid though, where you should not be able to set certain permissions if you are not the owner of the file I think? I did find information in the man page about setuid & setgid that I don’t believe are supported in dchmod:


SETUID AND SETGID BITS

       chmod  clears  the set-group-ID bit of a regular file if the file's group ID

       does not match the user's effective group ID or one of the user's supplemen-

       tary  group  IDs,  unless  the  user has appropriate privileges.  Additional

       restrictions may cause the set-user-ID and  set-group-ID  bits  of  MODE  or

       RFILE  to be ignored.  This behavior depends on the policy and functionality

       of the underlying chmod system call.  When in doubt,  check  the  underlying

       system behavior.



       chmod  preserves  a directory's set-user-ID and set-group-ID bits unless you

       explicitly specify otherwise.  You can set or clear the bits  with  symbolic

       modes  like  u+s  and  g-s,  and you can set (but not clear) the bits with a

       numeric mode.


Thanks,
Danielle

adammoody:
This is not urgent, but I thought I saw some cases with chmod where umask comes into play when using symbolic mode without a leading letter, like "+rw" rather than "u+rw".  I just threw this up as a reminder to go look into it later so I don't forget.

Checking the man page, it does look like umask comes into play:

       A  combination of the letters ugoa controls which users' access to the file will be changed: the user
       who owns it (u), other users in the file's group (g), other users not in the file's group (o), or all
       users  (a).   If  none of these are given, the effect is as if a were given, but bits that are set in
       the umask are not affected.

We can figure this out and fix this later.

adammoody:
So I think this means that if my umask=077 (all group and other bits set), and if I then do a "+r", this will enable just the owner read bit.  If my umask=007, then a "+r" would enable both the owner and group read bits.  So the operation only affects bits for which the umask is 0.

adammoody:
@dsikich, this is done now, right?  Do we also have some test cases for this in the test suite?

dsikich:
[...]

PR #133, opened 2018-06-13
Fix parse error in dchmod

dsikich:
When using a+r for instance, dchmod was incorrectly parsing this
as something that specified no letter (i.e +r), which uses umask.
This was causing dchmod to incorrectly assign permissions when
using the "all" attribute.


adammoody:
Here's the old issue just for reference:  https://github.com/hpc/mpifileutils/issues/36

PR #135, opened 2018-07-05
dcp: add a barrier between copying data and setting metadata

KnightKu:
Add a barrier between copying data and setting metadata, to make sure the
content update is synced before we start to set metadata, to avoid some metadata
mismatch problem, e.g. wrong modify timestamp.


adammoody:
Good idea.  Thanks @KnightKu .

PR #138, opened 2018-07-10
mfu: move link's attributes set into mfu_copy_set_metadata

KnightKu:
Move link's ownership, permissions, and timestamps setting later into
mfu_copy_set_metadata, and now the link's timestamp will be synced up
if option preserve enabled, the same as we do to files.


dsikich:
@KnightKu Where did you move the code to? I don't see it added anywhere in this PR?

dsikich:
@KnightKu Ok, I see. Looks good.

PR #140, opened 2018-08-01
Dsync improve

KnightKu:
Here we add two new options to dsync:
-n/--dry-run, used to support dry-run mode, similar to dcmp if enabled. It should be helpful to users, who want to estimate the workload before doing the real syncing job.
 -N/--no-delete, used to support keeping extraneous files from destination dirs. Because at default, we remove  extraneous files/dirs from destination dirs, so here we can use this option to avoid to remove these extraneous files.

### DIFF
--- a/src/dsync/dsync.c
+++ b/src/dsync/dsync.c
@@ -50,6 +50,7 @@ static void print_usage(void)
     printf("  -o, --output field0=state0@field1=state1,field2=state2:file "
            "- write list to file\n");
     printf("  -n, --dry-run dry-run, just show the diff, don't do real sync\n");
+    printf("  -N, --no-delete  don't delete extraneous files from destination dirs\n");
     printf("  -d, --debug enable debug mode\n");
     printf("  -v, --verbose\n");
     printf("  -h, --help  - print usage\n");
@@ -155,6 +156,7 @@ struct dsync_options {
     int dry_run;                   /* dry run */
     int verbose;
     int debug;                     /* check result after get result */
+    int delete;                    /* delete extraneous files from destination dirs */
     int need_compare[DCMPF_MAX];   /* fields that need to be compared  */
 };
 
@@ -163,6 +165,7 @@ struct dsync_options options = {
     .dry_run      = 0,
     .verbose      = 0,
     .debug        = 0,
+    .delete       = 1,
     .need_compare = {0,}
 };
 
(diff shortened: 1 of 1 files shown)


src/dsync/dsync.c
```diff
@@ -1173,9 +1188,14 @@ static void dsync_strmap_compare(mfu_flist src_list,
     mfu_flist dst_remove_list = MFU_FLIST_NULL; 
     mfu_flist src_cp_list     = MFU_FLIST_NULL; 
 
-    /* create dst remove list if sync option is on */
-    dst_remove_list = mfu_flist_subset(dst_list);
-    src_cp_list = mfu_flist_subset(src_list);
+    if (mfu_copy_opts->do_sync) {
+        /* create dst remove list if sync option is on */
+        if (options.delete) {
+            dst_remove_list = mfu_flist_subset(dst_list);
```
dsikich: @KnightKu It looks like the dst_remove_list is only being created in the delete option case now? If so, I would think the dst_remove_list also needs to be created when a regular sync is done.

KnightKu: Hi @dsikich, I don't think we need dst_remove_list if delete option is disabled. You know we are using compare expression, so we can get the "extraneous files in dest" via EXIST=ONLY_DEST now, and dst_remove_list here is only used to remove the extraneous files if delete option enabled.

KnightKu:
Hi @dsikich, could you please review this series furtherly? thanks.

dsikich:
@KnightKu I'm not sure this is the most intuitive interface for dsync. It makes more sense to me to have the default remove files only in the destination directory for a sync. And have an option that doesn't remove files only in the destination if that is the desired functionality. What is your motivation for doing it this way?

KnightKu:
Hi @dsikich, makes sense, let me rework it soon, thanks!

KnightKu:
Hi @dsikich , please help to review it again, thanks!

dsikich:
@KnightKu It isn't clear to me why there is so much checking added for the do_sync option? Since this is the dsync tool do_sync should always be set to true by default.

KnightKu:
Hi @dsikich , sorry, here just make use of do_sync for dry-run check, so that we don't need to add additional variable for dry-run.

dsikich:
@KnightKu Oh ok. Makes sense.

dsikich:
@KnightKu But I think it would be good to add another variable for the dry run check instead. It makes the checks more clear.

KnightKu:
Hi @dsikich , OK, you're right, let's keep "one variable only for one thing", will rework it soon, thanks.

KnightKu:
Hi @dsikich , reworked it, please review it again, thanks.

adammoody:
@KnightKu, for the --no-delete, what are you thinking for files that already exist, but have the wrong type or the wrong content?  Do you anticipate users wanting to replace those?

I think the code is not quite right either way, but we need to decide in order to fix it.

adammoody:
Got a PR to fix that here:
https://github.com/hpc/mpifileutils/pull/155

PR #142, opened 2018-08-03
Set errno = 0 before calling readdir()

morrone:
Technically speaking when calling readdir() "If  the  end  of  the
directory stream is reached, NULL is returned and errno is not changed."
Therefore if we want to test errno, we need to properly initialize it to zero
before calling readdir().


adammoody:
Thanks, @morrone

adammoody:
And thank you @dinatale2 for the reviews!

PR #155, opened 2018-08-23
Dsyncfix

adammoody:
- lines to actually compare files when sizes equal were commented out
- need to delete items from destination in case where they must be overwritten (e.g., new types).


adammoody:
I also notice that dsync doesn't currently update permissions or timestamps if the type/size/content is the same, but just the metadata has changed from source to destination.  We probably need a new list to track files that need a metadata refresh.

PR #161, opened 2018-08-27
dsync: add the missing *contents* to long options array

KnightKu:
add the missing *contents* to long options array


adammoody:
Thanks!

PR #171, opened 2018-08-29
fix bugs turned up on sierra testing

adammoody:
- fix bug surfaced on sierra with reading cache files
- improve error messages
- fix so dsync will create target directory if it doesn't exist


adammoody:
@dsikich , can you think whether these fixes (especially the one for dsync) will trip up anything, or do we need more like them?

adammoody:
the second fix here is so that someone can do "dsync src dest", where the dest directory does not exist yet, it should now be created

dsikich:
@adammoody I can't think of anything that would trip up. I know dcp already creates the dest dir if it doesn't already exist. I don't think any other tool will need that though?

PR #197, opened 2018-10-26
dcmp lite mode

dsikich:
Add option to dcmp to allow comparison of only metadata. If metadata
is the same, then contents are assumed to be the same. Similarly,
if the metadata is different, the contents are assumed to be
different. The lite mode does no comparison of data/content.

### DIFF
--- a/doc/rst/dcmp.1.rst
+++ b/doc/rst/dcmp.1.rst
@@ -42,6 +42,14 @@ OPTIONS
    command. Files walked, started, completed, seconds, files, bytes
    read, byte rate, and file rate.
 
+.. option:: -l, --lite
+
+  lite mode does a comparison of file modification time and size. If
+  modification time and size are the same, then the contents are assumed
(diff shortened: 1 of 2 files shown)


doc/rst/dcmp.1.rst
```diff
@@ -42,6 +42,13 @@ OPTIONS
    command. Files walked, started, completed, seconds, files, bytes
    read, byte rate, and file rate.
 
+.. option:: -l, --lite
+
+  lite mode does a comparison of metadata. If metadata is the same,
```
adammoody: Let's say "size and mtime" instead of "metadata" here and in the -h output so the user knows which particular metadata fields we're looking at.

dsikich: @adammoody Got it. The code is updated.

adammoody:
Looks good.  Thanks again.

PR #198, opened 2018-10-26
Fix dwalk default output

dsikich:
dwalk should have been printing the summary to the screen by default.
The reason it was only happening when verbose was set was because
there were a couple of bugs in the mfu_flist_print_summary function.
MFU_LOG_INFO was not set by default, and it was requiring VERBOSE to
be set in order to use MFU_LOG_INFO log/print statements. So
mfu_flist_print_summary would print nothing unless the debug level
was set to VERBOSE. Also, mfu_flist_print_summary does not need
to be called in dwalk.c, because it is already called in mfu_flist_walk.

### DIFF
--- a/src/dwalk/dwalk.c
+++ b/src/dwalk/dwalk.c
@@ -289,6 +289,9 @@ int main(int argc, char** argv)
     int text = 0;
     struct distribute_option option;
 
+    /* set debug level to MFU_LOG_INFO since it defaults to ERROR */
+    mfu_debug_level = MFU_LOG_INFO;
+
     int option_index = 0;
     static struct option long_options[] = {
         {"input",        1, 0, 'i'},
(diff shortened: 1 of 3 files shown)


src/dwalk/dwalk.c
```diff
@@ -498,9 +498,6 @@ int main(int argc, char** argv)
         mfu_flist_print(flist);
     }
 
-    /* print summary about all files */
-    mfu_flist_print_summary(flist);
```
adammoody: Actually, this line is still needed to print the summary.  The mfu_flist_print() call above prints info about individual files, whereas mfu_flist_print_summary() prints summary information about the list as a whole.

dsikich:
@dsikich The mfu_flist_print_summary was being from mfu_flist_walk, so the flist summary was printing twice. But based on our conversations I'm going to remove that call, and just put it in the tool itself.

dsikich:
@adammoody code is updated.

adammoody:
Thanks!  Looks good to me.

PR #199, opened 2018-11-02
Batchcopy

adammoody:
- fix for INVALID_DATATYPE due to DTCMP call in chunk_list_lor with empty list
- fix bug preventing copying of symlinks
- adds functions for allocating and deleting copy_opts structures, sets default values in allocated object
- extends flist_copy to operate in batches of files of specified size, useful in dsync to act as a type of checkpoint in dsync
- adds new --batch-files option to dsync


adammoody:
Needs to be rebased

PR #201, opened 2018-11-28
Fix dcmp lite mode metadata check

dsikich:
This mode was not comparing the modification times correctly
in the compare function. It was mistakenly using the compare lists
instead of the original src and dst lists. The compare lists are
empty in lite mode and only used when comparing the contents of a
file.

### DIFF
--- a/src/dcmp/dcmp.c
+++ b/src/dcmp/dcmp.c
@@ -923,10 +923,10 @@ static int dcmp_strmap_compare(mfu_flist src_list,
         }
 
         /* get mtime seconds and nsecs to check modification times of src & dst */
-        src_mtime      = mfu_flist_file_get_mtime(src_compare_list, src_index);
-        src_mtime_nsec = mfu_flist_file_get_mtime_nsec(src_compare_list, src_index);
-        dst_mtime      = mfu_flist_file_get_mtime(dst_compare_list, dst_index);
-        dst_mtime_nsec = mfu_flist_file_get_mtime_nsec(dst_compare_list, dst_index);
+        src_mtime      = mfu_flist_file_get_mtime(src_list, src_index);
+        src_mtime_nsec = mfu_flist_file_get_mtime_nsec(src_list, src_index);
+        dst_mtime      = mfu_flist_file_get_mtime(dst_list, dst_index);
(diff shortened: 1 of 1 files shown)


src/dcmp/dcmp.c
```diff
@@ -948,10 +948,14 @@ static int dcmp_strmap_compare(mfu_flist src_list,
...
 
-        /* make a copy of the src and dest files where the data needs
-         * to be compared and store in src & dest compare lists */
-        mfu_flist_file_copy(src_list, src_index, src_compare_list);
-        mfu_flist_file_copy(dst_list, dst_index, dst_compare_list);
+        /* Make a copy of the src and dest files where the data needs
+         * to be compared and store in src & dest compare lists. We
+         * dont need to bother copying into compare list if lite
+         * option is on */
+        if (options.lite) {
```
adammoody: should this be?
```
if (! options.lite)
```

dsikich: @adammoody yep. thank you. I just updated it.

adammoody:
I'm confused a bit looking at that surrounding code.  There are a couple lines that set the CONTENTS to be COMMON after the if block which compares mtimes.

https://github.com/hpc/mpifileutils/blob/f5be21671f357435d5fbdbff6607d29d11bf1a74/src/dcmp/dcmp.c#L943

In the non-lite mode, it seems like we wouldn't know that until after reading bytes to compare data.

Should those COMMON lines be moved up inside of the if lite block?

https://github.com/hpc/mpifileutils/blob/f5be21671f357435d5fbdbff6607d29d11bf1a74/src/dcmp/dcmp.c#L931

dsikich:
@adammoody yeah I suppose it is better to put it in the if block. In non-lite mode it would have marked them COMMON but copied those files into the two compare lists, and checked re-updated them anyway.

dsikich:
@adammoody i updated it. thanks.

adammoody:
Ok, thanks.  In that case, let's just have the if (options.lite) block always call continue, and then we can remove the later "if (!options.lite)" guard.

dsikich:
@adammoody ok. that makes the most sense. it is updated again.

adammoody:
Looks good!

PR #203, opened 2018-12-12
drm: Add remove file capability during walk

dsikich:
- Add remove option to drm
    - Add mfu_walk_opts_t struct to hold
      desired functionality during walk
    - Pass walk_opts struct to
      mfu_flist_walk_param_paths instead of
      individual parameters (dir_perms & remove)
    - Delete files in mfu_flist_walk for the
      walk with stat and without
    - Delete directories by level in drm

### DIFF
--- a/src/common/mfu_flist.c
+++ b/src/common/mfu_flist.c
@@ -50,6 +50,31 @@
 #include <sys/ioctl.h>
 #endif
 
+/* return a newly allocated walk_opts structure, set default values on its fields */
+mfu_walk_opts_t* mfu_walk_opts_new(void)
+{
+    mfu_walk_opts_t* opts = (mfu_walk_opts_t*)MFU_MALLOC(sizeof(mfu_walk_opts_t));
+
+    /* Don't set permissions on walk by default */
+    opts->dir_perms = 0;
+
+    /* Remove files in walk by default */
+    opts->remove = 0;
+
+    /* Don't stat files in walk by default */
+    opts->use_stat = 1;
+
+    return opts;
+}
+
+void mfu_walk_opts_delete(mfu_walk_opts_t** popts)
+{
+  if (popts != NULL) {
+    mfu_walk_opts_t* opts = *popts;
+    mfu_free(popts);
+  }
+}
+
 /****************************************
  * Functions on types
  ***************************************/
--- a/src/common/mfu_flist.h
+++ b/src/common/mfu_flist.h
@@ -144,7 +144,7 @@ int mfu_perms_parse(const char* modestr, mfu_perms** pperms);
  * execute for the "all" bits as well because those can also turn on the user's
  * read and execute bits, sets dir_perms to 1 if "rx" should be set on directories
  * and set to 0 otherwise */
-void mfu_perms_need_dir_rx(const mfu_perms* head, int* dir_perms);
+void mfu_perms_need_dir_rx(const mfu_perms* head, mfu_walk_opts_t* walk_opts);
 
 /* free the permissions linked list allocated in mfu_perms_parse,
  * sets pointer to NULL on return */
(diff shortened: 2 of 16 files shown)


src/common/mfu_flist.h
```diff
@@ -444,8 +444,14 @@ int mfu_flist_copy(
...
-/* create all directories in flist */
-void mfu_flist_mkdir(mfu_flist flist);
+/* allocate a new mfu_walk_opts structure,
+ * and set its fields with default values */
+mfu_walk_opts_t* mfu_walk_opts_new(void);
+
+/* free object allocated in mfu_walk_opts_new */
+void mfu_walk_opts_delete(mfu_walk_opts_t** opts);
+
+/* create all directories in flist */ void mfu_flist_mkdir(mfu_flist flist);
```
adammoody: Lose a newline here?

src/common/mfu_flist.h
```diff
@@ -169,27 +169,27 @@ void mfu_flist_free(mfu_flist* flist);
  * set directory permission bits in order to walk into a directory,
  * whose permission bits are otherwise restrictive (e.g., useful for recursive unlink) */
 void mfu_flist_walk_path(
-    const char* path,    /* IN  - path to be walked */
-    int use_stat,        /* IN  - whether to stat each item (1) or not (0) */
-    int dir_permissions, /* IN  - whether to set directory permission bits (1) or not (0) during walk */
-    mfu_flist flist      /* OUT - flist to insert walked items into */
+    const char* path,         /* IN  - path to be walked */
+    int use_stat,             /* IN  - whether to stat each item (1) or not (0) */
```
adammoody: While we're at it, I think it'd be nice to roll this "use_stat" flag into the walk options structure as well.

src/common/mfu_flist_walk.c
```diff
@@ -656,9 +669,6 @@ static void walk_stat_process(CIRCLE_handle* handle)
 
     /* TODO: filter items by stat info */
 
-    /* record info for item in list */
-    mfu_flist_insert_stat(CURRENT_LIST, path, st.st_mode, &st);
-
```
adammoody: Can't think of anything off the top of my head, but I think it'd be good to keep the ordering here where we insert the directory into the list before we recurse into it.

adammoody:
Thanks for putting this together, @dsikich .

src/common/mfu_flist.h
dsikich: @adammoody use stat by default?

adammoody: Good question, let's turn it on by default.

dsikich:
@adammoody I updated the code based on your comments. I also moved the `--aggressive` remove option to drm instead of having it in dwalk.

src/drm/drm.c
```diff
@@ -21,10 +21,6 @@
 #include "dtcmp.h"
 #include "mfu.h"
 
-/* TODO: change globals to struct */
-static int walk_stat = 0;
```
adammoody: With drm, it looks like we were avoiding the stat call by default.  I think it would be good to keep this, since that's easier on the file system.  Although I realize that means the --lite option is pointless, and we don't have an option to turn stat back on right now.

dsikich: @adammoody I updated to avoid stat by default, and I turn on the stat for the aggressive option. Yeah, the lite option is pointless now.

src/common/mfu_flist.c
```diff
@@ -50,6 +50,31 @@
...
+/* return a newly allocated walk_opts structure, set default values on its fields */
+mfu_walk_opts_t* mfu_walk_opts_new(void)
+{
+    mfu_walk_opts_t* opts = (mfu_walk_opts_t*)MFU_MALLOC(sizeof(mfu_walk_opts_t));
+
+    /* Don't set permissions on walk by default */
+    opts->dir_perms = 0;
+
+    /* Remove files in walk by default */
+    opts->remove = 1;
```
adammoody: Let's keep the defaults in opts_new() to be (dir_perms=0, remove=0, use_stat=1) since that's want most tools want.  Then in the drm tool, let's set opts->use_stat = 0 right after making the opts_new() call.

dsikich: @adammoody ok, updated.

PR #206, opened 2018-12-18
Add support for copying GPFS ACLs

sjtbham:
We were interesting in using mpifileutils for copying data from one GPFS file-system to another however needed ACL support.

This patch adds support to enable `--enable-gpfs` to configure and adds calls to GPFS API `gpfs_getacl` and `gpfs_putacl` into `mfu_copy_permissions`.

Note that I have added include for `config.h` which was previously missing. There are bunch of `#ifdef LUSTRE_SUPPORT` lines already in the code, so adding config.h may have enabled some LUSTRE functionality? (I have no lustre systems to check this).

IBM docs on the API are at https://www.ibm.com/support/knowledgecenter/en/STXKQY_5.0.1/com.ibm.spectrum.scale.v5r01.doc/bl1adm_gpfs_getacl.htm

Code checks that when calling gpfs_getacl, 0 is returned - i.e. valid ACL had been collected, before attempting to write to dest_file. This means reading from a non-gpfs file-system there is no issue as the ACL will fail to read. On writing to a non GPFS file-system a warning will be produced as the ACL cannot be written.

I've done some basic testing of files and directories with both directly placed and inherited ACLs with both dcp and dsync.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -69,6 +69,9 @@ PKG_CHECK_MODULES([libcircle], libcircle)
 # Check for libdtcmp.
 X_AC_DTCMP
 
+# Check for GPFS
+X_AC_GPFS
+
 # Check for Lustre
 X_AC_LUSTRE
 
--- a/m4/lx_find_gpfs.m4
+++ b/m4/lx_find_gpfs.m4
@@ -1,0 +1,39 @@
+dnl #
+dnl # Determines if GPFS API is installed.
+dnl #
+AC_DEFUN([X_AC_GPFS], [
+	AC_ARG_ENABLE(gpfs,
+		AC_HELP_STRING([--enable-gpfs],
+		[enable GPFS/Spectrum Scale support]),
+		[], [enable_gpfs=no])
+
+	AS_CASE(["x$enable_gpfs"],
+		["xcheck"],
+		[gpfs_found="yes"],
+		["xyes"],
+		[gpfs_found="yes"],
+		["xno"],
+		[gpfs_found="no"],
+		[AC_MSG_ERROR([Unknown option $enable_gpfs])])
+
+
+	AS_IF([test "x$enable_gpfs" != "xno"], [
+		AC_CHECK_HEADERS([gpfs.h],
+			[], [gpfs_found="no"], [])
+
+		AS_IF([test "x$gpfs_found" != "xno"], [
+			AC_DEFINE([GPFS_SUPPORT], [1],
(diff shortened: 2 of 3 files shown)


dsikich:
@sjtbham After the config.h was added this does cause errors when copying on NFS as it turns on LUSTRE_SUPPORT and then it attempts to run code that it shouldn't.

adammoody:
Thanks for your contribution, @sjtbham .  GPFS is certainly a system that we'd like to support.

FYI, it's on the TODO list to automatically detect the type of underlying file systems (NFS vs Lustre vs GPFS) and then enable the correct code paths.  This will need to be done for both source and destination, since they could be different.  Until that's done, we can consider falling back to having the user declare the type of file system via options if necessary.  @dsikich and I will look at the NFS vs Lustre issue in more detail here.

sjtbham:
I guess maybe the code got refactored at some point as I couldn't see how the Lustre code was enabled otherwise.

Note I've used `gpfs_opaque_acl_t` to read and write directly to GPFS, if the code is changed at some point to auto detect the fstype, it might be worth looking at using `gpfs_acl_t` instead as this might enable a copy to/from NFSv4 and GPFS (don't know what ACL support is provided in Lustre). Though being sure any kind of conversion was correct would need a lot of testing!

adammoody:
@sjtbham , would you please update each of your commit messages to include the Signed-off-by tag and then force push back to this PR?

https://github.com/hpc/mpifileutils/blob/master/.github/CONTRIBUTING.md#contributors-declaration

This is to ensure that the incoming code is under BSD and you have the right to contribute it as such.  Thanks.

dsikich:
@sjtbham I want to get your PR merged in this week. Other than changing the signed-off-by tag in the commit message it looks good. I want to wait until I fix the issues/errors this causes with LUSTRE_SUPPORT before merging it into master.

sjtbham:
@adammoody I've managed to update the last commit, but struggling with git to let me update the previous commits ... when I push it claims all up to date!

I was going to say the contributors link in the readme was broken, but it like someone fixed that last night!

sjtbham:
OK . ... managed to fight with git and get my the correct sign off at the end of each commit now!

adammoody:
Thanks @sjtbham .

dsikich:
@sjtbham Thanks for your contribution. I just submitted a fix to the issue we had when LUSTRE_SUPPORT was available but NFS was being used. I am going to merge this, and we can merge the CMAKE PR after we update/add a GPFS check.

PR #208, opened 2018-12-19
CMake

gonsie:
I've added cmake! I can also delete the autotools stuff in this PR, if we want an abrupt transition. Once this is in, I'll have to fix the spack package.


adammoody:
Thanks @gonsie !  I also vote for moving sooner rather than later.  This will improve the build process for people.

adammoody:
dreln is merged

PR #209, opened 2018-12-19
fix broken link to contributor docs

adammoody:
- Fix broken link to contributor document
- Provide example Signed-off-by text


gonsie:
Out of curiosity, why are we enforcing sign-off now? What do we gain by this?

adammoody:
It had been there for a while.  It's meant to ensure that the code coming in from external contributions is under BSD and the author has the right to give it to us.  We don't need to enforce it for internal contributions, although it's ok to have it.

PR #210, opened 2018-12-19
Doc cleanup

gonsie:
Several items needed to be cleaned up:

- needed to clean up some things from moving dfind out of experimental
- created dbz2 rst, from the readme
- `man/` directory was re-generated


doc/rst/index.rst
```diff
@@ -86,7 +87,7 @@ Man Pages
    dstripe.1
    dsync.1
    dwalk.1
-   experimental/dfind.1
+   dfind.1
```
adammoody: Let's move this up in the list, too, to keep it alphabetical.

PR #211, opened 2018-12-19
dsync: tweak new --delete option

adammoody:


src/dsync/dsync.c
```diff
@@ -935,7 +935,7 @@ static int dsync_sync_files(strmap* src_map, strmap* dst_map,
 
     /* delete files from destination if needed */
     uint64_t remove_size = mfu_flist_global_size(dst_remove_list);
-    if (remove_size > 0 && options.delete) {
+    if (remove_size > 0) {
```
adammoody: We want to delete files here even if --delete it not thrown, because it handles files that are in dest and source but need to be overwritten, i.e., deleted from dest and then copied fresh from source.

Issue #177, opened 2018-08-30
libmfu: read/write files via POSIX I/O since external32 MPI I/O is often broken


adammoody:

gonsie:
Fixed by #215

PR #215, opened 2018-12-28
Convert integers to network order when writing flist to files

adammoody:
The external32 format across MPI libraries is broken and non-standard when dealing with host-to-network order conversion of integers.  Until this is fixed, convert integers to network order manually and write everything with MPI_BYTE instead of using more complicated datatypes.

This breaks backwards compatibility with MFU flist binary file format in v0.8.  That is files written in v0.8 cannot be read in v0.9.

This solves https://github.com/hpc/mpifileutils/issues/177


adammoody:
In short, there are currently the following changes:
1) The file version has been bumped from "3" to “4”
2) All integers are converted to big endian (network byte order)
3) The walk_start and walk_end fields have been dropped from the header.  In v3, these were two consecutive unit64_t fields that immediately followed the file version field

adammoody:
I have verified that MVAPICH2-2.3 and Open MPI 2.0.0 now produce identical flist output files with this fix, and IBM MPI can read files written by MVAPICH2.

dsikich:
@adammoody Is this still a work in progress?

adammoody:
One thing I don't have with this is a mechanism to check for byteswap.h on the system, and a possible replacement if it's not there.  It could bite folks who are missing this file, or if it is not in the default include path.  We could probably merge this PR and deal with it in another PR.

gonsie:
We can check for byteswap.h in CMake. SCR's cmake does this. If byteswap.h is found, then the HAVE_BYTESWAP_H macro is defined.

adammoody:
@gonsie , thanks, let me push the cmake check as another commit and have you review it.

adammoody:
@gonsie , this look to be correct? https://github.com/hpc/mpifileutils/pull/215/commits/faf406e0a3a5cd297427b47ac2cb74558f52c986

gonsie:
Will the #error pragma actually fail at compile time? Otherwise, you might want an else statement in the CMake to error out when byteswap isn't found.

CMakeLists.txt
```diff
@@ -35,6 +35,13 @@ ENDIF(ENABLE_GPFS)
 
 OPTION(ENABLE_EXPERIMENTAL "Build experimental tools" OFF)
 
+## HEADERS
+INCLUDE(CheckIncludeFile)
+CHECK_INCLUDE_FILE(byteswap.h HAVE_BYTESWAP_H)
+if(HAVE_BYTESWAP_H)
+  ADD_DEFINITIONS(-DHAVE_BYTESWAP_H)
+ENDIF(HAVE_BYTESWAP_H)
```
gonsie:
```suggestion
ELSE(HAVE_BYTESWAP_H)
  MESSAGE(SEND_ERROR "byteswap.h is required")
ENDIF(HAVE_BYTESWAP_H)
```

PR #218, opened 2019-01-10
spack build deps instructions

gonsie:
Spack views will soon get a feature upgrade, but in the mean time, these are the instructions for building the dependencies.


adammoody:
Thanks @gonsie

PR #223, opened 2019-01-25
touch ups for v0.9 release

adammoody:
- add dreln to cmake
- install man pages with crake
- fix most cmake compiler warnings


gonsie:
I still need to clean up the cmake for this branch, but I should get that done later today.

dsikich:
@adammoody @gonsie did you both update the dbz2 .rst? looks like there is a conflict

PR #231, opened 2019-02-08
update old dcp1 tests

dsikich:
Updated dcp1 tests scripts to use proper arguments for mpirun as
well as dcp1. The scripts were using many arguments that did not
exist anymore or had been changed/updated, which is why the tests
were failing. All tests for dcp1 pass now by running ./test_all
and setting two env variables.

### DIFF
--- a/test/legacy/dcp1_tests/test_all.sh
+++ b/test/legacy/dcp1_tests/test_all.sh
@@ -9,14 +9,15 @@
 # Turn this on if you want output from each test printed out.
 DEBUG=1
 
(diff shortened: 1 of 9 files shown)


gonsie:
would it be possible to add these to the automatic travis testing? if it ends up involving too much (like setting up nose), we should create an issue.

dsikich:
@gonsie I just updated the old dcp bash scripts, which are not being called by nose. I could call the test_all script for the dcp1 tests in .travis.yaml? But it isn't part of nose currently.

PR #233, opened 2019-02-23
BLT

gonsie:
After trying a billion combinations, I could not get CMake to link the tools in the right way during the install step. Either libmfu was linked (at the install location) or dtcmp and libcircle were linked, with libmfu at the build location. (side note: spack somehow makes it do the right thing, go figure). 

BUT, [BLT](https://github.com/llnl/blt) has some magic to actually do the proper linking! I've added only the bare bones BLT files we need.


dsikich:
@gonsie looks like the build is failing says it cannot find 'mfu.h' in travis output?

adammoody:
Solved rpath in cmake, so I think we can close this

PR #234, opened 2019-02-25
dbz2: add static round-robin work balancing

adammoody:
This adds an alternative methods for load-balancing work for compression/decompression, in which each process is assigned the set of blocks based on its rank.


adammoody:
This is close to ready, but I'd like to let it sit for a few days in case I think of more things before we merge.

PR #236, opened 2019-02-26
make tools verbose by default, and add quiet option

dsikich:
Make tools verbose by default, and add quiet option to turn off output if desired. This did not apply to dbcast and both dfilemaker tools as neither are currently using verbose output.

    - Make tools verbose by default
    - add quiet option
    - pass mfu_debug_level to mfu_flist_unlink
    - update docs

### DIFF
--- a/doc/rst/dchmod.1.rst
+++ b/doc/rst/dchmod.1.rst
@@ -62,6 +62,10 @@ OPTIONS
    directory tree, and the number of files the command operated on, and
    the files/sec rate for each of those.
 
+.. option:: -q, --quiet
+
+   Run tool silently. No output is printed.
+
 .. option:: -h, --help
 
    Print the command usage, and the list of options available.
--- a/doc/rst/dcmp.1.rst
+++ b/doc/rst/dcmp.1.rst
@@ -42,6 +42,10 @@ OPTIONS
    command. Files walked, started, completed, seconds, files, bytes
    read, byte rate, and file rate.
 
+.. option:: -q, --quiet
+
+   Run tool silently. No output is printed.
+
 .. option:: -l, --lite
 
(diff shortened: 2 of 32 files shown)


adammoody:
Let's define a MFU_LOG_NONE = 0 value here that we can use for silent:

https://github.com/hpc/mpifileutils/blob/97cabf759e2b4cbf3dad9984154897339e36b574/src/common/mfu_util.h#L64

adammoody:
@dsikich As you mentioned, it seems to me that mfu_flist_unlink should be using the global mfu_debug_level variable so that we don't need to pass it in.  I can help debugging for that if you'd like.

gonsie:
Thanks for making sure the documentation RST files are updated :smile:

dsikich:
@adammoody yeah, i'm not sure why the mfu_debug_level is not working without passing it into the unlink function in drm. My first thought is that maybe in one of the header files that global is getting over written.

dsikich:
@adammoody the mfu_debug_level variable is definitely getting over written. Right before the mfu_flist_unlink call the level is 5, and in the mfu_flist_unlink call the value is 2 (the default value). The default is set in mfu_util.h which is added as one of the headers for mfu_flist_remove.c.

dsikich:
> Let's define a MFU_LOG_NONE = 0 value here that we can use for silent:
> 
> [mpifileutils/src/common/mfu_util.h](https://github.com/hpc/mpifileutils/blob/97cabf759e2b4cbf3dad9984154897339e36b574/src/common/mfu_util.h#L64)
> Line 64 in [97cabf7](/hpc/mpifileutils/commit/97cabf759e2b4cbf3dad9984154897339e36b574)
>  MFU_LOG_FATAL   = 1,

@adammoody this has been updated.

Issue #205, opened 2018-12-15
dwalk: Add dynamically generated size histogram


daltschu22:
I believe the `--distribution` flag for size should generate a histogram automatically if the size distributions are not specified. 

For example, it could dump out a histogram of say, powers of 2.
I have a python tool that does filesystem analytics and it does this sort of a thing. So you just run the tool using the `--size-histogram` flag and it pops out a nice human friendly dictionary like:
      ```  'SizeHistogram': {   '1024KiB': 33,
                         '128KiB': 441,
                         '16KiB': 2292,
                         '16MiB': 2,
                         '1GiB': 2,
                         '1KB': 6142,
                         '1KiB': 1764,
                         '1MiB': 23,
                         '256KiB': 175,
                         '2KiB': 2853,
                         '2MiB': 13,
                         '32KiB': 1517,
                         '32MiB': 1,
                         '4KiB': 3434,
                         '4MiB': 6,
                         '512KiB': 100,
                         '64KiB': 943,
                         '8KiB': 2556,
                         '8MiB': 3},```

This dictionary grows dynamically with each new file no matter what the size is (Entries to the dictionary are added automatically when a file of that general size is found, and any new files have the count incremented). 

This makes it much easier to generate size histograms, as you dont need to think too much about the size ranges that you want if you just want a general overview.

dsikich:
@daltschu22 Thanks for the feedback. These are not all in the same units. How is this program deciding whether to use say KB or MB as the units?

daltschu22:
Basically the way it works is it gets the size of a file in bytes.
Then you convert that size into a human readable format. 
My code for that is here: 
https://github.com/daltschu22/crawlinator/blob/master/crawlinator.py#L152-L178

If that file is say, 16 Gibibytes, then it uses the "GiB" unit range, and then it finds the closest power of 2 to that file size. So 16 is already a power of 2 it adds a counter to the '16GiB' entry in the dictionary.
If you had say 

If that file is 150 Kibibytes, then round down to 128KiB and add a counter in that section of the dictionary.

The histogram generation is here:
https://github.com/daltschu22/crawlinator/blob/master/crawlinator.py#L132-L150

Does that make sense? I can explain further if need be.

dsikich:
@daltschu22 Yep, makes sense. Thanks.

daltschu22:
Wonderful work, thanks so much!

PR #237, opened 2019-03-01
dwalk: add default file histogram

dsikich:
When file_histogram option is used this creates a file histogram of all walked files. This option does not require any range inputs. It will give the user a count of files in each bin. The bins are created based on the max file size. So if the max file size is between 2^20 - 2^30, then this is the last bin. The bins start with the first bin for 0 byte files, then goes up from 2^10, 2^20, 2^30, etc. The bins are created dynamically based on the max file size.


dsikich:
@gonsie @adammoody I have the implementation mostly good to go for the file histogram in dwalk. I am still working out the interface. I am/was working off the interface used for the distribution option. It looks like this:

```bash
quartz1148{sikich1}250: dwalk -f pmix
[2019-02-28T17:04:41] [0] [/g/g0/sikich1/mpifileutils/src/common/mfu_flist.c:1758] Items: 7881
[2019-02-28T17:04:41] [0] [/g/g0/sikich1/mpifileutils/src/common/mfu_flist.c:1759]   Directories: 1037
[2019-02-28T17:04:41] [0] [/g/g0/sikich1/mpifileutils/src/common/mfu_flist.c:1760]   Files: 6664
[2019-02-28T17:04:41] [0] [/g/g0/sikich1/mpifileutils/src/common/mfu_flist.c:1761]   Links: 180
[2019-02-28T17:04:41] [0] [/g/g0/sikich1/mpifileutils/src/common/mfu_flist.c:1779] Data: 152.564 MB (23.442 KB per file)
Range	Number
[0 - 0.000 B]	148
[1.000 B - 1.000 KB]	1856
[1.001 KB - 2.000 KB]	686
[2.001 KB - 4.000 KB]	1734
[4.001 KB - 8.000 KB]	754
[8.001 KB - 16.000 KB]	813
[16.001 KB - 32.000 KB]	781
[32.001 KB - 64.000 KB]	715
[64.001 KB - 128.000 KB]	306
[128.001 KB - 256.000 KB]	44
[256.001 KB - 512.000 KB]	24
[512.001 KB - 1.000 MB]	6
[1.000 MB - 2.000 MB]	6
[2.000 MB - 4.000 MB]	6
[4.000 MB - 8.000 MB]	1
[8.000 MB - 16.000 MB]	1
[16.000 MB - 32.000 MB]	0
[32.000 MB - 64.000 MB]	0
[64.000 MB - 128.000 MB]	0
[128.000 MB - 256.000 MB]	0
[256.000 MB - 512.000 MB]	0
[512.000 MB - 1.000 GB]	0
[1.000 GB - 2.000 GB]	0
[2.000 GB - 4.000 GB]	0
[4.000 GB - 8.000 GB]	0
[8.000 GB - 16.000 GB]	0
[16.000 GB - 32.000 GB]	0
[32.000 GB - 64.000 GB]	0
[64.000 GB - 128.000 GB]	0
[128.000 GB - 256.000 GB]	0
[256.000 GB - 512.000 GB]	0
[512.000 GB - 1.000 TB]	0
[1.000 TB - MAX]	0
```

dsikich:
@adammoody @gonsie looking for suggestions on how to output this file info the best way? Any suggestions?

gonsie:
At first I thought you had too many buckets... but wow, there an a bunch of small files it a variety of sizes! So I think the detail is good for now (a future feature might be to make the histogram buckets adaptive). 

You could also choose to always print 11 buckets, with each bin showing i*(MAX/10) for i = 0...10

In any case, I would change the output so that it’s pretty printed, something like:

```
[            0  B ] 148
[   1  B -   1 KB ] 1856
[   1 KB -   2 KB ] 686
...
[ 512 KB -   1 MB ] 73 
...
[ 128 MB - 512 MB ] 234
```

adammoody:
Let's use mathematical range syntax to help users see which bin counts the boundary values, e.g.,
```
[1  B - 1 KB)
[1 KB - 2 KB)
```
rather than the double [ ] brackets.

This would mean files of exactly 1KB are included in the second bin, but excluded from the first.

dsikich:
@gonsie so i can print 10 buckets based on the max size, would you rather have that or all of the bins up to 2^40?

gonsie:
After a discussion with @dsikich, I think we should stick to orders of magnitude buckets. There is another flag to get the exactly binning a user might want. 

Also, this is related to the discussion in olcf/pcircle#45

gonsie:
once merged, this fixes #205

dsikich:
@gonsie @adammoody new output creates bins by default dynamically based on max file size:

```bash
quartz1922{sikich1}295: dwalk -f pmix
[2019-03-20T12:22:46] Walking /g/g0/sikich1/install/bin/pmix
[2019-03-20T12:22:51] Walked 7881 items in 5.231172 seconds (1506.545874 files/sec)
[2019-03-20T12:22:51] Items: 7881
[2019-03-20T12:22:51]   Directories: 1037
[2019-03-20T12:22:51]   Files: 6664
[2019-03-20T12:22:51]   Links: 180
[2019-03-20T12:22:51] Data: 152.564 MB (23.442 KB per file)
Max File Size: 8.346 MB
Range                     Number
[  0.000 B   - 0.000 B  ] 148
[  1.000 B   - 1.000 KB ] 1856
[  1.001 KB  - 1.000 MB ] 5863
[  1.000 MB  - 1.000 GB ] 14
```

dsikich:
@gonsie @adammoody i see i need to update the bins to use the mathematical notation, and i need to update the docs for the option. Let me know if you see anything else that needs updating. If not, i will fix those today and remove the WIP label.

adammoody:
Thanks, @dsikich . On the math range notation comment, I see this new format has 3 decimal places that are new, which may be verbose, but otherwise ok.  However, I also see that sometimes, we end up with a .001 ending:

[  1.000 B   - 1.000 KB ] 1856
[  1.001 KB  - 1.000 MB ] 5863

We should either try to fix that so that it starts where the line above left off, or we should be careful with the range notation, since my comment there was assuming the endpoints are exact.

adammoody:
Oh, I guess those extra decimal places were there originally.  Anyway, it'd still be good to make those the same (end point of previous line matches start point of current line) and then use inclusive/exclusive range notation.

dsikich:
@adammoody yeah I forgot to comment on that. I can update to start the range where the other left off in the printing of the range.

dsikich:
@adammoody @gonsie ok. I've updated those issues. This should be ready for review.

dsikich:
updated output with range notation:

``` bash
quartz1922{sikich1}340: dwalk -f pmix
[2019-03-20T15:19:19] Walking /g/g0/sikich1/install/bin/pmix
[2019-03-20T15:19:27] Walked 7881 items in 8.370324 seconds (941.540626 files/sec)
[2019-03-20T15:19:27] Items: 7881
[2019-03-20T15:19:27]   Directories: 1037
[2019-03-20T15:19:27]   Files: 6664
[2019-03-20T15:19:27]   Links: 180
[2019-03-20T15:19:27] Data: 152.564 MB (23.442 KB per file)
Max File Size: 8.346 MB
Range                     Number
[  0.000 B   - 0.000 B  ) 148
[  0.000 B   - 1.000 KB ) 1856
[  1.000 KB  - 1.000 MB ) 5863
[  1.000 MB  - 1.000 GB ) 14
```

adammoody:
Looks good.  Thanks.  For that very first line, can you make the endpoint 1.000 B.  Currently, we have a NULL set for the range [0, 0)

dsikich:
@adammoody not sure what you mean?

adammoody:
I was suggesting that we change:
```
[  0.000 B   - 0.000 B  ) 148
[  0.000 B   - 1.000 KB ) 1856
```
to:
```
[  0.000 B   - 1.000 B  ) 148
[  1.000 B   - 1.000 KB ) 1856
```
I think the mathematical range of [0, 0) is the NULL set if anything, whereas [0, 1) means file sizes of exactly 0 bytes.

dsikich:
@adammoody i can change this in the default case, but if the user specifies a '0' bucket then that is what the first line will look like. I can update the ranges for the default case but have no control over the range in the first bucket if the user specifies '0'. Does that make sense?

dsikich:
@adammoody this has been updated

```bash
quartz2306{sikich1}406: dwalk -f pmix
[2019-04-08T18:57:33] Walking /g/g0/sikich1/install/bin/pmix
[2019-04-08T18:57:38] Walked 7881 items in 4.329665 seconds (1820.233328 files/sec)
[2019-04-08T18:57:38] Items: 7881
[2019-04-08T18:57:38]   Directories: 1037
[2019-04-08T18:57:38]   Files: 6664
[2019-04-08T18:57:38]   Links: 180
[2019-04-08T18:57:38] Data: 152.564 MB (23.442 KB per file)
Max File Size: 8.346 MB
Range                                             Number
[         0.000   B          -        1.000   B ) 148
[         1.000   B          -        1.000   KB) 1856
[         1.000   KB         -        1.000   MB) 5863
[         1.000   MB         -        1.000   GB) 14
```

dsikich:
@adammoody I also changed the help menu output for dwalk. Let me know if this is good to go.

gonsie:
Can you make sure this branched is rebased to master? It looks like the changes you made in #236 might have been reverted? Or maybe I’m reading the diff wrong.

dsikich:
@gonsie yep, rebased it.

src/dwalk/dwalk.c
```diff
@@ -62,71 +114,83 @@ static int print_flist_distribution(struct distribute_option *option, mfu_flist*
...
+         for (int i = 0; i <= separators; i++) {
+             printf("%-10s", "[");
              if (i == 0) {
-                 printf("0");
+                 printf("%-7.3lf %-10s", 0.000, "B");
              } else {
-                 printf("%"PRIu64, option->separators[i - 1] + 1);
+                 mfu_format_bytes((uint64_t)option->separators[i - 1],
+                                  &size_tmp, &size_units);
+                 printf("%-7.3lf %-10s", size_tmp, size_units);
```
adammoody: Is this using 10 characters?  Seems like a lot if the units string is only 2 characters long.  Trying to trim extra spacing.

src/dwalk/dwalk.c
```diff
@@ -62,71 +114,83 @@ static int print_flist_distribution(struct distribute_option *option, mfu_flist*
...
              }
- 
-             printf("~");
- 
-             if (i == option->separator_number) {
-                 number = disttotal[i];
-                 printf("MAX]\t%"PRIu64"\n", number);
+
+             printf("%-10s", " - ");
```
adammoody: can we just use printf(" - ") here?

src/dwalk/dwalk.c
```diff
@@ -62,71 +114,83 @@ static int print_flist_distribution(struct distribute_option *option, mfu_flist*
...
-                 printf("%"PRIu64"]\t%"PRIu64"\n",
-                 option->separators[i], number);
+                if (i == separators) {
+                    number = disttotal[i];
+                    printf("%-10s) %-3"PRIu64"\n", "MAX", number);
+                } else {
+                    number = disttotal[i];
+                    mfu_format_bytes((uint64_t)option->separators[i],
+                              &size_tmp, &size_units);
+                    printf("%-7.3lf %-2s) %-3"PRIu64"\n", size_tmp, size_units, number);
```
adammoody: Does setting this to 3 truncate numbers longer than 3 characters?  Anyway, I don't think we need to limit it, so we could just drop the 3.

src/dwalk/dwalk.c
```diff
@@ -62,71 +114,83 @@ static int print_flist_distribution(struct distribute_option *option, mfu_flist*
...
          /* number of files in a bin */
          uint64_t number;
-         printf("Range\tNumber\n");
-         for (int i = 0; i <= option->separator_number; i++) {
-             printf("[");
+         double size_tmp;
+         const char* size_units;
+         printf("%-49s %-10s\n", "Range", "Number");
+         for (int i = 0; i <= separators; i++) {
+             printf("%-10s", "[");
```
adammoody: Another 10 here, maybe just replace with ```printf("[ ");```

src/dwalk/dwalk.c
```diff
@@ -62,71 +114,83 @@ static int print_flist_distribution(struct distribute_option *option, mfu_flist*
...
     /* Print the file distribution */
     if (rank == 0) {
          /* number of files in a bin */
          uint64_t number;
-         printf("Range\tNumber\n");
-         for (int i = 0; i <= option->separator_number; i++) {
-             printf("[");
+         double size_tmp;
+         const char* size_units;
+         printf("%-49s %-10s\n", "Range", "Number");
```
adammoody: After dropping all of the 10's below, I think this 49 would become:
```
[ 0000.00 KB - 0000.00 KB ]
```
I count 27 chars long from bracket to bracket, inclusive.

src/dwalk/dwalk.c
adammoody: I see there are examples with 4 characters in the numbers, so I guess it doesn't limit it.  Let's drop it anyway, though, unless it's there for a reason.

PR #248, opened 2019-04-02
Changed line 1704 to say items instead of files.

gonsie:


adammoody:
Thanks @levatin1 and @gonsie .

PR #250, opened 2019-04-05
move source for experimental tools to src

adammoody:
This moves the source for experimental tools to src/ and drops the experimental/ subdirectory.  We still only build the experimental tools if requested.  Moving them simplifies the process of moving a tool from experimental to released.


dsikich:
@adammoody makes sense to me

adammoody:
Thanks @dsikich .

Issue #229, opened 2019-02-06
dwalk: document --text option


gonsie:
Apparently dwalk has an undocumented option `--text` to generate a human readable output file. We should document this.

adammoody:
Related: https://github.com/hpc/mpifileutils/issues/129

PR #254, opened 2019-04-15
documentation for dwalk --text option

gonsie:


adammoody:
Thanks, @gonsie !  Let's also add a line to the print usage section: https://github.com/hpc/mpifileutils/blob/9e2bdfd4ad816696c37e5ec0436028c6ba48710b/src/dwalk/dwalk.c#L241

adammoody:
Thanks, @gonsie !

src/dwalk/dwalk.c
```diff
@@ -238,7 +238,8 @@ static void print_usage(void)
     printf("\n");
     printf("Options:\n");
     printf("  -i, --input <file>                      - read list from file\n");
-    printf("  -o, --output <file>                     - write processed list to file\n");
+    printf("  -o, --output <file>                     - write processed list to FILE in binary format\n");
```
adammoody: To be consistent, let's use "file" instead of "FILE" in the print_usage.  I do prefer FILE, actually, since that sets it off more that it's a placeholder.  However, then we should update the rest of the print_usage to use a style of all caps.

What do others prefer?

gonsie: I think we are more consistent in the man pages. I'll revert to "file" here and leave it as a future exercise to fix-up the usage documentation.

gonsie:
fixes #229

PR #258, opened 2019-04-17
Dsync2

levatin1:
Please test new dsync.  Thanks.


gonsie:
These commits are captured in #261.

PR #266, opened 2019-04-22
dsync: add link-dest support

KnightKu:
-l, --link-dest=DIR   - hardlink to files in DIR when unchanged

dsync -l /link_dest_dir/ /src/ /dst/

Signed-off-by: Gu Zheng <gzheng@ddn.com>

### DIFF
--- a/src/common/mfu_flist.h
+++ b/src/common/mfu_flist.h
@@ -448,6 +448,16 @@ int mfu_flist_copy(
     mfu_copy_opts_t* mfu_copy_opts  /* IN - options to be used during copy */
 );
 
+/* link items in list from source paths to destination,
+ * each item in source list must come from the
+ * source path, returns 0 on success -1 on error */
+int mfu_flist_hardlink(
+    mfu_flist src_link_list,         /* IN - flist providing source items */
+    const mfu_param_path* srcpath,   /* IN - the source patht */
+    const mfu_param_path* destpath,  /* IN - destination path */
+    mfu_copy_opts_t* mfu_copy_opts   /* IN - options to be used during copy */
+);
+
 /* allocate a new mfu_walk_opts structure,
  * and set its fields with default values */
 mfu_walk_opts_t* mfu_walk_opts_new(void);
--- a/src/common/mfu_flist_copy.c
+++ b/src/common/mfu_flist_copy.c
@@ -1129,6 +1129,44 @@ static int mfu_create_file(mfu_flist list, uint64_t idx,
     return rc;
 }
 
+/* creates hardlink in destpath for specified file, identifies source path
+ * returns 0 on success and -1 on error */
+static int mfu_create_hardlink(mfu_flist list, uint64_t idx,
+        const mfu_param_path* srcpath,
+        const mfu_param_path* destpath, mfu_copy_opts_t* mfu_copy_opts)
+{
+    /* assume we'll succeed */
+    int rc = 0;
+
+    /* get source name */
+    const char* src_path = mfu_flist_file_get_name(list, idx);
+
+    /* get destination name */
+    const char* dest_path = mfu_param_path_copy_dest(src_path, 1,
+            srcpath, destpath, mfu_copy_opts);
+
+    /* No need to copy it */
+    if (dest_path == NULL) {
+        return 0;
+    }
+
+    rc = mfu_hardlink(src_path, dest_path);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to create hardlink %s --> %s",
+                dest_path, src_path);
+        mfu_free(&dest_path);
+        return rc;
+    }
+
+    /* free destination path */
+    mfu_free(&dest_path);
+
+    /* increment our file count by one */
+    mfu_copy_stats.total_files++;
+
+    return rc;
+}
+
 /* progress message to print while creating files */
 static void create_progress_fn(const uint64_t* vals, int count, int complete, int ranks, double secs)
 {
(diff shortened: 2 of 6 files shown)


dsikich:
@KnightKu Thanks for the PR! I have not looked over it closely yet. Can you also update the documentation in doc/rst/dsync.1.rst with your new option. And some more description would be helpful about the option as well.

KnightKu:
Hi Danielle,
Yeah, right, I should add section about this option into Doc for this option, will update it soon, thanks.

adammoody:
Thanks, @KnightKu .  One of the changes in this PR that I'm stuck pondering is whether it's the right thing to do to add device and inode fields to our flist.  Currently, all of the fields in the flist are effectively file system agnostic, but these two fields tie the entry to a particular file system.  That in turn has implications in how we use the flist in other cases.  If we do want to go this route, we'll need to update our I/O and MPI routines that wrap flist operations, and our flist file format will need to change.  In that case, there might be other fields we should look at including, e.g., recording lustre striping info for a file, and someone recently asked about st_blocks.

I'm wondering if we can break this into two steps.  As a first step, can we bring hardlink support into dsync without changing the flist?  We could stat the files again within dsync to grab the device and inode info at the point where we do the check.  That's not as efficient as stat'ing the file once, but it should be functional.

KnightKu:
> Thanks, @KnightKu . One of the changes in this PR that I'm stuck pondering is whether it's the right thing to do to add device and inode fields to our flist. Currently, all of the fields in the flist are effectively file system agnostic, but these two fields tie the entry to a particular file system. That in turn has implications in how we use the flist in other cases. If we do want to go this route, we'll need to update our I/O and MPI routines that wrap flist operations, and our flist file format will need to change. In that case, there might be other fields we should look at including, e.g., recording lustre striping info for a file, and someone recently asked about st_blocks.
> 
Indeed, agreed with your thought.
> I'm wondering if we can break this into two steps. As a first step, can we bring hardlink support into dsync without changing the flist? We could stat the files again within dsync to grab the device and inode info at the point where we do the check. That's not as efficient as stat'ing the file once, but it should be functional.
Makes sense, let me rework it accordingly.
Thanks for your comments, Adam.

KnightKu:
Hi Danielle, Adam,
Could you please help to review it?

src/common/mfu_flist.h
```diff
@@ -377,6 +377,8 @@ uint64_t mfu_flist_file_get_ctime(mfu_flist flist, uint64_t index);
 uint64_t mfu_flist_file_get_ctime_nsec(mfu_flist flist, uint64_t index);
 uint64_t mfu_flist_file_get_size(mfu_flist flist, uint64_t index);
 uint64_t mfu_flist_file_get_perm(mfu_flist flist, uint64_t index);
+uint64_t mfu_flist_file_get_ino(mfu_flist flist, uint64_t index);
+uint32_t mfu_flist_file_get_dev(mfu_flist flist, uint64_t index);
```
adammoody: Let's drop these lines.

adammoody:
Thanks @KnightKu .  I took a quick first pass.  There is enough here that I need to more detailed look to wrap my head around it all.  Will try to get to that later this week.

In the meantime, we've merged some new commits that are causing a conflict with your PR.  Could you please rebase on the latest master to resolve those and force push?

Be aware that these other commits made changes to the dsync options and we also added periodic progress messages.

KnightKu:
> In the meantime, we've merged some new commits that are causing a conflict with your PR. Could you please rebase on the latest master to resolve those and force push?

Sure, will fix the tiny issue and rebase it against latest upstream.
>Be aware that these other commits made changes to the dsync options and we also added periodic progress messages.

Yeah, I see, thanks.

src/common/mfu_flist.h
KnightKu: Done

KnightKu:
ping

adammoody:
Thanks for the update @KnightKu .  Hope to review this in a day or two.

KnightKu:
gentle ping :)

src/dsync/dsync.c
```diff
@@ -1884,6 +2358,10 @@ static void dsync_option_fini(void)
         dsync_output_free(output);
     }
     assert(list_empty(&options.outputs));
+
+    if (options.link_dest != NULL) {
+        mfu_free(&options.link_dest);
```
adammoody: It's safe to call mfu_free without checking that the item is NULL, because mfu_free will do that check itself.  To simplify the code, let's remove the if check here.

src/dsync/dsync.c
```diff
@@ -2201,6 +2679,41 @@ static int dsync_option_output_parse(const char *option, int add_at_head)
     return ret;
 }
 
+/* link_dest doesn't support dirs cross filesystems */
+static int dsync_validate_link_dest(const char *link_dest, const char *dest)
```
adammoody: So that we don't have every process call stat, let's change the body of this routine so that only rank 0 does the work, and then MPI_Bcast the result to everyone.

src/dsync/dsync.c
```diff
@@ -2384,6 +2901,26 @@ int main(int argc, char **argv)
...
+    if (options.link_dest != NULL) {
+        rc = dsync_validate_link_dest(options.link_dest, destpath->path);
+        if (rc < 0) {
+             /* disable link_dest if not avliable */
+            if (rank == 0) {
+                MFU_LOG(MFU_LOG_ERR, "Invalid link_dest: [%s], ret:%d. Disabled link_dest",
+                        options.link_dest, rc);
+            }
+            mfu_free(&options.link_dest);
+            options.link_dest = NULL;
```
adammoody: mfu_free will also set the pointer back to NULL, so let's remove this line.

src/dsync/dsync.c
```diff
@@ -2410,42 +2953,75 @@ int main(int argc, char **argv)
...
+    if (options.link_dest != NULL) {
+        mfu_flist_free(&flist_link_dest);
+    }
 
     /* map each file name to its index and its comparison state */
     strmap* map1 = dsync_strmap_creat(flist3, path1);
     strmap* map2 = dsync_strmap_creat(flist4, path2);
+    strmap* map3 = NULL;
 
     /* compare files in map1 with those in map2 */
```
adammoody: This comment got separated from its code block, so let's move that back down to the strmap_compare() block.

src/dsync/dsync.c
```diff
@@ -1250,6 +1696,20 @@ static int dsync_strmap_compare(mfu_flist src_list,
...
     MPI_Barrier(MPI_COMM_WORLD);
+    if (link_dest_path != NULL) {
+        rc = dsync_strmap_compare_link_dest(src_list, src_same_list, src_map, link_dst_list,
+             link_dst_same_list, link_dst_map, strlen_prefix, mfu_copy_opts, src_path, dest_path);
+
+        dsync_generate_real_lists(strlen_prefix, dst_map, src_path, dest_path, link_dest_path,
+             dst_list, src_ignore_list, src_cp_list, src_real_cp_list, dst_remove_list,
+             src_same_list, link_dst_same_list, link_dst_link_list);
+
+        uint64_t size1 = mfu_flist_size(src_cp_list);
```
adammoody: These sizes don't seem to be used.

src/dsync/dsync.c
```diff
@@ -1071,6 +1502,18 @@ static int dsync_strmap_compare(mfu_flist src_list,
      * list to track files to be deleted from destination */
     mfu_flist src_cp_list     = mfu_flist_subset(src_list);
     mfu_flist dst_remove_list = mfu_flist_subset(dst_list);
+    mfu_flist src_ignore_list = NULL;
```
adammoody: Let's add some comments to describe the purpose of each of these new lists.

src/common/mfu_flist_copy.c
```diff
@@ -1129,6 +1129,44 @@ static int mfu_create_file(mfu_flist list, uint64_t idx,
...
+            srcpath, destpath, mfu_copy_opts);
+
+    /* No need to copy it */
+    if (dest_path == NULL) {
+        return 0;
+    }
+
+    rc = mfu_hardlink(src_path, dest_path);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to create hardlink %s --> %s\n",
```
adammoody: MFU_LOG will add a trailing \n, so let's drop that.

src/common/mfu_flist_copy.c
```diff
@@ -1129,6 +1129,44 @@ static int mfu_create_file(mfu_flist list, uint64_t idx,
...
+                dest_path, src_path);
+        mfu_free(&dest_path);
+        return -rc;
```
adammoody: Let's return the raw error rc here, rather than flipping its sign.  May need to update code that calls this correspondingly.

src/common/mfu_flist_copy.c
```diff
@@ -1284,6 +1322,113 @@ static int mfu_create_files(int levels, int minlevel, mfu_flist* lists,
...
+
+    int level;
+    for (level = 0; level < levels; level++) {
+        /* time how long this takes */
+        double start = MPI_Wtime();
+
+        /* get list of items for this level */
+        mfu_flist list = lists[level];
+
+        /* iterate over items and set write bit on directories if needed */
```
adammoody: let's update this comment to match the code

src/common/mfu_flist_copy.c
```diff
@@ -2406,6 +2551,132 @@ int mfu_flist_copy(mfu_flist src_cp_list, int numpaths,
...
+    /* TODO: filter out files that are bigger than 0 bytes if we can't read them */
+
+    /* create directories, from top down */
+    int tmp_rc = mfu_create_directories(levels, minlevel, lists, 1,
+            srcpath, destpath, mfu_copy_opts);
+    if (tmp_rc < 0) {
+        rc = -1;
+    } else {
+        /* create hard links */
+        tmp_rc = mfu_create_hardlinks(levels, minlevel, lists,
```
adammoody: To be consistent with a normal copy, let's go ahead and try to create hardlinks even if we failed to create directories.  This is so that we may have partial success (we will still create hard links under any directories that were created).  Same for metadata.

Unless you think this is a bad idea?

KnightKu:
Thanks for your comments, Adam, will fix them soon.

src/common/mfu_flist_copy.c
KnightKu: No objections, let's follow current partial success syntax.
Added a FIXME comment for this, if other ones have better idea, will improve it then.

src/common/mfu_flist_copy.c
KnightKu: Done

src/common/mfu_flist_copy.c
KnightKu: Done

src/common/mfu_flist_copy.c
KnightKu: Done

src/dsync/dsync.c
KnightKu: Done, Added comment to each list.

src/dsync/dsync.c
KnightKu: Done

src/dsync/dsync.c
KnightKu: Done

src/dsync/dsync.c
KnightKu: Fixed it.

src/dsync/dsync.c
KnightKu: Done

src/dsync/dsync.c
KnightKu: Done

adammoody:
Thanks @KnightKu .  I'll make one more pass on it after those changes.

adammoody:
Looks good.  I'll merge this and then make a pass with some edits.

Thanks for all of your work on this, @KnightKu !  This will be a nice feature.

PR #267, opened 2019-04-25
Add an error for >1 dtar rank

gonsie:


gonsie:
Shoot, I did this on my phone and forgot the sign off. I’ll force add that next time I’m at a computer.

dsikich:
@gonsie your rules ;)

src/dtar/dtar.c
```diff
@@ -95,6 +95,15 @@ int main(int argc, char** argv)
     MPI_Comm_rank(MPI_COMM_WORLD, &rank);
     MPI_Comm_size(MPI_COMM_WORLD, &size);
 
+    if (size > 1) {
+        if (rank == 0) {
+            printf(“ERROR: dtar only works with one MPI rank.\n”);
```
adammoody: `MFU_LOG(MFU_LOG_ERR, "dtar currently only works with one MPI rank");`

adammoody: Figured we can use our MFU_LOG function here.

gonsie:
thanks @adammoody should be good (if it builds)

PR #268, opened 2019-04-25
malloc error message

gonsie:
When out of memory, try using more nodes!

Signed-off-by: Elsa Gonsiorowski (Hylonome)


adammoody:
Thanks @gonsie .  Can you also update the messages in mfu_memalign and mfu_strdup?  There are also mfu_strdupf and mfu_bcast_strdup, but those each call MFU_MALLOC so they should be covered.

adammoody:
Thanks @gonsie !

PR #269, opened 2019-04-25
Modified code to take both short and long options.

levatin1:
Did not implement -w and -r yet, but added a -n for number of files.


levatin1:
Adam or anyone: 

      Please test dfilemaker with a few options.  Thanks.

JoAnne

adammoody:
Thanks @levatin1 .  For the size parameters that take units, let's pass those to mfu_abtoull, which will parse a string like 1GB and return the number of bytes in an unsigned long long.  You can see an example of it being used here:

https://github.com/hpc/mpifileutils/blob/master/src/dfilemaker1/dfilemaker1.c#L49

Can you update the PR to drop the makefile and example run script.  Those are useful to keep in your directory, but we'll keep them out of the repo.  To keep those around so that you don't lose your own copies of them, first make a copy of each file (cp makefile makefile.bak), then run "git rm" on each file (git rm makefile), push your commit.  Then you can copy them back if you'd like (cp makefile.bak makefile)

src/dfilemaker/dfilemaker.c
```diff
@@ -580,25 +653,106 @@ int main(int narg, char** arg)
...
+              // range for maxflen
+              mmparse(optarg,&minterm,&maxterm);
+
+              /* read file size */
+              if (mfu_abtoull(minterm, &sizeminl) != MFU_SUCCESS)
+              {
+                if (rank == 0)
+                {
+                  printf("Could not interpret %s as file size\n", minterm);
+                  fflush(stdout);
```
adammoody: For error messages, let's replace printf("") flush() with MFU_LOG(MFU_LOG_ERR, "")

src/dfilemaker/dfilemaker.c
```diff
@@ -566,6 +616,29 @@ int main(int narg, char** arg)
     int* lind; // list of ints in order to resort things
     int initsum, noff;
 
+    int kft=0,ifac=0; // kft is index of filltype
+    unsigned long long sizeminl,sizemaxl;  // for mfu_abtoul
+    long int sizemin=0,sizemax=0;
```
adammoody: For file sizes, let's store these as uint64_t to be consistent with the other tools.

src/dfilemaker/dfilemaker.c
```diff
@@ -580,25 +653,106 @@ int main(int narg, char** arg)
...
+        if (c <= 0) break;
+        minterm=(char*)MFU_MALLOC(10*sizeof(char));
+        maxterm=(char*)MFU_MALLOC(10*sizeof(char));
+        switch (c) {
+           case 'i':
+             jseed = atoi(optarg);
+             if (jseed) iseed=jseed;
+             break;
+           case 'f':
+             for (kft=0;kft<4;kft++) if (strcmp(optarg,filltype[kft])==0) break;
```
adammoody: To match style we're using in the code base, let's add braces to all for/if/while statements and place the code block indented on the following lines.

src/dfilemaker/dfilemaker.c
```diff
@@ -580,25 +653,106 @@ int main(int narg, char** arg)
...
+              if (rank == 0) print_usage();
+              break;
+            default: 
+              break;
+        }
+     };
+     srand(iseed);
+     if (nmax > 0) ntotal=nmin+rand()%(1+nmax-nmin);
+     if (depmax > 0) nlevels=depmin+rand()%(1+depmax-depmin);
+     if (sizemax > 0) maxflen=sizemin+rand()%(1+sizemax-sizemin);
```
adammoody: Let's use braces and indentation here, too.

adammoody:
Thanks, @levatin1 .  Coming along nicely.  I added a few comments just about fixing up the style to match the rest of the code.

adammoody:
Also, @levatin1 .  For new commits that you push, we've added a check to make sure people have signed off.  To do this, just add "-s" to your "git commit" step.

ofaaland:
Replaced by #604

src/dfilemaker/dfilemaker.c
ofaaland: Addressed in #604

src/dfilemaker/dfilemaker.c
ofaaland: Addressed in #604

src/dfilemaker/dfilemaker.c
ofaaland: Addressed in #604

PR #271, opened 2019-05-03
Build fixes

ana:
These two commits fix the build problems I found while building mpifileutils in a recent openSUSE system.


adammoody:
Thanks @ana and @gonsie .

PR #275, opened 2019-05-14
Progress

adammoody:


adammoody:
MPI doesn't seem to progress ibcast/ireduce in the background very well.  I was worried that might be the case.  The net effect it that progress messages don't get printed very frequently in some situations, and it seems to get worse with increasing scale.

We could try to roll our own bcast/reduce out of isend/irecv as a second option using a tree with a much wider degree.

PR #282, opened 2019-05-31
dcp: add --chunksize and --blocksize options

adilger:
Add command-line options for --chunksize and --blocksize, and add
them to the dcp.1 man page, since these options now exist for both
dcp and dcp1.

Improve the usage message for these options in dcp.1 to not be
self-referential, and explain what they are actually used for.
It was otherwise unclear what the difference between them was.

Keep the defaults at 1MB for this patch, but it probably makes
sense to make them at least the Lustre stripe size to minimize
contention between multiple threads reading/writing the same file.


adammoody:
Great, thanks @adilger !

src/dcp/dcp.c
```diff
@@ -135,6 +140,16 @@ int main(int argc, char** argv)
...
         switch(c) {
+            case 'b':
+                if (mfu_abtoull(optarg, &megs) != MFU_SUCCESS || megs == 0) {
+                    if (rank == 0) {
+                        MFU_LOG(MFU_LOG_ERR,
+				"Failed to parse block size: '%s'\n", optarg);
+                        usage = 1;
+                    }
+                }
+                mfu_copy_opts->block_size = (size_t)megs * 1024 * 1024;
```
adammoody: @adilger , the mfu_abtoull function parses strings like "1MB" already, so this line would actually change such a string to produce a value of 1GB.  Let's just have the user specify the size in bytes rather than MB.  They'll have to remember to give an M or MB suffix on the value, but I think our hint of a 1MB default should be a good reminder for them.

src/dcp/dcp.c
```diff
@@ -65,7 +65,9 @@ void print_usage(void)
 #ifdef LUSTRE_SUPPORT
     /* printf("  -g, --grouplock <id> - use Lustre grouplock when reading/writing file\n"); */
 #endif
+    printf("  -b, --blocksize     - IO buffer size in MB (default 1MB)\n");
     printf("  -i, --input <file>  - read source list from file\n");
+    printf("  -k, --chunksize     - work size per task in MB (default 1MB)\n");
```
adammoody: According to the comment about mfu_abtoull, let's update this and the blocksize descriptions to be "work size per task in bytes (default 1MB)"

doc/rst/dcp.1.rst
```diff
@@ -11,26 +11,37 @@ DESCRIPTION
...
 Parallel MPI application to recursively copy files and directories.
 
-dcp is a file copy tool in the spirit of :manpage:`cp(1)` that evenly distributes
-work across a large cluster without any centralized state. It is
+dcp is a file copy tool in the spirit of :manpage:`cp(1)` that evenly
+distributes the work of scanning the directory tree, and copying file
+data across a large cluster without any centralized state.  It is
 designed for copying files that are located on a distributed parallel
-file system.
+file system, and will split large file copies across multiple threads.
```
adammoody: "across multiple processes."  So far, we don't have any threads.  Threads likely would improve performance for us, at least for small process counts, but it'll take some effort to add.

src/dcp/dcp.c
adilger: Since the default transfer size is 1 MB, and if the user provides an input size that is too small would just result in an error (which IMHO isn't very useful), I think it would make sense to accept values that are small (say < 1024) as units of MB, and if any units are given the values will always be large enough to avoid this case.

src/dcp/dcp.c
adilger: This wouldn't be needed with the above change.

PR #284, opened 2019-06-05
dcp: add --chunksize and --blocksize options

adilger:
Add command-line options for --chunksize and --blocksize, and add
them to the dcp.1 man page, since these options now exist for both
dcp and dcp1.

Improve the usage message for these options in dcp.1 to not be
self-referential, and explain what they are actually used for.
It was otherwise unclear what the difference between them was.

Keep the default values at 1MB for now, but it probably makes
sense to make them at least the Lustre stripe size to minimize
contention between multiple threads reading/writing the same file.


adilger:
@adammoody this is a refresh of PR #282 but it seems to have created a new PR.  The main change is that `--chunksize 64` or other numbers below 1024 are considered to be in units of MiB and explicitly converted to bytes, while anything with a unit specifier will be 1024 or more (e.g. `1KB` or `64M`) will use the supplied units and be converted to bytes by `mfu_abtoull()`.  I considered making the cutoff to be 4096 or even 1048576, since it really doesn't make sense to split a file into 1KB chunks or even 4KB chunks, so I'm open to suggestions here.

IMHO, this is more user friendly than returning an error for `--chunksize=8` or `--chunksize=16` that basically amounts to "you need to add 'M' at the end", which we could do internally and save everyone grief.  I suspect this is the most intuitive way that most people would be using it.

I didn't update `dcp1.c` since it still isn't clear to me if that tool is deprecated, or if it is still being used?  If it is still being used then I will update it to behave similarly.

adilger:
During internal discussion of this patch, a question arose whether it should enforce `chunksize >= N * blocksize` or if it is acceptable if these values are not even multiples and/or `blocksize > chunksize`?

adammoody:
The chunksize and blocksize can both be arbitrary, or at least that was the original intent.

The chunksize is used to determine boundary points when logically slicing a file into portions to be distributed to different ranks.  The starting offset of each work item will be aligned to a chunksize boundary within a file.  The length of a work item can be arbitrary.  If given the middle portion of some file, the length of the work item will be an integer multiple of the chunksize.  The length of the work item corresponding to the end of a file can be arbitrary.

The blocksize defines the size of the buffer used to read and write to the file system.  It defines the maximum I/O transfer size, and a process will read/write in blocks of that size assuming its work item has sufficient data.  A process can also read/write data in smaller amounts, e.g., if at the end of a file or the end of the region assigned to that process.

Given a work item, whose starting offset will be aligned on some integer multiple of the chunksize and whose length is arbitrary, a process will read/write the file in amounts of at most blocksize bytes at a time until it has processed all data in its work item.  The last read/write may be less than a full blocksize.

adammoody:
To answer your specific question, it is acceptable for ```blocksize > chunksize```.

adammoody:
There are some cases where dcp1 is faster than dcp.  We'd like to eventually merge the dcp1 algorithm into dcp, which we would either select automatically or expose to the user explicitly through an option.  However, that work is non-trivial, so for now we expose these two algorithms to the user via two separate tools (dcp and dcp1).  One or the other may be best for a particular copy operation.

adammoody:
@adilger , thanks for pushing the updated PR.

I appreciate your point about checking on the size of the entry passed in and assisting the user by implicitly tacking on the MB units.  However, even though it requires a bit more work, let's keep with the requirement on having the user add units when needed.

There are a few reasons for this, but the primary one is to provide consistency in the user interface across tools.  There are options in other mpiFileUtils tools that take byte sizes, and some of those may refer to things like page sizes or total memory consumption that are smaller or larger than MB units.  We'd like to keep the user interface consistent so that the user can enter a byte size the same way across all tools.  This way the user doesn't need to remember which units go with which tool.  It does require the user to know what a reasonable value is if they opt to change it, but we can signal that by listing the default in our help pages.

src/dcp/dcp.c
```diff
@@ -124,9 +128,10 @@ int main(int argc, char** argv)
 
     /* Parse options */
     int usage = 0;
+    int megs = 0;
```
adilger: It is worthwhile to note that this is declared as an `int` but `mfu_abtoull() takes an `unsigned long long *` argument, but no warning was generated by the builder for this.  I just noticed it when updating the patch to use "bytes" instead of "megs" for this variable name.  I'd guess you are not using `-Werror` on your builders, which likely allows a lot of warnings to slip through unnoticed...

I've fixed this patch, but something to think about.

src/dcp1/dcp1.c
```diff
@@ -268,11 +268,11 @@ void DCOPY_print_usage(void)
...
     printf("  -p, --preserve      - preserve permissions, ownership, timestamps, extended attributes\n");
-    printf("  -p, --verbose       - verbose output\n");
-    printf("  -p, --quiet         - quiet output\n");
+    printf("  -v, --verbose       - verbose output\n");
+    printf("  -q, --quiet         - quiet output\n");
     printf("  -s, --synchronous   - use synchronous read/write calls (O_DIRECT)\n");
-    printf("  -k, --chunksize     - specify chunksize in MB unit (default 1MB)\n");
-    printf("  -b, --blocksize     - specify blocksize in MB unit (default 1MB)\n");
+    printf("  -k, --chunksize     - work size per task in MB (default 1MB)\n");
+    printf("  -b, --blocksize     - IO buffer size in MB (default 1MB)\n");
```
adilger: The other confusing aspect is that the `dcp/dcp1.c` usage message also says that `--chunksize` and `--blocksize` are specified in MB (as above), when they are actually in units of bytes.  I've corrected the usage message there to match the actual behaviour.

src/dcp/dcp.c
adammoody: Thanks!  Good point.  Yes, we do need to enable that option in our build.  We likely have a lot of warnings that need to be fixed.

src/dcp1/dcp1.c
adammoody: Thanks!

adammoody:
Looks good.  I'll merge this in and then make a few edits to match the code style.

Thanks again, @adilger !

PR #285, opened 2019-06-10
gpfs: acl copy messages and return codes

adammoody:
* eliminate some GPFS ACL debug messages, since verbosity was increased in earlier commit
* change messages to errors
* avoid printing messages if either source or destination are not GPFS
* return error if source and destination are GPFS but ACL copy fails
* move ACL copy logic to mfu_copy_acls() and only copy ACL when -p is thrown
* fix for buffer size bug when original ACL buffer is too small

### DIFF
--- a/src/common/mfu_flist_copy.c
+++ b/src/common/mfu_flist_copy.c
@@ -356,87 +356,106 @@ static int mfu_copy_permissions(
                );
             rc = -1;
         }
+    }
+
+    return rc;
+}
+
+/* copy GPFS ACLs from source to destination */
+static int mfu_copy_acls(
+    mfu_flist flist,
+    uint64_t idx,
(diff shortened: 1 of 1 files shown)


adammoody:
@sjtbham , I made some changes to the GPFS ACL copy code you contributed earlier.  I started looking at this because I was getting lots of debug messages printed after we bumped up the default verbosity of our tools in an earlier commit.  However, I found a few other things while looking to fix that.

If you have time, would you please take a look at these changes and provide feedback?

adammoody:
@sjtbham , I'll go ahead and merge this.  It would still help if you have time to review it in case I missed something.  In particular, one change I made is to adjust the stated size of the buffer in a retry due to a buffer-too-small condition.

PR #287, opened 2019-06-11
dcp: style edits to chunk/block size options

adammoody:
* always use braces
* no need for newline in MFU_LOG messages
* list options in consistent order between man page, print_usage, long opts, getopt string, and getopt while loop
* list verbose, quiet, and help as last options
* dcp: fix: have all ranks exit if chunksize/blocksize argument fails to parse


adammoody:
Thanks for the review, @KnightKu .

Issue #283, opened 2019-06-04
drm: output list of deleted files option


gonsie:
admins would like a list of the files that drm deletes... just in case a user asks about it later.

adammoody:
We could do this with a new --output option in drm.

There are two lists I can think of that might be of interest.
    
    1) files that were deleted
    2) files that were not deleted, because they didn't exist or there was some other I/O error
    
Perhaps one could break list 2 into sublists based on the type of error?

adammoody:
Working this issue in PR https://github.com/hpc/mpifileutils/pull/288

PR #288, opened 2019-06-11
drm: add output option

adammoody:
Adds new --output and --text options to drm to write list of files that drm attempted to delete.  Note, since drm does not stat files by default, its default output files does not include stat info.  Does not currently work with --aggressive option.


gonsie:
In #283 you mention 2 possible lists... but I think in this implementation you’ve chosen a third option. Just to be clear, the output file is the list of files that drm *attempted* to delete, correct? 
Could you clarify the man page text? Otherwise, looks good to me!

adammoody:
Yes, this is the list of items drm attempts to delete, which is a third option.

Good idea.  Updated the man page.

PR #292, opened 2019-06-21
libmfu: do not overwrite block_size during flist_copy

adammoody:
The default block_size in mfu_copy_opts will be set when creating a new object.  Someone may override that default, e.g., through the new --blocksize option in dcp.  We do not want to overwrite it during mfu_flist_copy back to the default.

### DIFF
--- a/src/common/mfu_flist_copy.c
+++ b/src/common/mfu_flist_copy.c
@@ -2264,9 +2264,6 @@ int mfu_flist_copy(mfu_flist src_cp_list, int numpaths,
     /* TODO: consider file system striping params here */
     /* hard code some configurables for now */
 
-    /* Set default block size */
-    mfu_copy_opts->block_size = FD_BLOCK_SIZE;
-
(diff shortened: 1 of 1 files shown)


adammoody:
@adilger , I just found this today.  Not sure if you have been doing performance tests with dcp by changing the --blocksize, but I think this would have prevent any new value from taking effect.

adilger:
Funny that you should ask, because we _have_ been doing performance testing and found no performance improvement (or any change) from changing --blocksize or --chunksize. 

Cheers, Andreas

adammoody:
Yes, it would be user-friendly to add some error checking in there.  I'd rather have it abort with an error in the case someone calls with a bad value instead of silently changing the value on them.  Alternatively, we could print a warning message and then change the value to something valid.

We are blindly dividing by chunk_size in a couple places, so if that comes in as 0, it'll blow up.

I'll go ahead and just remove the overwrite for now.

PR #314, opened 2019-12-19
dfind: fix -type to allow all file types

adilger:
The check for MFU_PRED_TYPE() incorrectly used the file type as a
mask for the file mode, which would lead to collisions between types
e.g. S_IFREG (0100000) and S_IFLNK (0120000) or S_IFSOCK (0140000).
This would result in "-type f" also finding symlinks.

Properly use S_IFMT as the type mask for the file, and simplify the
return type since it is already a boolean when computed.

Update the man page to report -type can check for all file types.

### DIFF
--- a/doc/rst/dfind.1.rst
+++ b/doc/rst/dfind.1.rst
@@ -128,11 +128,19 @@ Numeric arguments can be specified as:
    File is of type C:
 
    +---+---------------+
+   | b | block device  |
+   +---+---------------+
+   | c | char device   |
+   +---+---------------+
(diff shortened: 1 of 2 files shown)


adammoody:
Thank you, @adilger !

FYI, we had left out documenting those other types in the man page, because much of the mfuFileUtils suite only really supports directories, regular files, and symlinks.  It's a good time to start to expand support for other file types where we can and then document on a tool-by-tool basis where we still have limitations.  We can aim to do that in the next release.

Issue #322, opened 2020-02-14
error: ‘struct lov_user_mds_data_v2’ has no member named ‘lmd_st’ with Lustre 2.13


kcgthb:
Hi!

I got the following error when trying to compile mpifileutils 0.10 with Lustre 2.13.0:

```
$ cmake [...] -DENABLE_LUSTRE=ON ..
$ make
Scanning dependencies of target mfu_o
[  1%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_bz2.c.o
[  2%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_bz2_static.c.o
[  4%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_compress_bz2_libcircle.c.o
[  5%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_decompress_bz2_libcircle.c.o
[  7%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_flist.c.o
[  8%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_flist_chunk.c.o
[ 10%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_flist_copy.c.o
[ 11%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_flist_io.c.o
[ 13%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_flist_chmod.c.o
[ 14%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_flist_create.c.o
[ 16%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_flist_remove.c.o
[ 17%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_flist_sort.c.o
[ 19%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_flist_usrgrp.c.o
[ 20%] Building C object src/common/CMakeFiles/mfu_o.dir/mfu_flist_walk.c.o
/dev/shm/mpifileutils-0.10/src/common/mfu_flist_walk.c: In function ‘lustre_mds_stat’:
/dev/shm/mpifileutils-0.10/src/common/mfu_flist_walk.c:157:69: error: ‘struct lov_user_mds_data_v2’ has no member named ‘lmd_st’
         lstat_t* ls = (lstat_t*) & ((struct lov_user_mds_data*) buf)->lmd_st;
                                                                     ^
[...]

PR #325, opened 2020-02-20
mfu: drop lustre-specific methods for fetching stat info

adammoody:
This drops the Lustre-specific functions to query stat information for a file.  The particular field used for this does not exist or has been renamed in Lustre v2.13.  Since we are not currently using these functions, the easiest fix is to remove them.  This is to fix https://github.com/hpc/mpifileutils/issues/322


adammoody:
@kcgthb , I've got this PR to just delete those Lustre MDS stat functions.  If there is still value in these in that it may provide a faster way to stat a file, we could add them back later.

kcgthb:
Thanks, that seems like a good plan!

PR #326, opened 2020-02-23
dfind: fix segfault when missing terminating semicolon in --exec

adammoody:


adammoody:
Replaced with https://github.com/hpc/mpifileutils/pull/327

PR #327, opened 2020-02-25
Dfindspace

adammoody:


adammoody:
This is an improved PR for https://github.com/hpc/mpifileutils/pull/326.  It uses fork/exec instead of system so that it's not necessary to escape characters in the string constructed for the system call.

This still has a limitation in that it does not support any --exec arguments that contain embedded '\0' characters.

PR #328, opened 2020-03-19
correct usage message in dfilemaker

ofaaland:
Give the user correct guidance about the relationship between
ntotal and nlevels.

The existing output is confusing:
```
bash-4.2$ dfilemaker --help                   
Usage: dfilemaker <nitems> <nlevels> <maxflen> [-i seed]
bash-4.2$ dfilemaker 5 4 3 
nfiles must be greater than nlevels
```

1. It refers to "nfiles" which is not the name of any of the arguments as described by dfilemaker
2. it doesn't indicate how nfiles is calculated

Signed-off-by: Olaf Faaland <faaland1@llnl.gov>

### DIFF
--- a/src/dfilemaker/dfilemaker.c
+++ b/src/dfilemaker/dfilemaker.c
@@ -585,7 +585,8 @@ int main(int narg, char** arg)
      *----------------------------------------------*/
     if (narg < 4) {
         if (rank == 0) {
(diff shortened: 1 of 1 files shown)


adammoody:
Thanks, @ofaaland !

PR #334, opened 2020-05-18
dcp: daos support

dsikich:


adammoody:
@dsikich , I checked out a copy to try to build, but I got an error about an unknown type for mpi_file_t.  Did you forget to check in some source files?

I see what you mean about one of those commits mixing in some of my commits.  Let's try to clean that up.  It's hard to see what all changed, but I'm more worried it might somehow revert some changes we've recently added to master.

I'm hoping that's not too much work, and I can maybe pitch in to help.  Pull the latest master and make a fresh daos branch from there.  You should be able to add any brand new files pretty easily.  Of course the hard work will be fixing up the existing files with your changes.

dsikich:
@adammoody it should be in https://github.com/hpc/mpifileutils/pull/334/files#diff-a467dc513fb2dbb468e8593ee3fb2466 in mfu_param_path.h. Do you not have the mfu_file_t struct in the changes you downloaded?

dsikich:
@adammoody for some reason there are changes in the first commit like in .travis.yml, etc and many more I did not modify at all, and git shows 0 changes too, but I think what happened is that the file mode was changed and when I downloaded it from git and added it back it shows up as a change? I think that is maybe what happened.

dsikich:
Closing and instead using #338

PR #333, opened 2020-05-18
dcp: add daos support

dsikich:
Add daos support to dcp

1. Add mfu_file_t struct that abstracts fd vs objects in daos, the struct also specifies a file type (POSIX vs. DAOS)
2. Each mfu_* I/O function that was required for the copy now has a mfu_file_* I/O function that will call the corresponding DAOS or POSIX functions, this can be seen in src/common/mfu_io.c
2. Add DAOS_SUPPORT option to mpifileutils build

PR #335, opened 2020-05-20
dfind: add DAOS support

dsikich:
dfind: Add DAOS support
    
    Use DAOS backend to support dfind tool:
        - Add command line options for DAOS
        - connect to pool and open cont
        - support DAOS unified namespace path (only from root)
        - unmount DFS and disconnect from DAOS pool
    Co-authored-by: Mohamad Chaarawi <mohamad.chaarawi@intel.com>

NOTE: depends on PR #333 being merged first.


dsikich:
@adammoody @gonsie This PR depends on #335 being merged first.

src/dfind/dfind.c
```diff
@@ -288,10 +341,15 @@ int main (int argc, char** argv)
     mfu_pred_times* now_t = mfu_pred_now();
 
     int ch;
+    int rc;
 
     mfu_pred* pred_head = mfu_pred_new();
     char* inputname  = NULL;
     char* outputname = NULL;
+    //char* mnewer     = NULL;
```
adammoody, line 349: Are these variables being used?

src/dfind/dfind.c
```diff
@@ -302,6 +360,12 @@ int main (int argc, char** argv)
         {"quiet",     0, 0, 'q'},
         {"help",      0, 0, 'h'},
 
+	/* DAOS specific options */
+        { "pool"                , required_argument, 0, 'x' },
```
adammoody, line 364: Let's namespace the daos-specific options to avoid name collisions with other options / storage systems.  Something like "daos-pool" instead of just "pool"?

src/dfind/dfind.c
```diff
@@ -539,6 +624,38 @@ int main (int argc, char** argv)
...
+    /* create new mfu_file objects */
+    mfu_file_t* mfu_file = mfu_file_new();
+
+#ifdef DAOS_SUPPORT
+    rc = daos_init();
+
+    /* If DAOS "path" then path needs to be reset to path within DAOS container */ 
+    const char** argpaths = (const char**)(&argv[optind]);
+
+    struct duns_attr_t dattr  = {0};
```
adammoody, line 636: I see this struct is passed as a value to one function, but it's not currently used outside of that function.  Can that be moved to be a local within the function, or do you mean to have the function fill it in and return it?

src/dfind/dfind.c
```diff
@@ -582,12 +699,11 @@ int main (int argc, char** argv)
     }
 
 
+    pred_commit(pred_head);
```
adammoody, line 702: We have pred_commit in here twice.  This one was added, but there is another call higher up.  Is that intentional?

src/dfind/dfind.c
dsikich: @adammoody fixed in #339

src/dfind/dfind.c
dsikich: @adammoody fixed in #339, I just used a local variable in the function

src/dfind/dfind.c
dsikich: @adammoody fixed in #339, no I removed one, good catch

dsikich:
@adammoody addressed fixes in #339 since I had to fix up one of the DAOS backend commits that the dcp and dfind tools were built on

PR #338, opened 2020-06-01
dcp: support DAOS

dsikich:
- Use mfu_file_t struct to wrap fd vs obj
    - create DAOS specific io functions
    - Create mfu_file_* functions that call DAOS backend
    - Add DAOS I/O functions to common library
    - call daos io backend in dcp
    - Add FindCART.cmake and FindDAOS.cmake
    - Add ifdef to DAOS specific code
    - create mfu_file in all tools for walk

Signed-off-by: Danielle Sikich <danielle.sikich@intel.com>

PR #339, opened 2020-06-01
dfind: daos support

dsikich:
Use DAOS backend to support dfind tool:
    - Add command line options for DAOS
    - connect to pool and open cont
    - support DAOS unified namespace path (only from root)
    - unmount DFS and disconnect from DAOS pool

NOTE: depends on #338 being merged first


daltonbohning:
@dsikich Do you know the status on this? Are we still trying to add DAOS support to dfind?

dsikich:
@daltonbohning at one point I had a version of this working, but things have changed so much I'd need to go over a lot of it again. Mohamad was using it for I/O 500 performance testing and he had a version where he added some code to libcircle as well, so the version I had didn't include all of his optimizations to libcircle and the walk code in mpifileutils. I think if we incorporate his optimizations it would be nice to have merged into the upstream, but dfind support is more of a low priority item right now. Adding dfind support should not be too much effort, but if we wanted those optimizations added in it might take a bit longer.

daltonbohning:
@dsikich I see. I'll just make a note of this for now, and we can prioritize it later. Thanks!

PR #343, opened 2020-06-20
mfu: always read full size blocks with O_DIRECT

adammoody:
When using O_DIRECT, different file systems have different requirements.  In general, one has to use aligned offsets within the file, aligned buffers in memory, and read/write counts that consist of an integral number of blocks.

The read count must always be an integral number of full blocks, even in the case where one knows that the file does not have that much data.  In this case, the read will simply return a short read.

Similarly when writing data, one may have to write past the end of a file by using a write count that exceeds the desired end of the file.  In that case, one must then truncate the file after it has been written.

To complicate matters further, one may be reading from a file system that requires one alignment block size but writing to a different file system that has a different block alignment.  Perhaps that could be solved by computing the minimum common multiple of the two block sizes?  That remains future work.

### DIFF
--- a/src/common/mfu_flist_copy.c
+++ b/src/common/mfu_flist_copy.c
@@ -1584,11 +1584,15 @@ static int mfu_copy_file_normal(
 
     /* write data */
     size_t total_bytes = 0;
-    while(total_bytes <= (size_t)length) {
+    while(total_bytes < (size_t)length) {
         /* determine number of bytes that we
          * can read = max(buf size, remaining chunk) */
         size_t left_to_read = (size_t)length - total_bytes;
         if(left_to_read > buf_size) {
+            left_to_read = buf_size;
+        }
+        if(mfu_copy_opts->synchronous) {
+            /* O_DIRECT requires particular read sizes */
             left_to_read = buf_size;
         }
 
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -358,6 +358,9 @@ ssize_t mfu_read(const char* file, int fd, void* buf, size_t size)
             /* read some data */
             n += rc;
             tries = MFU_IO_TRIES;
+
+            /* return, even if we got a short read */
+            return n;
         }
         else if (rc == 0) {
             /* EOF */


flx42:
I can try it,  but what about the `EINTR` case? Shouldn't we restart the read/write syscall in this case?

adammoody:
If the read hits an EINTR and then returns -1, the function should retry the read.

However, there is another case that I think is what you're asking about, where we end up with a short read because the read is interrupted but the return value from the read call is greater than 0 because it read some bytes but not everything.  In this case, we'd currently return the read as short.  In the case of O_DIRECT, the subsequent write would then not work as desired.

Is that what you are thinking?

flx42:
> However, there is another case that I think is what you're asking about, where we end up with a short read because the read is interrupted but the return value from the read call is greater than 0 because it read some bytes but not everything
> Is that what you are thinking?

Yes, but my message above was a bit misleading, apologies.
If the process is interrupted by a signal, you will indeed get a short read/write, but not `EINTR` (that's only when no I/O was performed, and the return value is indeed -1).

This is tricky to detect/recover for the direct I/O case though, especially if you're unlucky enough that it happens on the last block and you end up with a short read/write smaller than the buffer size.

For instance, would this case be handled properly by the current logic?
- Block size is 4096 bytes.
- Last block is size 2000, but we need to do a write(2) of size 4096. 
- A signal interrupts the write(2) calls and it returns "1000" instead of "2000".

adammoody:
> 
> 
> > However, there is another case that I think is what you're asking about, where we end up with a short read because the read is interrupted but the return value from the read call is greater than 0 because it read some bytes but not everything
> > Is that what you are thinking?
> 
> Yes, but my message above was a bit misleading, apologies.
> If the process is interrupted by a signal, you will indeed get a short read/write, but not `EINTR` (that's only when no I/O was performed, and the return value is indeed -1).
> 
> This is tricky to detect/recover for the direct I/O case though, especially if you're unlucky enough that it happens on the last block and you end up with a short read/write smaller than the buffer size.
> 
> For instance, would this case be handled properly by the current logic?
> 
>     * Block size is 4096 bytes.
> 
>     * Last block is size 2000, but we need to do a write(2) of size 4096.
> 
>     * A signal interrupts the write(2) calls and it returns "1000" instead of "2000".

Yes, you're correct in that both the read and write paths need to be updated to account for short read/write operations with O_DIRECT.

Another problem that shows up in the current code is that the file pointer will have been moved.  We need to be sure to perform O_DIRECT read/write operations using aligned file offsets in addition to the aligned buffer and counts with full block sizes.  That means the pointer has to be lseek'd backwards before we retry the operation or perhaps we should track offsets manually and switch from read/write to use pread/pwrite instead.

Thanks for your careful review on this.

flx42:
Sure, I will test this PR soon anyway, since this situation is quite low probability.

flx42:
Close, but no cigar:
```
[pid 107454] read(16, "\264\245IU#\17\306\322j\256\205\320\216\391\235\222)\344U\215\2667\269\\\216\276\325Qb"..., 1048576) = 494700
[pid 107454] write(17, "\264\245IU#\17\306\322j\256\205\320\216\391\235\222)\344U\215\2667\269\\\216\276\325Qb"..., 1048576) = 1048576
[pid 107454] read(16, 0x55a75d100000, 1048576) = -1 EINVAL (Invalid argument)
[pid 107454] read(16, 0x55a75d100000, 1048576) = -1 EINVAL (Invalid argument)
[pid 107454] read(16, 0x55a75d100000, 1048576) = -1 EINVAL (Invalid argument)
[pid 107454] read(16, 0x55a75d100000, 1048576) = -1 EINVAL (Invalid argument)
[pid 107454] read(16, 0x55a75d100000, 1048576) = -1 EINVAL (Invalid argument)
[pid 107454] write(2, "ABORT: rank X on HOST: ", 23ABORT: rank X on HOST: ) = 23
```

Here we have `total_bytes == length` so it does one more iteration of the loop:
https://github.com/hpc/mpifileutils/blob/57a3578b68d8df0fa53a7662c33c6c21ce72400f/src/common/mfu_flist_copy.c#L1587

And then it tries one last `read` before exiting the loop:
https://github.com/hpc/mpifileutils/blob/57a3578b68d8df0fa53a7662c33c6c21ce72400f/src/common/mfu_flist_copy.c#L1596-L1601
But the `read` returns `EINVAL`, probably because the file pointer is now unaligned as you mentioned above. Or it's just a normal behavior when past EOF with direct I/O, not sure.

adammoody:
Ok thanks.  Yes, we're converging on a solution.  Checking the code, it should be safe to change the while loop condition from ```total_bytes <= (size_t)length``` to ```total_bytes < (size_t)length```.  That would avoid that last read attempt at the end of the file.

adammoody:
@flx42 , just curious whether making that ```<=``` to ```<``` change in the while loop condition resolves the most recent issue with the final read using an unaligned file offset.

We'll to stamp a 0.10.1 bug fix release of mpiFileUtils next week.  I'll roll all of these fixes into that.  This should help address some common cases.

For a more complete fix, we need to rewrite this portion to ensure we're always using aligned offsets, memory buffers, and proper read/write counts, including when recovering from error situations.  That means rolling back file pointers with lseek or tracking them directly and switching to use pread/pwrite.  I think we need to also account for potential differences in alignment requirements between source and destination file systems.  For example, the O_DIRECT block size might be different between the source and destination, meaning we need a way to identify those values for each system and then compute a least common multiple.  That will require some more careful consideration, so we'll push this additional work to the next major release in v0.11.

flx42:
Yes, when I tried changing the loop condition (a few days ago), it seemed to work fine.
But then it's a bit weird to still have the `break` in the loop, no?

adammoody:
Ok, good to know.  I'll go ahead and merge this in since we know it helps.

I'll keep the break statement in here for now and consider it again on the rewrite for v0.11.  I think the break is still useful so that we exit the loop in the case a file has been shortened after we got its size with stat() but before it the copy completes.

adammoody:
@flx42 , thanks for your help in testing this along the way!

flx42:
@adammoody sure, glad to help.

Do you have an estimate for the release of v0.11 by the way?

adammoody:
> 
> 
> @adammoody sure, glad to help.
> 
> Do you have an estimate for the release of v0.11 by the way?

Hoping to target a v0.11 release in 3 months on the short end, but I can imagine that might get stretched to 6 months.

PR #348, opened 2020-06-30
Dwalk filter text out

brockpalen:
adds a large number of filters (predicates) from dfind to dwalk

### DIFF
--- a/src/dwalk/dwalk.c
+++ b/src/dwalk/dwalk.c
@@ -303,6 +303,7 @@ static void print_usage(void)
     printf("Options:\n");
     printf("  -i, --input <file>      - read list from file\n");
     printf("  -o, --output <file>     - write processed list to file in binary format\n");
+    printf("  --text-output <file>    - write processed list to file in ascii format\n");
     printf("  -t, --text              - use with -o; write processed list to file in ascii format\n");
     printf("  -l, --lite              - walk file system without stat\n");
     printf("  -s, --sort <fields>     - sort output by comma-delimited fields\n");
@@ -447,6 +448,7 @@ int main(int argc, char** argv)
     mfu_pred* pred_head = mfu_pred_new();
     char* inputname      = NULL;
     char* outputname     = NULL;
+    char* textoutputname     = NULL;
     char* sortfields     = NULL;
     char* distribution   = NULL;
 
@@ -464,6 +466,7 @@ int main(int argc, char** argv)
     static struct option long_options[] = {
         {"input",          1, 0, 'i'},
         {"output",         1, 0, 'o'},
+        {"text-output",     required_argument, NULL, 'z' },
         {"text",           0, 0, 't'},
         {"lite",           0, 0, 'l'},
         {"sort",           1, 0, 's'},
@@ -517,6 +520,9 @@ int main(int argc, char** argv)
                 break;
             case 'o':
                 outputname = MFU_STRDUP(optarg);
+                break;
+            case 'z':
+                textoutputname = MFU_STRDUP(optarg);
                 break;
             case 'l':
                 /* don't stat each file on the walk */
@@ -828,6 +834,12 @@ int main(int argc, char** argv)
         }
     }
 
+    /* write text version if also requested 
+ *   independent of --output */
+    if (textoutputname != NULL) {
+        mfu_flist_write_text(textoutputname, flist2);
+    }
+
     /* free off the filtered list */
     mfu_flist_free(&flist2);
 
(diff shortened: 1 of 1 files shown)


adammoody:
Thanks for submitting this, @brockpalen .  I should be able to take a look early next week.

src/dwalk/dwalk.c
```diff
@@ -303,7 +303,7 @@ static void print_usage(void)
     printf("Options:\n");
     printf("  -i, --input <file>      - read list from file\n");
     printf("  -o, --output <file>     - write processed list to file in binary format\n");
-    printf("  -t, --text              - use with -o; write processed list to file in ascii format\n");
+    printf("  --text-output <file>    - write processed list to file in ascii format\n");
```
adammoody, line 306: Let's rename this option to ``--output-text`` just to keep it close to ``--output`` if we decide to list these alphabetically.

src/dwalk/dwalk.c
```diff
@@ -417,15 +541,97 @@ int main(int argc, char** argv)
...
+        
             case 'v':
                 mfu_debug_level = MFU_LOG_VERBOSE;
                 break;
             case 'q':
                 mfu_debug_level = 0;
                 break;
-            case 't':
-                text = 1;
-                break;
```
adammoody, line 428: Let's keep the ``--text`` option active for a bit, so we can ease people over in to switching.

src/dwalk/dwalk.c
```diff
@@ -583,42 +799,52 @@ int main(int argc, char** argv)
...
+        print_flist_distribution(file_histogram, &option, &flist2, rank);
     }
 
     /* write data to cache file */
     if (outputname != NULL) {
-        if (!text) {
-            mfu_flist_write_cache(outputname, flist);
-        } else {
-            mfu_flist_write_text(outputname, flist);
-        }
```
adammoody, line 612: Same as above in keeping ``--text`` active for a bit.

adammoody:
Thanks again @brockpalen .  Sorry that it took so long to get to this!

The ``--text-output`` option looks good, though let's flip the name to be ``--output-text`` to keep it close to ``--output`` under alphabetical listings.  And we should keep the ``--text`` option active for a while for people who may have that option encoded into their scripts.  We can deprecate ``-text``, print a warning that points to the new option name, and then drop it in a future release.  To be consistent, we'll want to add that option to each of the tools.

Would you be up for doing that?

I think the filtering looks fine.  We could run into an explosion of options as we talked about before, especially if we decide other tools would also benefit from built-in filtering.  Until we figure out how to deal with that, we could bring this into dwalk if you want as hidden functionality (supported in the source code, but undocumented in the help output and man pages).  That would give you something that works until we can figure out if we want to go with this or come up with a different but equivalent approach.

One idea we talked about... I feel like there is probably a role to create a tool like ``dpipe`` that would parse and process a chained command like:

``dpipe dfind --amin 60 /path/to/walk | dwalk --sort --output-text foo.txt``

We'd probably want to pick something other than a ``|`` character so the user doesn't have to escape it in their shell command all of the time.  To support this kind of structure easily, we'll need each tool to be implemented as a callable function.  The tool function could take an input list (the pipe input) and produce an output list (the pipe output).  This would be a lot of work, but it would let us keep a smaller number of options in each tool interface.

brockpalen:
Adam,

Sorry just getting back to this.  I'm ok with what your propose.  I will likely keep my own branch where I keep the help options.  I have not put much effort into this because this gave me waht I wanted for my https://github.com/brockpalen/archivetar/ tool and is working quite well.  Though it would be nice if there was a python module that could parse mfu file lists, and that's way beyond my copy and pasting your work.

I never did changes to a pull request before so hopefully hitting those resolve buttons worked.  If not let me know, and if there is github documentation for what to do you can point me to it.

src/dwalk/dwalk.c
```diff
@@ -315,12 +315,104 @@ static void print_usage(void)
...
+    return 1;
+}
+
+static void pred_commit (mfu_pred* p)
+{
+    int need_print = 1;
+
+    mfu_pred* cur = p;
+    while (cur) {
+//        if (cur->f == MFU_PRED_PRINT || cur->f == MFU_PRED_EXEC) {
```
carbonneau1, line 404: Why do you leave the dead code. Please, clean it up.

src/dwalk/dwalk.c
```diff
@@ -315,12 +315,104 @@ static void print_usage(void)
...
+//            need_print = 0;
+//            break;
+//        }
+        cur = cur->next;
+    }
+
+    if (need_print) {
+//        mfu_pred_add(p, MFU_PRED_PRINT, NULL);
```
carbonneau1, line 412: Why do you leave the dead code. Please, clean it up.

carbonneau1 (changes requested):
You left some dead code that needs to be cleaned up. Once done, can you grab the latest and make sure to test with the latest.

thanks

-eric

carbonneau1:
Could you integrate your changes with the latest code line? thanks

PR #358, opened 2020-08-20
fix #357

### LINKED ISSUE #357: daos: make daos_set_paths more robust when checking dfs_prefix
Within `daos_set_paths`, the comparison and error checking should be more robust when `dfs_prefix` is set.

https://github.com/hpc/mpifileutils/blob/db3e7177dc1c8220fef9a389bfc26a9e8d67b2c8/src/dcp/dcp.c#L90-L92

Consider the following use case:
```
dcp --daos-prefix /src/dir  /some/other/src/dir  /some/dest
```
`strstr` would return a non-NULL value, since `/src/dir` is a substring of `/some/other/src/dir`, which passes the comparison check. But this should not be valid, because `dfs_prefix` should strictly be a prefix, and not just a substring. This also causes `argpaths[0]` to be set incorrectly, since it is offset based on the length of `dfs_prefix`.

Furthermore, if `strstr` does return a NULL value, then `dfs_prefix` is assumed (without comparison) to be in the destination path, which also might be incorrect.

The solution is to first check if `dfs_prefix` is strictly at the beginning of `src_path`. If not, then it should check if `dfs_prefix` is strictly at the beginning of `dst_path`. Appropriate error logging is needed at each step.

dsikich:
Use strncmp instead of strstr to check if dfs_prefix is a
prefix of the src or dst path. Using strstr was incorrect
because it was checking for a matching substring instead
of a prefix.

### DIFF
--- a/src/dcp/dcp.c
+++ b/src/dcp/dcp.c
@@ -23,6 +23,21 @@
 #include "mfu.h"
 
 #ifdef DAOS_SUPPORT
+static bool daos_check_prefix(char* path, const char* dfs_prefix) 
+{
+    bool is_prefix = false;
+    /* figure out if dfs_prefix is a prefix of the file path */
+    int prefix_len = strlen(dfs_prefix);
+    if (strncmp(path, dfs_prefix, prefix_len) == 0) {
+        /* if we have another character, it must be '/' */
+        if (strlen(path) > prefix_len &&
+            path[prefix_len] == '/') {
+            is_prefix = true;
+        }
+    }
+    return is_prefix;
+}
+
 static void daos_set_paths(
     char** argpaths,
     const char* dfs_prefix,
@@ -86,14 +101,13 @@ static void daos_set_paths(
         }
 
         /* figure out if prefix is on dst or src for 
-         * for copying container subsets */
-        char* src_ret = strstr(src_path, dfs_prefix);
-        if (src_ret != NULL) {
+         * copying container subsets */
+	if (check_prefix(src_path, dfs_prefix)) {
             mfu_src_file->type = DAOS;
             uuid_copy(src_pool_uuid, dattr.da_puuid);
             uuid_copy(src_cont_uuid, dattr.da_cuuid);
             argpaths[0] = src_path + strlen(dfs_prefix);
-        } else {
+        } else if (check_prefix(dst_path, dfs_prefix)) {
             mfu_dst_file->type = DAOS;
             uuid_copy(dst_pool_uuid, dattr.da_puuid);
             uuid_copy(dst_cont_uuid, dattr.da_cuuid);


dsikich:
fix #357

adammoody:
Oh, this looks familiar.  I recently did something similar in UnifyFS.  However, there is another case that we should check, because some might have a src string like
```
dfs_prefix=/dfs
src=/dfs_not
```
In this case, src is naming a different file system that /dfs.

We should check that:
A) the first portion of src matches dfs_prefix
B) any character in src beyond the length of dfs_prefix must be a '/'

Here is what I did in UnifyFS:
https://github.com/LLNL/UnifyFS/blob/03a48535bf5eba108fec142ee8b56b98af9c71f8/client/src/unifyfs.c#L288

dsikich:
> Oh, this looks familiar. I recently did something similar in UnifyFS. However, there is another case that we should check, because some might have a src string like
> 
> ```
> dfs_prefix=/dfs
> src=/dfs_not
> ```
> 
> In this case, src is naming a different file system than /dfs.
> 
> We should check that:
> A) the first portion of src matches dfs_prefix
> B) if src is longer than dfs_prefix, the next char in src must be a '/'
> 
> Here is what I did in UnifyFS:
> https://github.com/LLNL/UnifyFS/blob/03a48535bf5eba108fec142ee8b56b98af9c71f8/client/src/unifyfs.c#L288

@adammoody Thanks! I did not think about that case initially. I added a similar check to what you have in UnifyFS.

src/dcp/dcp.c
```diff
@@ -86,18 +86,26 @@ static void daos_set_paths(
...
-        } else {
-            mfu_dst_file->type = DAOS;
-            uuid_copy(dst_pool_uuid, dattr.da_puuid);
-            uuid_copy(dst_cont_uuid, dattr.da_cuuid);
-            argpaths[1] = dst_path + strlen(dfs_prefix);
+         * copying container subsets */
+	int prefix_len = strlen(dfs_prefix);
+	if (strncmp(src_path, dfs_prefix, prefix_len) == 0) {
+            /* if we have another character, it must be '/' */
+            if (strlen(src_path) > prefix_len &&
```
adammoody, line 93: Due to this nested if structure, I'm not sure this will do what you want for something like:
```
dfs_prefix=/dfs
src=/dfs_not
dst=/dfs
```
Maybe move this nested if to a function and then call that from daos_set_paths?

dsikich: @adammoody hmm yes in that case it doesn't check the dest path. I added a check_prefix function that I think covers all of the cases now hopefully 👍

src/dcp/dcp.c
```diff
@@ -23,6 +23,21 @@
 #include "mfu.h"
 
 #ifdef DAOS_SUPPORT
+static bool check_prefix(char* path, const char* dfs_prefix)
```
adammoody, line 26:
```suggestion
static bool daos_check_prefix(char* path, const char* dfs_prefix) 
```

adammoody:
Yeah, that looks good.  One last request to namespace the function with a leading "daos" and should be good to go.

dsikich:
@adammoody ok, I committed the name suggestion you made. Do you want me to put these together in one commit or leave them as is?

adammoody:
thanks, go ahead and squash to one commit

dsikich:
> thanks, go ahead and squash to one commit

@adammoody ok, thanks. They are all in one commit now.

PR #359, opened 2020-08-20
Add documentation about DAOS Support

dsikich:
Adds documentation about DAOS Support as well as instructions for enabling DAOS when building MpiFileUtils.


dsikich:
@adammoody I updated based on your suggestions, and I also updated the option names since those have changed I realized since I wrote this

adammoody:
Thanks @dsikich !

PR #362, opened 2020-08-25
Add option for daos-dst-svcl

dsikich:
Previously you could only provide one service level to DAOS,
but we should be able to pass a service level for the
source and destination pools explicitly.

### DIFF
--- a/src/dcp/dcp.c
+++ b/src/dcp/dcp.c
@@ -102,12 +102,12 @@ static void daos_set_paths(
 
         /* figure out if prefix is on dst or src for 
          * copying container subsets */
-	if (check_prefix(src_path, dfs_prefix)) {
+	if (daos_check_prefix(src_path, dfs_prefix)) {
             mfu_src_file->type = DAOS;
             uuid_copy(src_pool_uuid, dattr.da_puuid);
             uuid_copy(src_cont_uuid, dattr.da_cuuid);
             argpaths[0] = src_path + strlen(dfs_prefix);
-        } else if (check_prefix(dst_path, dfs_prefix)) {
+        } else if (daos_check_prefix(dst_path, dfs_prefix)) {
             mfu_dst_file->type = DAOS;
             uuid_copy(dst_pool_uuid, dattr.da_puuid);
             uuid_copy(dst_cont_uuid, dattr.da_cuuid);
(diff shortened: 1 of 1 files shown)


src/dcp/dcp.c
```diff
@@ -491,14 +497,14 @@ int main(int argc, char** argv)
...
-        daos_connect(rank, svc, src_pool_uuid, src_cont_uuid, &src_poh, &src_coh); 
+        daos_connect(rank, src_svc, src_pool_uuid, src_cont_uuid, &src_poh, &src_coh); 
     }
 
     if (mfu_dst_file->type == DAOS) {
         if (daos_uuid_valid(dst_pool_uuid) && !same_pool) {
             /* if DAOS is the source and destination type, and containers are in different pools,
              * then connect to the second pool */
-            daos_connect(rank, svc, dst_pool_uuid, dst_cont_uuid, &dst_poh, &dst_coh); 
+            daos_connect(rank, dst_svc, dst_pool_uuid, dst_cont_uuid, &dst_poh, &dst_coh);
```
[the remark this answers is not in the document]
dsikich: @adammoody that is a good question. It isn't totally clear to me the more I think about it, because I'm not sure if you would just connect to the pool with the greater svc or connect to the same pool twice with different service levels. I will ask about the best way to do this.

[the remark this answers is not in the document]
dsikich: > I don't know the details well enough, but I see this is an input arg to connect. Would we also need to call connect twice if dst_svc != src_svc?

@adammoody I updated this to print an error in the case that the same pool uuid is being used and the dst_svc and src_svc are not equal. This would likely never happen and if it does it is really not efficient to connect to the same pool twice when you can reuse the handle, so I am logging an error here.

src/dcp/dcp.c
```diff
@@ -491,17 +497,20 @@ int main(int argc, char** argv)
...
             /* if DAOS is the source and destination type, and containers are in different pools,
              * then connect to the second pool */
-            daos_connect(rank, svc, dst_pool_uuid, dst_cont_uuid, &dst_poh, &dst_coh); 
+            daos_connect(rank, dst_svc, dst_pool_uuid, dst_cont_uuid, &dst_poh, &dst_coh); 
         } else {
-            /* if DAOS is source and destination type, and containers are in the same pool,
-             * then pool is already connected, so we just need to open and/or create the container */
+            /* Containers are using the same pool uuid.
+             * This is unlikely to ever happen but we can print an error just in case */
+            if (strcmp(dst_svc, src_svc) == 0) {
```
adammoody, line 511: Do you want ``!=0`` here?

And if one of these pointers happens to be NULL, is it valid to call strcmp?

adammoody: I thought we should also check the compound if condition ``if (daos_uuid_valid(dst_pool_uuid) && !same_pool)``.  There are two ways to get to the else, either dst_pool_uuid is not valid or same_pool is set.

Are we always doing the right thing in this else block for the case that dst_pool_uuid is not valid?

dsikich: > I thought we should also check the compound if condition `if (daos_uuid_valid(dst_pool_uuid) && !same_pool)`. There are two ways to get to the else, either dst_pool_uuid is not valid or same_pool is set.
> 
> Are we always doing the right thing in this else block for the case that dst_pool_uuid is not valid?

@adammoody I removed the check for a valid dest uuid since if the type is set the pool uuid is already parsed/checked, and it would fail before this if it wasn't valid. And yes, I did mean for the check to be != 0 👍

src/dcp/dcp.c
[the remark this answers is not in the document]
adammoody: Thanks, the spacing looks off in the code you just added, maybe some TABS work their way in?  Other than that, it's good with me.

src/dcp/dcp.c
dsikich: @adammoody ok, which part specifically? I saw that I originally had some tabs in there, and removed them. But I might have missed one?

adammoody: Actually, it looks good now.  Wonder if my browser just messed up the formatting somehow.  Anyway, looks good to me.  Good to you?

dsikich: @adammoody yep

PR #367, opened 2020-09-04
dsync: add Direct IO support to dsync; speed up metadata updates. 

hwleong:
Signed-off-by: Honwai Leong <hleong@ddn.com>

### DIFF
--- a/src/dsync/dsync.c
+++ b/src/dsync/dsync.c
@@ -189,6 +189,11 @@ const char *dsync_default_outputs[] = {
     "EXIST=COMMON@TYPE=COMMON",
     "EXIST=DIFFER",
     "EXIST=COMMON",
+    "EXIST=COMMON@UID=DIFFER",
+    "EXIST=COMMON@GID=DIFFER",
+    "EXIST=COMMON@PERM=DIFFER",
+    "EXIST=COMMON@ATIME=DIFFER",
+    "EXIST=COMMON@MTIME=DIFFER",
     NULL,
 };
 
@@ -1591,9 +1596,6 @@ static int dsync_strmap_compare(
             continue;
         }
 
-        /* add any item that is in both source and destination to meta
-         * refresh list */
-        strmap_setf(metadata_refresh, "%llu=%llu", src_index, dst_index);
 
         /* item exists in both source and destination,
          * so update our state to record that fact */
@@ -1604,6 +1606,25 @@ static int dsync_strmap_compare(
              dst_list, dst_map, dst_index,
              key);
         assert(tmp_rc >= 0);
+
+        /* add any item that is in both source and destination to meta
+         * refresh list, only include those that have different metadata. */
+        dsync_state uid_state, gid_state, perm_state, atime_state, mtime_state;
+        tmp_rc = dsync_strmap_item_state(src_map, key, DCMPF_UID, &uid_state);
+        assert(tmp_rc == 0);
+        tmp_rc = dsync_strmap_item_state(src_map, key, DCMPF_GID, &gid_state);
+        assert(tmp_rc == 0);
+        tmp_rc = dsync_strmap_item_state(src_map, key, DCMPF_PERM, &perm_state);
+        assert(tmp_rc == 0);
+        tmp_rc = dsync_strmap_item_state(src_map, key, DCMPF_ATIME, &atime_state);
+        assert(tmp_rc == 0);
+        tmp_rc = dsync_strmap_item_state(src_map, key, DCMPF_MTIME, &mtime_state);
+        assert(tmp_rc == 0);
+        if ((uid_state == DCMPS_DIFFER) || (gid_state == DCMPS_DIFFER) ||
+            (perm_state == DCMPS_DIFFER) || (atime_state == DCMPS_DIFFER) ||
+            (mtime_state == DCMPS_DIFFER)) {
+            strmap_setf(metadata_refresh, "%llu=%llu", src_index, dst_index);
+        }
 
         /* Skip if no need to compare type.
          * All the following comparison depends on type. */
--- a/src/dsync/dsync.c
+++ b/src/dsync/dsync.c
@@ -57,6 +57,7 @@ static void print_usage(void)
     printf("      --progress <N>    - print progress every N seconds\n");
     printf("  -v, --verbose         - verbose output\n");
     printf("  -q, --quiet           - quiet output\n");
+    printf("  -s, --synchronous     - use synchronous read/write calls(O_DIRECT)\n");
     printf("  -h, --help            - print usage\n");
     printf("\n");
     printf("For more information see https://mpifileutils.readthedocs.io.\n");
@@ -2826,6 +2827,7 @@ int main(int argc, char **argv)
         {"progress",      1, 0, 'P'},
         {"verbose",       0, 0, 'v'},
         {"quiet",         0, 0, 'q'},
+        {"synchronous",   0, 0, 's'},
         {"help",          0, 0, 'h'},
         {0, 0, 0, 0}
     };
@@ -2841,7 +2843,7 @@ int main(int argc, char **argv)
 
     while (1) {
         int c = getopt_long(
-            argc, argv, "b:cDo:Svqh",
+            argc, argv, "b:cDo:Svqsh",
             long_options, &option_index
         );
 
@@ -2890,6 +2892,12 @@ int main(int argc, char **argv)
             /* since process won't be printed in quiet anyway,
              * disable the algorithm to save some overhead */
             mfu_progress_timeout = 0;
+            break;
+        case 's':
+            mfu_copy_opts->synchronous = true;
+            if(rank == 0) {
+                MFU_LOG(MFU_LOG_INFO, "Using synchronous read/write (O_DIRECT)");
+            }
             break;
         case 'h':
         case '?':


adammoody:
Thanks for the PR, @hwleong .  This generally looks good.

I think the existing code should be avoiding setting metadata on files that are already the same metadata values.  The existing code does add all files to the metadata_refresh list.  But we then eventually call the following function on each of those:

https://github.com/hpc/mpifileutils/blob/a9cd09d6a4412b5a5e87419c8423d6d79f32caab/src/common/mfu_flist_copy.c#L3081

That function checks  each field, and it skips the metadata update for a field if the source and destination are already equal.  Was this not working for you?

For --synchronous I realize we should support O_DIRECT to our function that compares file contents.  I'll open an issue to track that as future work.

hwleong:
The existing code works as it should, but it took another round of walk through of metadata_refresh list to do the check, i.e. all files that exist in both source and destination. If there are millions of files, this step spends a lot of time in checking the metadata and update them where necessary. My modification move the metadata check to earlier loop, and list only files that require metadata update into the metadata_refresh list, greatly reduce the amount of files to be updated later by the mfu_flist_file_sync_meta function. I could see a reduction of few minutes in this step down to less than few seconds. 

The O_DIRECT is important for syncing NFS to NFS, I added this because it reduces intensive memory caching on the server, at some point the heavy caching could potentially knock down a node.

adilger:
On Sep 13, 2020, at 1:01 AM, hwleong <notifications@github.com> wrote:
> The existing code works as it should, but it took another round of walk through of metadata_refresh list to do the check, i.e. all files that exist in both source and destination. If there are millions of files, this step spends a lot of time in checking the metadata and update them where necessary. My modification move the metadata check to earlier loop, and list only files that require metadata update into the metadata_refresh list, greatly reduce the amount of files to be updated later by the mfu_flist_file_sync_meta function. I could see a reduction of few minutes in this step down to less than few seconds.

It makes sense to reduce/avoid duplicate accesses to files.  For distributed
filesystems, repeat access to the files is not free, especially if the access
is far apart and the files are unlikely to still be in cache on the client.

> The O_DIRECT is important for syncing NFS to NFS, I added this because it reduces intensive memory caching on the server, at some point the heavy caching could potentially knock down a node.

Using O_DIRECT is definitely useful for avoiding kernel->user->kernel data
copies, as long as the IO size is large enough (10s of MB) to avoid the sync
latency for the read/write operations (which cannot be cached).

Even better would be trying to use the copy_file_range() interface.  For NFS
this can result in the file being copied directly on the server, and support
for this is just being added to Lustre for 2.14.  Another appealing IO option
is to use libaio and/or io_uring (for newer kernels) to allow direct-but-async
read/write requests to get the best of both worlds.

Cheers, Andreas

adammoody:
> 
> 
> The existing code works as it should, but it took another round of walk through of metadata_refresh list to do the check, i.e. all files that exist in both source and destination. If there are millions of files, this step spends a lot of time in checking the metadata and update them where necessary. My modification move the metadata check to earlier loop, and list only files that require metadata update into the metadata_refresh list, greatly reduce the amount of files to be updated later by the mfu_flist_file_sync_meta function. I could see a reduction of few minutes in this step down to less than few seconds.
> 

I see.  That's a nice performance gain.

I think we need a check for atime/mtime, as well.  See the corresponding check at the bottom of the mfu_flist_file_sync_meta function here:
https://github.com/hpc/mpifileutils/blob/a9cd09d6a4412b5a5e87419c8423d6d79f32caab/src/common/mfu_flist_copy.c#L3119

Does including that atime/mtime check impact the cost?

adammoody:
Thanks for the tips on those other functions and planned future work, @adilger .

I suspect someone could add copy_file_range as an option without too much work.

For O_DIRECT, I know that we also need to do some work to align buffers/offsets properly, especially when copying between source/destination file systems that have different block alignment requirements.  I think we need to compute the minimum common multiple.?  Also, can someone use statfs or something to figure out what the block size is in a portable way?

hwleong:
> > The existing code works as it should, but it took another round of walk through of metadata_refresh list to do the check, i.e. all files that exist in both source and destination. If there are millions of files, this step spends a lot of time in checking the metadata and update them where necessary. My modification move the metadata check to earlier loop, and list only files that require metadata update into the metadata_refresh list, greatly reduce the amount of files to be updated later by the mfu_flist_file_sync_meta function. I could see a reduction of few minutes in this step down to less than few seconds.
> 
> I see. That's a nice performance gain.
> 
> I think we need a check for atime/mtime, as well. See the corresponding check at the bottom of the mfu_flist_file_sync_meta function here:
> https://github.com/hpc/mpifileutils/blob/a9cd09d6a4412b5a5e87419c8423d6d79f32caab/src/common/mfu_flist_copy.c#L3119
> 
> Does including that atime/mtime check impact the cost?

I could add in mtime check, it would cost some cpu time, but I don't think it would cost much, since the check would be done in the same loop. I didn't add it in this change because I thought the following dsync_strmap_compare_lite function would check for it, and if mtime is different, that file would be marked for copy over anyway, which will also take care of updating the mtime. Logically speaking, mtime changes when a file is edited, it is rare to have mtime changed outside of editing the file itself. atime likely is less important from "rsync/dsync" perspective, as long as the file content is not changed. I mean we don't resync a file because the atime is different, we resync it if the mtime is different (which means the content of the file may have been updated that requires to be synced over).

adammoody:
Currently, dsync copies all of uid, gid, mode bits, atime, and mtime from source to destination files so that those values match.  As a first step, I'd like to keep this optimization in this PR consistent with those existing semantics.

Right now, I think src/dest files whose atimes are different would sneak through unchanged.  My other thought was to see whether this additional check on atime/mtime might be causing dsync to execute more metadata updates due to atime differences, which could help explain some of the higher cost.

As future work , it's worth considering whether we should really be copying atime over by default.  You make a point that it's likely not necessary.  If we decide to not copy atime by default, we may want an option to enable it by an option.

hwleong:
Ok, I added mtime and atime (including nanoscale) check for listing into metadata_refresh. Ran the test in my environment, looks good. The time spent in updating timestamps will relatively be longer (which is expected) as more files are accessed at the source (despite no change in content).

adammoody:
Thanks @hwleong .

One more thing... I don't think we have to call dsync_strmap_item_update here, because I think that should be handled in this call:
https://github.com/hpc/mpifileutils/blob/a9cd09d6a4412b5a5e87419c8423d6d79f32caab/src/dsync/dsync.c#L1602

Did you find that it was not?

If it is, then I think we could modify ``mfu_flist_file_sync_meta`` to have a check-but-don't-change mode.  Under this mode, it would do all of its normal checks, but it would not actually update the destination file.  Instead, it would just indicate though a return code whether it found that the destination file needs to be changed (1) or not (0).  Then we can call that function and add the file to the list only if a change is actually needed.  I think that would save us from having to duplicate the code that does the checks.

hwleong:
Honestly speaking, I'm not quite sure what that mfu_flist_file_sync_meta function does, or how to use it. Looks like it will only do the check if dsync_option_need_compare of those attribute returns true. By default they are not checked, unless we explicitly specify it in the undocumented "-o" option. If I understand the code correctly, the defaults are: 

"EXIST=COMMON@CONTENT=DIFFER",
    "EXIST=COMMON@CONTENT=COMMON",
    "EXIST=COMMON@TYPE=DIFFER",
    "EXIST=COMMON@TYPE=COMMON",
    "EXIST=DIFFER",
    "EXIST=COMMON"

So, not sure if we should include all of the attributes like UID=COMMON@UID=DIFFER, GID=COMMON@GID=DIFFER etc. into the defaults? 

Also, with or without the dsync_strmap_item_update, it doesn't affect whether a file would be listed into metadata_refresh. So, we can omit it, if we want.

src/dsync/dsync.c
```diff
@@ -1591,8 +1592,50 @@ static int dsync_strmap_compare(
...
+                dsync_strmap_item_update(dst_map, key, DCMPF_ATIME, DCMPS_DIFFER);
+            }
+            if ((src_mtime != dst_mtime) || (src_mtime_nsec != dst_mtime_nsec)) {
+                dsync_strmap_item_update(src_map, key, DCMPF_MTIME, DCMPS_DIFFER);
+                dsync_strmap_item_update(dst_map, key, DCMPF_MTIME, DCMPS_DIFFER);
+            }
+
+            /* mark file as metadata update required */
+            strmap_setf(metadata_refresh, "%llu=%llu", src_index, dst_index);
+        }
```
adammoody, line 1638: What I'm thinking is that rather than duplicating all of the code that checks for differences in metadata, we could modify the ``mfu_flist_file_sync_meta`` function to add a new test/change flag parameter.  If the test flag is set, it would not actually make any changes, but it would return 1 if it finds that it needs to make a change and returns 0 otherwise.  Then we could simplify this code to be something like:
```suggestion
         * refresh list, only include those that have different metadata. */
        if (mfu_flist_file_sync_meta(src_list, src_index, dst_list, dst_index, mfu_dst_file, 1) == 1) {
            /* mark file as metadata update required */
            strmap_setf(metadata_refresh, "%llu=%llu", src_index, dst_index);
        } 
```

hwleong:
I agree that we should not repeat the code, but the current design of dsync uses two loops. First loop is to sync data, second loop to sync metadata. My change included the check for metadata difference in first loop, then short list files that need to be updated. This is the reduced list which will be used in the second loop. The purpose of my change is to reduce the number of files to go through metadata update in the second loop, and thus reducing the time. 

I sort of understand the purpose of using two loops. We can't seem to avoid having separate check for data and metadata, thus the two loops. I wonder if we could just move the mfu_flist_file_sync_meta directly into the first loop, removing the need of listing into metadata_refresh and run another second loop. 

Doing so may cause conflict with the data check, which also use mtime to determine if a file content is different. so, if we somehow do the sync_meta before the data check, that would sync the mtime, and mark a file to have no file content difference. If we do it afterwards, probably better? I need to have a look to see what makes sense.

hwleong:
So, I had a look again. Looks like it is not possible to put sync_meta into the first loop. So, after evaluating the flow of the code again, I think the latest change I made should make it cleaner. I removed the redundant functions as you suggested, and leverage on existing dsync_compare_metadata function to check for the metadata difference. I added the following to dsync_default_outputs so that dsync_compare_metadata will run the checks on those attributes: 

"EXIST=COMMON@UID=DIFFER",
"EXIST=COMMON@GID=DIFFER",
"EXIST=COMMON@PERM=DIFFER",
"EXIST=COMMON@ATIME=DIFFER",
"EXIST=COMMON@MTIME=DIFFER",

Then after dsync_metadata_compare which is already in the code, I check the state of UID, GID, PERM, ATIME, and MTIME to see if they are DCMPS_DIFFER, if they are we include the file into metadata_refresh.

adammoody:
I think I've confused the situation now.  Sorry.  I am hoping we can call ``mfu_flist_file_sync_meta`` in both loops.

We'd need to modify that function.  The first call would execute that function in a check-only mode.  We'd use the result of that check to know whether to add the file to the metadata_refresh list, like I showed above.

The second call would execute that function in a check-and-change mode which would actually update the metadata on the destination files.  This approach contains the code that does the check to one function, which is safer.

Is that not possible?

hwleong:
I think that's exactly what I have done now. The first loop uses the dsync_compare_metadata function, which is a check-only mode. And the second loop executes the dsync_sync_metadata function to check and update metadata of the short-listed files (from the first loop). 

Both functions are existing functions in the code already. I did not change any of these functions. I simply enable the check for uid, gid, perm, mtime and atime. The original does not include these in the dsync_default_outputs variable.

adammoody:
The part that I'm worried about is we have two different bits of code that both check whether meta data values match that could be a tempting source for bugs to creep in.  However, that's a separate problem from what you're adding in this PR.  We could likely merge in your changes and deal with that as future work.  Let me take a fresh look at things.

adammoody:
Thanks @hwleong .  This looks good.

Before we merge, would you please update the title portion of your commits with something like:
```
dsync: add Direct IO support
dsync: speed up timestamp updates
```
I like to prefix the commits with what tool they affect if the changes are limited to one tool.

Then it's ready to go.

hwleong:
> Thanks @hwleong . This looks good.
> 
> Before we merge, would you please update the title portion of your commits with something like:
> 
> ```
> dsync: add Direct IO support
> dsync: speed up timestamp updates
> ```
> 
> I like to prefix the commits with what tool they affect if the changes are limited to one tool.
> 
> Then it's ready to go.

Do you mean the pull request title? It is a one line title, so I changed it to "dsync: add Direct IO support to dsync; speedup metadata updates". My apology to have included to quite different changes in one pull request. They are supposed to be two different one. 

If you mean changing the commit comment, I'm not sure what would be the best way to change that. As there are three commits that refines to the current state of metadata speed up, do you want all of them to have same comments? The first one is quite long actually.

adammoody:
Oh, right.  Yes, there are 4 commits here, not 2.  I was hoping you could rewrite the commit message on each using a ``git rebase -i master`` and then select ``reword`` to add a short one-line title for each.

So for example, modify the commit message to include a short one-line summary like:
```
dsync: speed up timestamp updates

This change speeds up updating timestamps on newly copied files.

Before this change, the dsync_strmap_compare function
walks through all files that have common existence in
both source and destination to check for difference
in metadata (uid, gid, perm, atime, and mtime), then
update the destination to match the source accordingly.

For large directory, metadata refresh spends a lot of
cycle time for a second walk through all the files.
As noted in dsync.c, this is not efficient, especially
if only a very small number of files needs metadata
update.

This change updates dsync to do an earlier check on
the metadata (uid, gid and perm) right after the
first walk, and include only those that
have different metadata to the metadata_refresh array.
Thus, the mfu_flist_file_sync_meta function only
checks and updates as necessary on a subset of files,
instead of all the files again.

I did not do a check on mtime or atime to determine
if the file need to be included into metadata_refresh
array, since files with different mtime/atime are
considered different, where dsync will copy the file
anyway and update along its metadata.

Signed-off-by: Honwai Leong <hleong@ddn.com>
```

That way it will display more cleanly when someone is browsing the commit history.

After that, you can force push back to your branch.

hwleong:
Alright, got it. Since three of the commits are related to improving metadata updates, I combined them into one single commit edited the comments. Hope this looks cleaner and better.

adammoody:
Looks good.  Thanks, @hwleong !

PR #368, opened 2020-09-09
Enhanced dcp error checking

### LINKED ISSUE #363: daos: return copy errors in dcp
Mirrored DAOS Jira ticket: https://jira.hpdd.intel.com/browse/DAOS-5577

`dcp.c` does not return the error code from `mfu_flist_copy`.

After all of the files have been copied (either with or without errors), `rc` is set to `1` if there is an error.
https://github.com/hpc/mpifileutils/blob/b9eca8e89d41097172836cf16a938d3094e1dcbe/src/dcp/dcp.c#L651-L658

However, farther down in the DAOS cleanup, `rc` is overwritten:
https://github.com/hpc/mpifileutils/blob/b9eca8e89d41097172836cf16a938d3094e1dcbe/src/dcp/dcp.c#L678-L691

Ultimately, this means that any errors coming from `mfu_flist_copy` are not captured in `dcp.c`. These errors are still logged in the corresponding places, but `dcp.c` ignores them and potentially `return 0`, which doesn't indicate an issue has occurred.

It would be useful if `dcp.c` returned this error code from `mfu_flist_copy`. Perhaps the DAOS cleanup can use a `tmp_rc` instead of `rc`, to ovoid overwriting the value. Then, the `rc` value can be returned at the end of `dcp.c`.

daltonbohning:
Error checking for dcp has been enhanced to be more thorough and report specific error codes depending on the failure.
Currently, the new error codes are contained to `dcp.c`.
In the future, it is possible that we could also use error codes in the lower level libraries, but **properly** propagating these errors back to `dcp.c` is a more difficult undertaking which I will save for a later time, as the need arises.

I ran the DAOS functional tests (in progress) against these changes to verify basic functionality for:
- DAOS -> DAOS
- DAOS -> POSIX
- POSIX -> DAOS
- POSIX -> POSIX

### New file: common/mfu_errors.h
- **Please review this for convention**
- Contains some error codes for dcp and DAOS
  - Within the range of 0-255  
- Can be expanded in the future as needed.

### Modified: dcp.c
- Use the new error codes in `mfu_errors.h`
- Better error checking for DAOS
  - Check for required/dependent args
  - Early exit if `dfs_mount` fails
- Print a final message after summary if there were any errors during copy

Closes #363

### DIFF
--- a/src/dcp/dcp.c
+++ b/src/dcp/dcp.c
@@ -40,7 +40,7 @@ static bool daos_check_prefix(char* path, const char* dfs_prefix)
     return is_prefix;
 }
 
-static void daos_set_paths(
+static int daos_set_paths(
     char** argpaths,
     const char* dfs_prefix,
     uuid_t src_pool_uuid,
@@ -120,6 +120,13 @@ static void daos_set_paths(
             argpaths[1] = "/";
         }
     }
+
+    if (dfs_prefix != NULL && !prefix_on_src && !prefix_on_dst) {
+        MFU_LOG(MFU_LOG_ERR, "Invalid DAOS prefix");
+        return ERR_DAOS_INVAL_PRFX;
+    }
+
+    return 0;
 }
 #endif 
 
@@ -487,8 +494,8 @@ int main(int argc, char** argv)
     char** argpaths = (&argv[optind]);
 
 #ifdef DAOS_SUPPORT
-    bool have_src_svc   = src_svc;
-    bool have_dst_svc   = dst_svc;
+    bool have_src_svc   = src_svc != NULL;
+    bool have_dst_svc   = dst_svc != NULL;
     bool have_src_pool  = daos_uuid_valid(src_pool_uuid);
     bool have_dst_pool  = daos_uuid_valid(dst_pool_uuid);
     bool have_src_cont  = daos_uuid_valid(src_cont_uuid);
@@ -510,10 +517,6 @@ int main(int argc, char** argv)
     }
     if (have_dst_cont && !have_dst_pool) {
         MFU_LOG(MFU_LOG_ERR, "Destination container requires destinaion pool");
-        args_valid = false;
-    }
-    if (have_dst_pool && !have_dst_svc) {
-        MFU_LOG(MFU_LOG_ERR, "Destination pool requires destination svcl");
         args_valid = false;
     }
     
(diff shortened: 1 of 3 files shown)


daltonbohning:
You may consider "squashing" these commits, since a few of the commits were work-in-progress.

dsikich (approved):

daltonbohning:
@adammoody 
Just a friendly reminder about this PR. (I'm sure you're probably busy)
Really, I would mostly like your input about common/mfu_errors.h in terms of convention. I.e. error ranges, naming conventions.
I chose to go with the 0-255 range because this is the most compatible range for reading error codes from a returned process.

daltonbohning:
Actually, this isn't quite ready to be merged.
I realize now that I need to rethink the way the errors are being returned, due to some limitations.

Instead of dcp.c ***returning*** the rc for different errors, it will be more compatible and flexible to return 0 if everything checks out, and 1 otherwise (how it was originally). But this time, the change will be that the logged messages will include the rc for different errors.
This is more similar to how it is done in DAOS, and I think now I see why this approach was taken.

adammoody:
Thanks @daltonbohning .  Got some time to take a look at this.  Looks good to me so far, but it sounds like you're wondering about something.  What are you looking at?

src/dcp/dcp.c
```diff
@@ -536,6 +664,7 @@ int main(int argc, char** argv)
                     rc = daos_cont_open(src_poh, cuuid, DAOS_COO_RW, &dst_coh, &co_info, NULL);
                     if (rc != 0) {
                         MFU_LOG(MFU_LOG_ERR, "Failed to open DFS2 container");
+                        daos_rc = 1;
```
adammoody, line 667: If we hit an error here, is the bcast below still valid?

daltonbohning: You are right, the bcast might not be valid, and that is one such issue I ran into. My latest updates have refactored the way this works so that the error checking is propagated among the processes.

daltonbohning:
> Thanks @daltonbohning . Got some time to take a look at this. Looks good to me so far, but it sounds like you're wondering about something. What are you looking at?

I ran into as issue with the way I was using the error codes. Initially, I went with a very simplified approach of introducing errors in the range of 0-255, since this is most compatible with getting the return code from the various ways a process can be called (command line, fork(), etc.)
But there were some issues with this approach:
- Many MFU functions return negative values, 0-255 wouldn't strictly be a drop-in replacement.
- Working with multi processes means that sometimes there just isn't easily a ***single*** return code for ***dcp***.
- 0-255 isn't a very large range, so it's difficult to designate ranges to different modules.

My latest changes have refactored this, and I'll explain how in more detail once those changes are ready for PR.

daltonbohning:
My branch deviated a bit from my previous PR, I had many small commits, and git wasn't cooperating, so I'm just creating a new PR entirely.

PR #369, opened 2020-09-14
daos: dcp: enhanced error checking

### LINKED ISSUE #361: daos: exit dcp when the source and destination are the same
1. If the DAOS source and destination container are the same, and the source and destination path are the same, then dcp should exit without trying to perform a copy. For example:
```
dcp --daos-svcl 0 --daos-src-pool $pool1 --daos-src-cont $p1cont1 --daos-dst-pool $pool1 --daos-dst-cont $p1cont1 / /
```
2. Similarly, for UNS paths:
```
dcp --daos-svcl 0  /some/uns/path  /some/uns/path
```
3. If feasible, it might also be a good idea to check if the source and destination are the same when one of them is mounted with dfuse. For example:
```
dcp --daos-svcl 0 --daos-src-pool $pool1 --daos-src-cont $p1cont1  /  /dfuse/mount/for/p1cont1
```
4. More generally, we might consider doing the same when the POSIX source and destination are the same. For example:
```
dcp  /posix/path  /posix/path
```

### LINKED ISSUE #363: daos: return copy errors in dcp

### LINKED ISSUE #370: daos: dcp: improved error checking
These DAOS Jira tickets correspond to dcp.

- dcp require pool UUID when cont UUID is given
  - https://jira.hpdd.intel.com/browse/DAOS-5595
- dcp: add error codes
  - https://jira.hpdd.intel.com/browse/DAOS-5600
- dcp gracefully handle failed daos_init()
  - https://jira.hpdd.intel.com/browse/DAOS-5613
- dcp alert user when using UNS path with no svcl
  - https://jira.hpdd.intel.com/browse/DAOS-5614
- dcp make sure pool/container UUID and path exists
  - https://jira.hpdd.intel.com/browse/DAOS-5615
- dcp auto-create destination container in different pool
  - https://jira.hpdd.intel.com/browse/DAOS-5616

daltonbohning:
My branch deviated a bit from my previous PR, I had many small commits, and git wasn't cooperating, so I'm just creating a new PR entirely.

**common/mfu_errors.h**
- New file
- Defines some generic error codes, codes for dcp, and codes for DAOS
  - Ranges are in the 1000s, 1100s, and 4000s
- Defines MFU_ERRF and MFU_ERRP for printing the error codes.
  - Formatted as: MFU_ERR(#)
  - Could be improved in the future to print the"string" error code instead of just MFU_ERR
    - DAOS has something similar to this

**common/mfu_io**
- Defined daos_access

**common/mfu_param_path.c**
- mfu_param_path_check_copy - Return on DAOS error

**common/mfu_util**
- daos_connect
  - Added flag for whether or not to connect to pool
  - Added flag for whether or not to auto-create the destination container
    - No longer creates the source container automatically
  - These flags allow a centralized way to capture errors. I.e. only one bcast is needed for pool connection, cont connection, cont creation
  - Rank 0 now tracks and broadcasts any errors before trying to share the handles
  - Function returns an error

**dcp/dcp.c**
- Use the new error codes from mfu_error.h
- DAOS errors are generally handled more precisely
  - Each process tracks local errors, and stops making DAOS calls once it has run into an error
  - Allreduce on local DAOS errors, so all processes will know to exit
  - Check DAOS arguments before trying to make DAOS calls
  - Refactored daos_connect calls to better capture errors and exit if found
  - More gracefully handle a failed daos_init call
  - Exit if DAOS source is DAOS destination

closes #363 
closes #361
closes #370

### DIFF
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -199,7 +199,7 @@ int mfu_file_access(const char* path, int amode, mfu_file_t* mfu_file)
         int rc = mfu_access(path, amode);
         return rc;
     } else if (mfu_file->type == DAOS) {
-        int rc = daos_access(path, amode);
+        int rc = daos_access(path, amode, mfu_file);
         return rc;
     } else {
         MFU_ABORT(-1, "File type not known: %s type=%d",
@@ -227,11 +227,23 @@ retry:
     return rc;
 }
 
-int daos_access(const char* path, int amode)
+int daos_access(const char* path, int amode, mfu_file_t* mfu_file)
 {
 #ifdef DAOS_SUPPORT
-    /* noop becuase daos have an access call */
-    return 0;
+    char* name     = NULL;
+    char* dir_name = NULL;
+    parse_filename(path, &name, &dir_name);
+    assert(dir_name);
+
+    dfs_obj_t* parent = NULL;
+    int rc = dfs_access(mfu_file->dfs, parent, name, amode);
+    if (rc) {
+        MFU_LOG(MFU_LOG_ERR, "dfs_access %s failed (%d %s)",
+                name, rc, strerror(rc));
+        errno = rc;
+        rc = -1;
+    }
+    return rc;
 #endif
 }
 
@@ -264,7 +276,7 @@ int daos_chmod(const char *path, mode_t mode, mfu_file_t* mfu_file)
     parse_filename(path, &name, &dir_name);
     assert(dir_name);
 
-    dfs_obj_t *parent = NULL;
+    dfs_obj_t* parent = NULL;
     int rc = dfs_lookup(mfu_file->dfs, dir_name, O_RDWR, &parent, NULL, NULL);
     if (parent != NULL) {
         rc = dfs_chmod(mfu_file->dfs, parent, name, mode);
@@ -346,7 +358,7 @@ int daos_stat(const char* path, struct stat* buf, mfu_file_t* mfu_file) {
     parse_filename(path, &name, &dir_name);
     assert(dir_name);
 
-    dfs_obj_t *parent = NULL;
+    dfs_obj_t* parent = NULL;
     int rc;
     if (mfu_file->only_daos) {
         rc = dfs_lookup(mfu_file->dfs, dir_name, O_RDWR, &parent, NULL, NULL);
@@ -564,7 +576,7 @@ void daos_open(const char* file, int flags, mode_t mode, mfu_file_t* mfu_file)
     parse_filename(file, &name, &dir_name);
     assert(dir_name);
 
-    dfs_obj_t *parent = NULL;
+    dfs_obj_t* parent = NULL;
     int rc = dfs_lookup(mfu_file->dfs, dir_name, O_RDWR, &parent, NULL, NULL);
     if (parent != NULL) {
         rc = dfs_open(mfu_file->dfs, parent, name,
@@ -1054,7 +1066,7 @@ int daos_mkdir(const char* dir, mode_t mode, mfu_file_t* mfu_file) {
     assert(dir_name);
 
     /* Need to lookup parent directory in DFS */
-    dfs_obj_t *parent = NULL;
+    dfs_obj_t* parent = NULL;
     int rc = dfs_lookup(mfu_file->dfs, dir_name, O_RDWR, &parent, NULL, NULL);
 
     /* only call mkdir if the dir_name is not the root DFS directory */
@@ -1131,7 +1143,7 @@ retry:
 
 #ifdef DAOS_SUPPORT
 struct dfs_mfu_t {
-    dfs_obj_t *dir;
+    dfs_obj_t* dir;
     struct dirent ents[NUM_DIRENTS];
     daos_anchor_t anchor;
     int num_ents;
--- a/src/common/mfu_util.c
+++ b/src/common/mfu_util.c
@@ -1123,11 +1123,11 @@ void mfu_stripe_set(const char *path, uint64_t stripe_size, int stripe_count)
 }
 
 /* executes a logical AND operation on flag on all procs on comm,
- * returns 1 if all true and 0 otherwise */
-int mfu_alltrue(int flag, MPI_Comm comm)
+ * returns true if all true and false otherwise */
+bool mfu_alltrue(bool flag, MPI_Comm comm)
 {
     /* check that all processes wrote successfully */
-    int alltrue;
-    MPI_Allreduce(&flag, &alltrue, 1, MPI_INT, MPI_LAND, comm);
+    bool alltrue;
+    MPI_Allreduce(&flag, &alltrue, 1, MPI_C_BOOL, MPI_LAND, comm);
     return alltrue;
 }
--- a/src/common/mfu_util.c
+++ b/src/common/mfu_util.c
@@ -107,58 +107,89 @@ void daos_bcast_handle(
     mfu_free(&global.iov_buf);
 }
 
-void daos_connect(
+int daos_connect(
   int rank,
   const char* svc,
   uuid_t pool_uuid,
   uuid_t cont_uuid,
   daos_handle_t* poh,
-  daos_handle_t* coh)
+  daos_handle_t* coh,
+  bool connect_pool,
+  bool create_cont)
 {
-    /* TODO: if src daos path and dst daos path are false 
-     * skip connecting to daos pool */
+    /* assume failure until otherwise */
+    int valid = 0;
+    int rc;
 
     /* have rank 0 connect to the pool and container,
      * we'll then broadcast the handle ids from rank 0 to everyone else */
     if (rank == 0) {
-        d_rank_list_t* svcl = daos_rank_list_parse(svc, ":");
-        if (svcl == NULL) {
-            MFU_ABORT(-1, "Failed to parse DAOS rank list: '%s'", svc);
+        /* Parse svc and connect to DAOS pool */
+        if (connect_pool) {
+            d_rank_list_t* svcl = daos_rank_list_parse(svc, ":");
+            if (svcl == NULL) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to parse DAOS rank list: '%s'", svc);
+                goto bcast;
+            }
+
+            daos_pool_info_t pool_info;
+            rc = daos_pool_connect(pool_uuid, NULL, svcl, DAOS_PC_RW,
+                    poh, &pool_info, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to connect to pool");
+                d_rank_list_free(svcl);
+                goto bcast;
+            }
+            d_rank_list_free(svcl);
         }
 
-        /* Connect to DAOS pool */
-        daos_pool_info_t pool_info;
-        int rc = daos_pool_connect(pool_uuid, NULL, svcl, DAOS_PC_RW,
-                               poh, &pool_info, NULL);
-        if (rc != 0) {
-            MFU_LOG(MFU_LOG_ERR, "Failed to connect to pool");
-        }
-        d_rank_list_free(svcl);
-
-        /* attempt to open the daos container */
+        /* Try to open the container 
+         * If NOEXIST we create it */
         daos_cont_info_t co_info;
         rc = daos_cont_open(*poh, cont_uuid, DAOS_COO_RW, coh, &co_info, NULL);
-
-        /* If NOEXIST we create it */
         if (rc != 0) {
-            /* create the container */
-            uuid_t cuuid;
-            rc = dfs_cont_create(*poh, cuuid, NULL, NULL, NULL);
+            if (!create_cont) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to open DFS container");
+                goto bcast;
+            }
+            
+            rc = dfs_cont_create(*poh, cont_uuid, NULL, NULL, NULL);
             if (rc != 0) {
                 MFU_LOG(MFU_LOG_ERR, "Failed to create DFS container");
+                goto bcast;
             }
 
             /* try to open it again */
-            rc = daos_cont_open(*poh, cuuid, DAOS_COO_RW, coh, &co_info, NULL);
+            rc = daos_cont_open(*poh, cont_uuid, DAOS_COO_RW, coh, &co_info, NULL);
             if (rc != 0) {
                 MFU_LOG(MFU_LOG_ERR, "Failed to open DFS container");
+                goto bcast;
             }
         }
+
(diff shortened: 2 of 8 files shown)


dsikich (approved):

src/common/mfu_io.c
```diff
@@ -227,11 +227,23 @@ int mfu_access(const char* path, int amode)
...
 {
 #ifdef DAOS_SUPPORT
-    /* noop becuase daos have an access call */
-    return 0;
+    char* name     = NULL;
+    char* dir_name = NULL;
+    parse_filename(path, &name, &dir_name);
+    assert(dir_name);
+
+    dfs_obj_t *parent = NULL;
```
adammoody, line 238:
```suggestion
    dfs_obj_t* parent = NULL;
```

src/common/mfu_util.c
```diff
@@ -107,58 +107,89 @@ void daos_bcast_handle(
...
-        int rc = daos_pool_connect(pool_uuid, NULL, svcl, DAOS_PC_RW,
-                               poh, &pool_info, NULL);
-        if (rc != 0) {
-            MFU_LOG(MFU_LOG_ERR, "Failed to connect to pool");
+            daos_pool_info_t pool_info;
+            rc = daos_pool_connect(pool_uuid, NULL, svcl, DAOS_PC_RW,
+                    poh, &pool_info, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to connect to pool");
+                goto bcast;
```
adammoody, line 143: do we want a d_rank_list_free(svcl) call in here?

src/common/mfu_util.c
```diff
@@ -107,58 +107,89 @@ void daos_bcast_handle(
...
         }
-        d_rank_list_free(svcl);
 
-        /* attempt to open the daos container */
+        /* Try to open the container */
         daos_cont_info_t co_info;
         rc = daos_cont_open(*poh, cont_uuid, DAOS_COO_RW, coh, &co_info, NULL);
-
+        
         /* If NOEXIST we create it */
```
adammoody, line 152: Let's pull the lines that set ``rc =`` and the corresponding ``if (rc != 0)`` statement that checks its value right beside each other without any empty lines in between.  I find that helps with long term code maintenance in avoiding someone from inserting code in between that might reuse rc before it is checked.

src/common/mfu_util.c
```diff
@@ -107,58 +107,89 @@ void daos_bcast_handle(
...
     }
 
-    /* broadcast pool and container handles from rank 0 */
+bcast:
+    /* broadcast valid from rank 0 */
+    MPI_Bcast(&valid, 1, MPI_INT, 0, MPI_COMM_WORLD);
+
+    /* return if invalid */
+    if (valid == 0) {
+        return -1;
```
adammoody, line 183: The valid check to guard the handle bast looks good.  Thanks.  One other thing to check is whether we should clean up any resources that may have been allocated if we fail.  I don't know the daos api well enough to check, but I did see one call that looked to trying to free something above which would not be called in certain error cases.

src/dcp/dcp.c
```diff
@@ -22,7 +22,120 @@
...
+
+    /* Containers are using the same pool uuid.
+     * Make sure they are also using the same svc.
+     * This is unlikely to ever happen but we can print an error just in case. */
+    if (same_pool && have_src_svc && have_dst_svc) {
+        if (strcmp(dst_svc, src_svc) != 0) {
+            if (rank == 0) {
+                MFU_LOG(MFU_LOG_ERR, "Using same pool uuid with different svcl's");
+            }
+        rc = 1;
```
adammoody, line 122:
```suggestion
            rc = 1;
```

src/dcp/dcp.c
```diff
@@ -22,7 +22,120 @@
...
+        }
+    }
+
+    /* Make sure the source and destination are different */
+    if (same_cont && have_src_path && have_dst_path) {
+        if (strcmp(src_path, dst_path) == 0) {
+            if (rank == 0) {
+                 MFU_LOG(MFU_LOG_ERR, "DAOS source is DAOS destination");
+            }
+        rc = 1;
```
adammoody, line 132:
```suggestion
            rc = 1;
```

src/common/mfu_util.c
daltonbohning: Ah, yes, I agree!

src/dcp/dcp.c
```diff
@@ -470,104 +595,160 @@ int main(int argc, char** argv)
...
             same_pool = true;
         }
-    } 
+    }
 
     /* connect to DAOS source pool if uuid is valid */
-    if (mfu_src_file->type == DAOS) {
-        /* if DAOS source pool uuid is valid, then set source file type to DAOS */
-        daos_connect(rank, src_svc, src_pool_uuid, src_cont_uuid, &src_poh, &src_coh); 
+    if (!local_daos_error && mfu_src_file->type == DAOS) {
```
adammoody, line 687: Skimming the code, I get the impression that local_daos_error is tracking whether the calling process has seen an error, which may be independent of other processes.  Is that right?

daltonbohning: Correct. It is *technically possible* that some DAOS call might fail for only a single process, some of which might stem from erroneous errors. Practically speaking, most processes at the point of the broadcast into global_daos_error are *most likely* in a similar state, but to be safe in calling MPI_Finalize(), etc., it's best to make sure. 
One such out-of sync issue that could possibly arise, though unlikely, would be in our call to duns_resolve_path. 
Let's say we have 2 processes. One of them calls duns_resolve_path and succeeds. Before the other process makes the call, the UNS path is removed from the DAOS system, and then the 2nd process makes the call, and fails. Now the processes are in different states.
This is where my reasoning for "local_daos_error" comes from, among other erroneous network errors that might crop up.

adammoody: Assuming that's the case, one thing to be careful of is mixed error conditions, like if local_daos_error=true on rank=0, but false on rank=1.  It looks like daos_connect includes a collective (bcast), but then you might have rank 0 skipping that call while rank 1 does.

src/common/mfu_util.c
daltonbohning: Yes, you are right. I missed that. Thank you!

src/dcp/dcp.c
adammoody: > 
> 
> Correct. It is _technically possible_ that some DAOS call might fail for only a single process, some of which might stem from erroneous errors. Practically speaking, most processes at the point of the broadcast into global_daos_error are _most likely_ in a similar state, but to be safe in calling MPI_Finalize(), etc., it's best to make sure.
> One such out-of sync issue that could possibly arise, though unlikely, would be in our call to duns_resolve_path.
> Let's say we have 2 processes. One of them calls duns_resolve_path and succeeds. Before the other process makes the call, the UNS path is removed from the DAOS system, and then the 2nd process makes the call, and fails. Now the processes are in different states.
> Another such issue might be with mounting, since we could possibly run out of mountpoints for only one process.
> This is where my reasoning for "local_daos_error" comes from, among other erroneous network errors that might crop up.

Makes sense.  Good thinking.

daltonbohning: Oh yes... I'll need to rework that portion.

daltonbohning:
Thanks for all the input @adammoody !
I should have these things all resolved tomorrow.

src/dcp/dcp.c
adammoody: This kind of check comes up a lot.  This utility function comes in handy:
https://github.com/hpc/mpifileutils/blob/a9cd09d6a4412b5a5e87419c8423d6d79f32caab/src/common/mfu_util.c#L1096

You can use it like `` if (! mfu_alltrue(fd >= 0, MPI_COMM_WORLD)) ``

src/common/mfu_util.c
daltonbohning: Luckily, I think only the d_rank_list_free was missing here, since the pool/container handles are declared statically as structs in dcp.c, and the daos calls expect them to not be NULL, so I don't believe the daos calls allocate anything else here.

src/common/mfu_util.c
daltonbohning: Yes, good catch!

src/dcp/dcp.c
daltonbohning: This is the full function:
https://github.com/daltonbohning/mpifileutils/blob/6f4a6d3252b720c770ed2bcb64c99dac3ae9f2bf/src/common/mfu_util.c#L1125-L1133

Then, I would use it like `if ! (mfu_alltrue(!local_daos_error, MPI_COMM_WORLD))`

But this function uses `MPI_INT`, `MPI_LAND`, whereas mine was using `MPI_C_BOOL`, `MPI_LOR`.
Do you know if there would be any performance difference here? I.e. perhaps MPI has an optimized implementation for `MPI_C_BOOL`, `MPI_LOR` vs `MPI_INT`, `MPI_LAND`?

daltonbohning: I actually found another hole in the logic here, where `same_pool` could potentially be mixed if a process has an error, and so the daos_connect call might be mixed. I ended up putting another reduction on `local_daos_error` before calling `daos_connect`. Even if refactored, I couldn't think of another way around this that was foolproof.

daltonbohning:
I made some style updates, corrected the "hole" with the way `local_daos_error` and `daos_connect` were being handled, and also corrected another hole with `local_daos_error` and `same_pool`.

src/dcp/dcp.c
adammoody: > 
> 
> This is the full function:
> https://github.com/daltonbohning/mpifileutils/blob/6f4a6d3252b720c770ed2bcb64c99dac3ae9f2bf/src/common/mfu_util.c#L1125-L1133
> 
> Then, I would use it like `if ! (mfu_alltrue(!local_daos_error, MPI_COMM_WORLD))`
> 
> But this function uses `MPI_INT`, `MPI_LAND`, whereas mine was using `MPI_C_BOOL`, `MPI_LOR`.
> Do you know if there would be any performance difference here? I.e. perhaps MPI has an optimized implementation for `MPI_C_BOOL`, `MPI_LOR` vs `MPI_INT`, `MPI_LAND`?
> Or even just a difference between only the datatypes? Or only the `LAND` vs `LOR`? _**Theoretically**_, `MPI_C_BOOL`, `MPI_LOR` _**could**_ be implemented in a more efficient manner.

Yes, that's right.  One could implement an LAND/LOR allreduce on a bool using a single bit versus the 32-bits an int type would require.  It should be fine to modify the mfu_alltrue function to take a bool as its input if you're up for it.

daltonbohning: Sure, I'll do that (after a quick sanity check to everything calling mfu_alltrue).

src/dcp/dcp.c
```diff
@@ -496,76 +711,68 @@ int main(int argc, char** argv)
...
+    /* If the source and destination are in the same pool,
+     * then open the container in that pool.
+     * Otherwise, connect to the second pool and open the container */
+    if (!local_daos_error && mfu_dst_file->type == DAOS) {
+        if (same_pool) {
+            /* Don't reconnect to pool, but do create container if non-existent */
+            if (rank == 0)
+                rc = daos_connect(rank, dst_svc, dst_pool_uuid, dst_cont_uuid, &src_poh, &dst_coh, false, true);
+            else
+                rc = daos_connect(rank, dst_svc, dst_pool_uuid, dst_cont_uuid, &src_poh, &dst_coh, false, true);
```
adammoody, line 735: Just another style thing is that we brace even one-line code blocks, again to avoid future bugs sneaking in where someone later adds a second line in the block without realizing the braces are missing.
```suggestion
            if (rank == 0) {
                rc = daos_connect(rank, dst_svc, dst_pool_uuid, dst_cont_uuid, &src_poh, &dst_coh, false, true);
            } else {
                rc = daos_connect(rank, dst_svc, dst_pool_uuid, dst_cont_uuid, &src_poh, &dst_coh, false, true);
            }
```

daltonbohning: This one is interesting... That was a quick debug I was doing, but then removed it. There should only be a singel `daos_connect` call there, and my local branch has the correct version...
I'm investigating now to see why git ignored these changes...

adammoody:
@daltonbohning , thanks for checking those other return code cases.  I made one more style request, but otherwise it looks good to me.

The DAOS code that leads into the tool is getting long enough that we can think about abstracting that into a function (starting at line 632 in dcp.c).  That can be done later, though.

src/dcp/dcp.c
daltonbohning: Okay this was very strange. My local repo was "up to date", but still did not match the branch on GitHub.
Anyway, I fixed it. Thanks for the catch!

daltonbohning:
@adammoody You're welcome, but really, thank you for catching this! Another set of eyes definitely helps.
I've had a similar thought about abstracting the daos-specific code. I'll go ahead and make an issue to keep track of this, and I'll make those changes when I get a chance, especially since we will be making more daos changes and additions in the future.

I pushed a couple more things in the last few minutes. If you don't find anything else, I don't have anything more to add to this PR.

adammoody:
No problem.  Those MPI corner cases can be tricky.

Thanks, @daltonbohning .  Looks good to me.  Would you mind squashing those commits into one and force pushing back?

adammoody:
Oh, you can leave the mfu_alltrue commit separate in the squash since that's a different fix.

daltonbohning:
Sorry, I'm working on this. My repo seemed to have a breakdown and it's taking a lot of time to squash these commits.
(It explains the strange out-of-sync issue I ran into earlier)

PR #376, opened 2020-09-16
daos: dcp: abstract and improve

### LINKED ISSUE #373: daos: dcp: abstract daos code into function
The DAOS-related code in dcp has grown to the point where the bulk of it should be abstracted, with functions created as necessary.

Mirrored DAOS Jira ticket:
https://jira.hpdd.intel.com/browse/DAOS-5639

### LINKED ISSUE #374: daos: warning d_rank_list_free
There is this compile warning related to DAOS:
common/mfu_util.c:140:17: warning: implicit declaration of function ‘d_rank_list_free’ [-Wimplicit-function-declaration]

This should be looked into and possibly fixed.

daltonbohning:
Three files, three commits:

**mfu_io.c**
_daos_mkdir_ - Fixed the sanity check against creating the root "/"

**mfu_util.c**
Added an include to remove a DAOS-related compile warning

**dcp.c**
Added a few more DAOS error cases.
Abstracted DAOS code into separate functions to simplify main()

closes #373 
closes #374

### DIFF
--- a/src/dcp/dcp.c
+++ b/src/dcp/dcp.c
@@ -25,30 +25,89 @@
 #include "mfu_errors.h"
 
 #ifdef DAOS_SUPPORT
-int daos_check_args(
+/* struct for holding DAOS arguments */
+typedef struct {
+    daos_handle_t src_poh; /* source pool handle */
+    daos_handle_t dst_poh; /* destination pool handle */
+    daos_handle_t src_coh; /* source container handle */
+    daos_handle_t dst_coh; /* destination container handle */
+    uuid_t src_pool_uuid;  /* source pool UUID */
+    uuid_t dst_pool_uuid;  /* destination pool UUID */
+    uuid_t src_cont_uuid;  /* source container UUID */
+    uuid_t dst_cont_uuid;  /* destination container UUID */
+    dfs_t* dfs1;           /* source dfs object */
+    dfs_t* dfs2;           /* destination dfs object */
+    char* src_svc;         /* source service level */
+    char* dst_svc;         /* destination service level */
+    char* dfs_prefix;      /* prefix for UNS */
+} daos_args_t;
+
+/* Return a newly allocated daos_args_t structure.
+ * Set default values on its fields. */
+static daos_args_t* daos_args_new(void)
+{
+    daos_args_t* da = (daos_args_t*) MFU_MALLOC(sizeof(daos_args_t));
+    
+    da->src_poh    = DAOS_HDL_INVAL;
+    da->dst_poh    = DAOS_HDL_INVAL;
+    da->src_coh    = DAOS_HDL_INVAL;
+    da->dst_coh    = DAOS_HDL_INVAL;
+    da->dfs1       = NULL;
+    da->dfs2       = NULL;
+    da->src_svc    = NULL;
+    da->dst_svc    = NULL;
+    da->dfs_prefix = NULL;
+
+    /* initalize value of DAOS UUID's to NULL with uuid_clear */
+    uuid_clear(da->src_pool_uuid);
+    uuid_clear(da->dst_pool_uuid);
+    uuid_clear(da->src_cont_uuid);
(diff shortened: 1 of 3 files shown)


src/dcp/dcp.c
```diff
@@ -25,30 +25,83 @@
...
+    uuid_clear(da->dst_cont_uuid);
+    return da;
+}
+
+/* free a daos_args_t structure */
+static void daos_args_delete(daos_args_t** da)
+{
+    if (da != NULL) {
+        mfu_free(da);
+    }
```
adammoody, line 74: Double check me on this, but I think mfu_free should account for this NULL check internally, in which case, we can simplify the code.
```suggestion
    mfu_free(da);
```

src/dcp/dcp.c
```diff
@@ -25,30 +25,83 @@
...
+    da->src_svc    = NULL;
+    da->dst_svc    = NULL;
+    da->dfs_prefix = NULL;
+
+    /* initalize value of DAOS UUID's to NULL with uuid_clear */
+    uuid_clear(da->src_pool_uuid);
+    uuid_clear(da->dst_pool_uuid);
+    uuid_clear(da->src_cont_uuid);
+    uuid_clear(da->dst_cont_uuid);
+    return da;
```
adammoody, line 66: just a style change
```suggestion
    uuid_clear(da->dst_cont_uuid);
    
    return da;
```

src/dcp/dcp.c
adammoody: Also, we should free off any strings that might have been allocated.  This could be a good place for that if you're not already doing it somewhere else.  For those, you can do something like ``mfu_free(&da->dfs1)``.  If that pointer happens to be NULL, ``mfu_free`` will skip it.

adammoody:
Thanks @daltonbohning .  This cleans the code up nicely.  Pulling this logic into functions will also help as we add DAOS support to other tools.

dsikich:
@daltonbohning this looks really good. Cleans it up a lot.

dsikich (approved):

src/dcp/dcp.c
daltonbohning: Oh, you're right about `mfu_free`. It does have a NULL check. I assumed it didn't, based on:
https://github.com/hpc/mpifileutils/blob/de44ae49129c3c7cea8342d65aab536f992b0448/src/common/mfu_flist_copy.c#L3158-L3164
But that has some other logic anyway.

I'm ashamed that I also missed freeing the strings (src_svc, dst_svc, dfs_prefix), but luckily dfs is handled in dfs_umount.

daltonbohning: Okay, this has been corrected. I looked at a lot of the other calls for mfu_*_delete and modeled after them. I.e. passing a ** and dereferencing

adammoody: Yes, good idea.  The copy_opts does the same thing.  That pattern is useful as a way to free any fields on the struct like here:

https://github.com/hpc/mpifileutils/blob/de44ae49129c3c7cea8342d65aab536f992b0448/src/common/mfu_flist_copy.c#L3209

adammoody:
One more minor thing.  I see you commit messages have some missing spaces.  After that, go ahead and hit the "Rebase and Merge" button.  We'll see if that fixes up the travis signed-by error.

daltonbohning:
> One more minor thing. I see you commit messages have some missing spaces. After that, go ahead and hit the "Rebase and Merge" button. We'll see if that fixes up the travis signed-by error.

I noticed that as well :)

I did also make one more change. After forgetting to free the daos_args_t properly, I ran valgrind and noticed our daos_* calls in mfu_io.c weren't freeing memory from the parse_filename call, so I added that as well.

daltonbohning:
Thanks @adammoody @dsikich !

adammoody:
Good idea to run valgrind.  Thanks, @daltonbohning

PR #378, opened 2020-09-18
daos: daos_mknod implementation

### LINKED ISSUE #353: daos: investigate potential daos_mknod as open/close
The mfu_mknod is not critical to dcp, but it may be required on other tools.  This is lower priority, but if dfs has no mknod, I'm wondering if it can be emulated with a dfs_open/dfs_close.

### LINKED ISSUE #377: daos: dcp: large files not copied correctly with multiple ranks
mfu_flist_copy uses a block size of 1MB. It seems that if a file to be copied is larger than 1MB, and multiple ranks are used, then this file is not **_always_** copied correctly. It appears that might be race conditions, because **_sometimes_** the file _**is**_ copied correctly.

_Steps to reproduce_:_
- Create DAOS pool and 2 containers
- dfuse mount both containers
- Create a test file
  - `truncate -s 2M  dfuse/container1/testFile` (Fills file with 0 bytes)
- Run dcp
  - ```
    mpirun -np 3 dcp --daos-src-svcl 0 \
    --daos-src-pool $pool1 --daos-src-cont $p1cont1 \
    --daos-dst-pool $pool1 --daos-dst-cont $p1cont2 \
    / /```
- Check  output file
  - `ls -l dfuse/container2/`

_Result_
_**Sometimes**_ the copied file has a size of 1MB exactly, but it should be 2MB. Running with a testFile of larger sizes, the output file always has a file size of exactly X MB. Considering the block size is 1MB, it seems there might be race conditions that prevent each block from being written, or some other similar issue.
When using a regular POSIX source and destination, I have yet to see this issue.

daltonbohning:
daos_mknod is implemented in two steps:
1. dfs_open(..., O_CREATE,...)
2. dfs_release(...)

This works for regular files only at this time.

closes #377 
closes #353 

### DIFF
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -449,10 +449,61 @@ retry:
     return rc;
 }
 
-/* use a noop, daos does not have a mknod function */
+/* emulates mknod with dfs_open, dfs_release */
 int daos_mknod(const char* path, mode_t mode, dev_t dev, mfu_file_t* mfu_file)
 {
-    return 0;
+#ifdef DAOS_SUPPORT
+    /* Only regular files are supported at this time */
+    mode_t dfs_mode = mode | S_IFREG;
+    mode_t filetype = dfs_mode & S_IFMT;
+    if (filetype != S_IFREG) {
+        MFU_LOG(MFU_LOG_ERR, "Invalid entry type (not a file)");
+        errno = EINVAL;
+        return -1;
+    }
+
+    char* name     = NULL;
+    char* dir_name = NULL;
+    parse_filename(path, &name, &dir_name);
+    assert(dir_name);
+
+    dfs_obj_t* parent = NULL;
+    int rc = dfs_lookup(mfu_file->dfs, dir_name, O_RDWR, &parent, NULL, NULL);
+    if (parent == NULL) {
+        MFU_LOG(MFU_LOG_ERR, "dfs_lookup %s failed", dir_name);
+        errno = ENOENT;
+        rc = -1;
+    }
+    else {
+        /* create regular file */
+        rc = dfs_open(mfu_file->dfs, parent, name,
+                      dfs_mode, O_CREAT,
+                      0, 0, NULL, &(mfu_file->obj));
+        if (rc) {
+            MFU_LOG(MFU_LOG_ERR, "dfs_open %s failed (%d %s)",
+                    name, rc, strerror(rc));
+            errno = rc;
+            rc = -1;
+        }
+        else {
+            /* close the file */
+            rc = dfs_release(mfu_file->obj);
+            if (rc) {
+                MFU_LOG(MFU_LOG_ERR, "dfs_release %s failed (%d %s)",
+                        path, rc, strerror(rc));
+                errno = rc;
+                rc = -1;
+            }
+        }
+        /* close the parent, ignoring errors since they have no direct impact */
+        dfs_release(parent);
+    }
+
+    mfu_free(&name);
+    mfu_free(&dir_name);
+
+    return rc;
+#endif
 }
 
 int mfu_mknod(const char* path, mode_t mode, dev_t dev)
@@ -581,9 +632,18 @@ retry:
 /*****************************
  * Files
  ****************************/
-void daos_open(const char* file, int flags, mode_t mode, mfu_file_t* mfu_file)
+int daos_open(const char* file, int flags, mode_t mode, mfu_file_t* mfu_file)
 {
 #ifdef DAOS_SUPPORT
+    /* Only regular files are supported at this time */
+    mode_t dfs_mode = mode | S_IFREG;
+    mode_t filetype = dfs_mode & S_IFMT;
+    if (filetype != S_IFREG) {
+        MFU_LOG(MFU_LOG_ERR, "Invalid entry type (not a file)");
+        errno = EINVAL;
+        return -1;
+    }
+
     char* name     = NULL;
     char* dir_name = NULL;
     parse_filename(file, &name, &dir_name);
@@ -593,7 +653,7 @@ void daos_open(const char* file, int flags, mode_t mode, mfu_file_t* mfu_file)
     int rc = dfs_lookup(mfu_file->dfs, dir_name, O_RDWR, &parent, NULL, NULL);
     if (parent != NULL) {
         rc = dfs_open(mfu_file->dfs, parent, name,
-                      S_IFREG | mode, flags,
+                      dfs_mode, flags,
                       0, 0, NULL, &(mfu_file->obj));
         if (rc) {
             MFU_LOG(MFU_LOG_ERR, "dfs_open %s failed (%d %s)",
@@ -609,6 +669,8 @@ void daos_open(const char* file, int flags, mode_t mode, mfu_file_t* mfu_file)
 
     mfu_free(&name);
     mfu_free(&dir_name);
+
+    return rc;
 #endif
 }
 
--- a/src/common/mfu_io.h
+++ b/src/common/mfu_io.h
@@ -114,7 +114,7 @@ int mfu_symlink(const char* oldpath, const char* newpath);
 
 /* open file with specified flags and mode, retry open a few times on failure */
 void mfu_file_open(const char* file, int flags, mfu_file_t* mfu_file, ...);
-void daos_open(const char* file, int flags, mode_t mode, mfu_file_t* mfu_file);
+int daos_open(const char* file, int flags, mode_t mode, mfu_file_t* mfu_file);
 int mfu_open(const char* file, int flags, ...);
 
 /* close file */


daltonbohning:
closes #377

adammoody:
Thanks @daltonbohning .  You're really knocking out a lot of items over this past week!  Nice!

It's fine to just support regular files for now, especially since the tools don't generally support types other than directories, regular files, and symlinks just yet.  However, let's add a check and print a warning message to alert the user if this mknod function happens to get called with something other than a regular file.  In that case, perhaps we could also set errno and return -1 if there is good errno value that matches up.

Also, it'd be helpful for future developers to add a comment in the function just to mention that the goal is to emulate a mknod operation using an open/release.

adammoody:
If printing/returning errors would cause problems, like lots of harmless warnings under typical usage, we can reconsider.  I think the main reason we call mknod right now is to set lustre striping on files, so it's kind of a special case.  We could double check the rest of mpiFileUtils code to be sure.

daltonbohning:
I revised my comment to be more clear, after some further investigation into dfs_open.
dfs_open allows regular files, symbolic links, and directories. If anything else is supplied, dfs_open returns an appropriate error, which daos_mknod captures and propogates:
https://github.com/daos-stack/daos/blob/73b42ec7870d74091a074e2b189f8d2d48f4fd53/src/client/dfs/dfs.c#L2315-L2340

daltonbohning:
As for mpifileutils calling mknod, I found this was necessary for DAOS (and possible regular POSIX as well, though I'm not sure).
I ran into an issue where if multiple processes were writing different chunks to the same file, and that file did not yet exist, then the destination file would _sometimes_ not contain all of the written chunks. But if the destination file did exist before copying any chunks, then there were no issues.
mfu_flist_copy calls mfu_file_create, which calls mfu_file_mknod BEFORE copying any data. And since mknod creates the file, it solves this issue.

src/common/mfu_io.c
```diff
@@ -449,10 +449,50 @@ int mfu_lstat64(const char* path, struct stat64* buf)
...
+    char* dir_name = NULL;
+    parse_filename(path, &name, &dir_name);
+    assert(dir_name);
+
+    dfs_obj_t* parent = NULL;
+    int rc = dfs_lookup(mfu_file->dfs, dir_name, O_RDWR, &parent, NULL, NULL);
+    if (parent != NULL) {
+        /* create regular file */
+        rc = dfs_open(mfu_file->dfs, parent, name,
+                      S_IFREG | mode, O_CREAT,
```
adammoody, line 467: I know @dsikich mentioned adding ``S_IFREG`` was needed here but now I forget why.  Can you also add a comment about that here so I don't forget again?

adammoody:
> 
> 
> As for mpifileutils calling mknod, I found this was necessary for DAOS (and possible regular POSIX as well, though I'm not sure).
> I ran into an issue where if multiple processes were writing different chunks to the same file, and that file did not yet exist, then the destination file would _sometimes_ not contain all of the written chunks. But if the destination file did exist before copying any chunks, then there were no issues.
> mfu_flist_copy calls mfu_file_create, which calls mfu_file_mknod BEFORE copying any data. And since mknod creates the file, it solves this issue.

Interesting.  Sounds like a race condition on creating the "inode" or something.  Based on your debugging, it sounds like this avoids the problem.  Do you know whether that is expected behavior in DAOS?

daltonbohning:
> Interesting. Sounds like a race condition on creating the "inode" or something. Based on your debugging, it sounds like this avoids the problem. Do you know whether that is expected behavior in DAOS?

I checked with regular POSIX files (removing the `mknod` call) and was not able to reproduce the issue.

I'm not sure if it is expected behavior in DAOS, but since POSIX works without issues, it seems `daos_mknod` "indirectly" corrects the issue. I would like to find a more proper fix for this. I'll need to do some further digging. It may be that the `dfs` library just isn't explicitly handling race conditions, since it doesn't use MPI, etc.

src/common/mfu_io.c
dsikich: @adammoody if you don't specify S_IFREG then dfs_open will return with an error saying "invalid inputs." It expects you to tell it if it is a regular file.

dsikich:
> > Interesting. Sounds like a race condition on creating the "inode" or something. Based on your debugging, it sounds like this avoids the problem. Do you know whether that is expected behavior in DAOS?
> 
> I checked with regular POSIX files (removing the `mknod` call) and was not able to reproduce the issue.
> 
> I'm not sure if it is expected behavior in DAOS, but since POSIX works without issues, it seems `daos_mknod` "indirectly" corrects the issue. I would like to find a more proper fix for this. I'll need to do some further digging. It may be that the `dfs` library just isn't explicitly handling race conditions, since it doesn't use MPI, etc.

@adammoody @daltonbohning I can't say for sure either. It is a bit puzzling to me since I would think that on the open call in mfu_copy_open_file it would open and create the file if it doesn't exist, then write to its particular offset with the length of bytes it was given.

daltonbohning:
> I can't say for sure either. It is a bit puzzling to me since I would think that on the open call in mfu_copy_open_file it would open and create the file if it doesn't exist, then write to its particular offset with the length of bytes it was given.

I'll look into this as well, to rule out `mfu` as being the culprit.

src/common/mfu_io.c
daltonbohning: Thanks @dsikich 
I'll check on this for my own information and leave an appropriate comment.

dsikich (approved):

src/common/mfu_io.c
dsikich: @daltonbohning yeah if you go tohttps://github.com/daos-stack/daos/blob/master/src/client/dfs/dfs.c in dfs_open it has a switch statement that checks the file type. If it doesn't match then it returns EINVAL, maybe there is a way around passing it in explicitly, not %100 sure on that.

daltonbohning: (Please let me know if any of this doesn't make sense)

So the call to `mfu_file_mknod` within `mfu_flist_copy` actually passes in `S_IFREG` directly:
https://github.com/daltonbohning/mpifileutils/blob/cb2d35c20662569dc1385f375fe17e267b898b90/src/common/mfu_flist_copy.c#L1130

But the definition for `mfu_file_open` does not have a `mode` parameter directly. Instead, `mode` can be passed to `mfu_file_open` through "variable arguments", and not all calls to `mfu_file_open` pass a `mode`, and hence do not pass `S_IFREG`, which is why `daos_open` uses `S_IFREG | mode`, in case it was not passed in.
https://github.com/daltonbohning/mpifileutils/blob/cb2d35c20662569dc1385f375fe17e267b898b90/src/common/mfu_io.c#L706

`mfu_open` checks for variable arguments, and passes `mode` to `open` only if `O_CREAT` is specified, as is required by `open`:
https://man7.org/linux/man-pages/man2/open.2.html
If `mode` is not passed to `open`, and the file already exists, it doesn't matter to POSIX, because it will open the file of whatever type it is.

But in DAOS, `dfs_open` DOES require `mode` with the filetype, even if the file already exists.

POSIX `mknod`, however, always requires `mode` to be passed in to specify the file type.

But I wonder if doing `S_IFREG | mode` might be a problem if `mode = S_IFDIR` is passed into `daos_open`?
Since `dfs_open` only supports `S_IFREG`, `S_IFDIR`, `S_IFLNK`, maybe `daos_mknod` and/or `daos_open` should more appropriately have something like this (but in a better-coded way):
```
switch (mode & S_IFMT) {
	case S_IFREG:
	        // good
	case S_IFDIR:
	        // good
	case S_IFLNK:
	        // good
	default:
                set mode = S_IFREG
```
This would allow any of `S_IFREG`, `S_IFDIR`, `S_IFLNK` to be be passed to `daos_open` or `daos_mknod`, and would only default `mode` if neither are passed in.

daltonbohning: Update: `dfs_open` requires `value` to be set if the type is `S_IFLINK`, so maybe we would only allow `S_IFREG`, `S_IFDIR` for now.

dsikich: @daltonbohning not sure why we would ever call dfs_open on a directory? we should be using daos_mkdir, daos_opendir for those operations I would think.

daltonbohning: In that case, I think the `S_IFREG` would then be the only one allowed, as it is now.
And, as Adam suggested previously, I think I'll add a check to make sure that's the only type passed to `mknod`, and leave `daos_open` with the `S_IFREG | mode`, as you originally put in.
(But I'll leave some comments so we don't need this discussion again)

daltonbohning:
> las p1cont1 p1cont2; rm p1cont2/a

@adammoody Turns out this was in fact a bug in `dfs.c` that was fixed on Thursday. But we needed the `daos_mknod` implementation anyway, so that's a win!

daltonbohning:
> If printing/returning errors would cause problems, like lots of harmless warnings under typical usage, we can reconsider. I think the main reason we call mknod right now is to set lustre striping on files, so it's kind of a special case. We could double check the rest of mpiFileUtils code to be sure.

I think it will be okay to do as you suggest, since `mfu_file->type == DAOS` is really only being used from `dcp.c` at this time, and the DAOS team (Danielle and I) will have to run our tests again if ever mpifileutils is updated, so we should catch any future issues.

src/common/mfu_io.c
dsikich: @daltonbohning yeah that sounds good

daltonbohning:
@adammoody I think I've finished with this one now. I added a little more cleanup, and switched an error check in the spirit of "keeping an `rc` check close to the function call", i.e. `if parent == NULL then error` instead of `if parent != NULL... else error`

src/common/mfu_io.c
```diff
@@ -609,6 +669,8 @@ void daos_open(const char* file, int flags, mode_t mode, mfu_file_t* mfu_file)
 
     mfu_free(&name);
     mfu_free(&dir_name);
+
+    return -1;
```
adammoody, line 673: Should this be ``return rc``?

Also, if ``#ifdef DAOS_SUPPORT`` is not active, I'm guessing some compilers will not be happy with this function, since it is declared to return an int, but it wouldn't have a return statement in that case.  To fix that, you could add an ``#else`` case or perhaps but an ``int rc = 0;`` at the top and an ``return rc;`` at the bottom?

src/common/mfu_io.c
```diff
@@ -609,6 +669,8 @@ void daos_open(const char* file, int flags, mode_t mode, mfu_file_t* mfu_file)
 
     mfu_free(&name);
     mfu_free(&dir_name);
```
adammoody, line 671: Oh, we should have to free up these strings under different error cases.

src/common/mfu_io.c
adammoody: On second thought, I see the ``#ifdef`` is the same on all daos functions, so let's wait to do that in a future PR if it's actually needed.

daltonbohning: > Should this be `return 0`?
Yes, that was careless of me... I'll need to investigate my tests didn't catch this...
But it has been fixed.

daltonbohning: > `int rc = 0;` at the top and an `return rc;` at the bottom

I think this should be easy enough to do. I can do that when I also go through this one:
https://github.com/hpc/mpifileutils/issues/382
Since I will be touching most functions anyway.

src/common/mfu_io.c
daltonbohning: I think this should be fine, no? The error cases fall through to the `mfu_free`s, except the error case before `name` and `dir_name` are declared.

adammoody: Yes, you're right.  Sorry, I was skimming too quickly and misreading those ``rc = -1`` statements in the errors as ``return -1``.

Good to rebase and merge.

Thanks!

daltonbohning: That's okay :)
Thanks for all the feedback!

PR #380, opened 2020-09-19
mfu: account for short read/writes with O_DIRECT

adammoody:


src/common/mfu_io.c
```diff
@@ -944,6 +944,150 @@ ssize_t daos_write(const char* file, const void* buf, size_t size, mfu_file_t* m
...
+        MFU_LOG(MFU_LOG_ERR, "dfs_read %s failed (%d %s)",
+            file, rc, strerror(rc)
+        );
+        errno = rc;
+        return -1;
+    }
+
+    return (ssize_t)got_size;
+#endif
+}
```
adammoody, line 987: @dsikich @daltonbohning I'm working to fix up our copy code to handle short read/write operations.  With that I'm replacing some ``mfu_file_read/write`` with new ``mfu_file_pread/pwrite`` calls.  A few questions:

1) When this is ready, would you please check that these functions actually work as expected on DAOS?
2) Does the ``d_iov_set`` call set the size field in the iov struct, and if so, can we drop the ``sgl.sg_iovs[0].iov_len = size;`` line?
3) Can ``dfs_write`` return with a short write?
4) Are there ``dfs_pread/pwrite`` calls we should use instead?

daltonbohning: @adammoody I'll take a look at this today.

daltonbohning: @adammoody 
2. Yes, it seems it does set the size:
https://github.com/daos-stack/daos/blob/1e221bdaf079092cdf26db563727a775864e1b29/src/include/gurt/types.h#L114-L119

3. I don't **think** it can return with a short write the same way `pwrite` does. If you tell it to write a certain amount, it seems to be all or nothing. Though, I'm not 100% familiar with this, and found myself tracing through several files of unfamiliar DAOS library code. If it IS possible, then I think the actual size written would be returned in `sg_iovs[0].iov_len`, since this is passed as a pointer. Worse case, this value would just be the same as the input.

4. There aren't `dfs_pread/dfs_pwrite` calls, since `dfs_read/dfs_write` always expect an `offset`.

adammoody: Thanks @daltonbohning .  I'll drop that line that is setting the length again.  Let me know if you find that breaks something.

If you or @dsikich later find that short writes *are* possible, we should update the ``daos_write`` and the new ``daos_pwrite`` call to account for that.  Right now those each assume all bytes were written if there is no error.  On the flip side, if you are sure short writes aren't possible, we can add a comment for future code reviewers.

daltonbohning: I'll leave myself a note to check into this some more, and ask someone who is familiar with the code.
In the meantime, I don't think it would _hurt_ to return `sg_iovs[0].iov_len` instead of `size`, since `sg_iovs[0].iov_len` is set to `size`, although it might be ambiguous why that is being done... What do you think?

adammoody: @dsikich , @daltonbohning .  Alright, I think is ready for a daos test now.  For ``mfu_flist_copy`` (used by ``dcp`` and ``dsync``) and ``mfu_compare_contents`` (used by ``dcmp`` and ``dsync -c``), this uses ``pread/pwrite`` calls with corresponding daos versions.  Could you please give this a whirl and verify whether it works ok?

adammoody: > 
> 
> I'll leave myself a note to check into this some more, and ask someone who is familiar with the code.
> In the meantime, I don't think it would _hurt_ to return `sg_iovs[0].iov_len` instead of `size`, since `sg_iovs[0].iov_len` is set to `size`, although it might be ambiguous why that is being done... What do you think?

I think I follow your suggestion here, but since I can't test the changes, I left it as is for now.

daltonbohning: @adammoody 
I ran these change will all of the current DAOS-related `dcp` tests, and everything looks good.

I dug some more into whether `dfs_write` could return a short read. As far as I'm aware, it doesn't, and the documentation also makes no mention of it, so I'm going to assume that it does _**not**_.
Maybe we can leave a comment stating this, so we don't forget?

Otherwise, I don't see anything in this code that might cause as issue.

PR #385, opened 2020-09-23
dcmp, dsync: add --chunksize and --blocksize

adammoody:
Now that ``dcmp`` and ``dsync`` both allocate a ``copy_opts`` struct, this adds ``--blocksize`` and ``--chunksize`` options for those tools.


adammoody:
That could be the case.  I'll take a look.  I also want to consider updated the defaults from 1MB to something bigger, andI need to do some experimentation to pick some new values.

daltonbohning:
> That could be the case. I'll take a look. I also want to consider updated the defaults from 1MB to something bigger, andI need to do some experimentation to pick some new values.

Thank you! And for what it's worth, I don't think the default would have a _functional_ impact on DAOS, but of course I can run some tests on my end if you need me to.

adammoody:
> 
> 
> > That could be the case. I'll take a look. I also want to consider updated the defaults from 1MB to something bigger, andI need to do some experimentation to pick some new values.
> 
> Thank you! And for what it's worth, I don't think the default would have a _functional_ impact on DAOS, but of course I can run some tests on my end if you need me to.

Ok, I think I found and fixed the spot that was overwriting the block_size value that had been set on the command line.  Can you give that another try and verify it's working for you now?

Yeah, I can imagine it might not change much in DAOS, but it's good to know that I didn't break something in DAOS even if it doesn't change performance.  Along those lines, does DAOS deal with someone opening a file using the O_DIRECT option?  This could be tested on this branch by adding the ``--direct`` option in dcp.

daltonbohning:
> Ok, I think I found and fixed the spot that was overwriting the block_size value that had been set on the command line. Can you give that another try and verify it's working for you now?

I think this is good!

> does DAOS deal with someone opening a file using the O_DIRECT option?

From what I gathered, it seems O_DIRECT is ignored by the `DFS` api, since the data is stored in DAOS objects. In the `DFUSE` api (DAOS mounts), it seems like O_DIRECT is handled in some cases, but then in others `DFUSE` overrides the O_DIRECT flag to always be true.
Overall, it seems like DAOS just always handles its own buffering, and thus either ignores or forces O_DIRECT.
But I'm not very familiar with this (I started on DAOS around the time I first started contributing to mpifileutils)
Either way, running `dcp` with `--direct` doesn't seem to break anything for DAOS, so that's good.

adammoody:
I'll merge this early since it fixes a bug and enables one to set chunksize/blocksize in dcmp and dsync now.  I may change the default from 64MB to something else after testing.

Issue #384, opened 2020-09-22
daos: mfu_io: enhance lookup_insert_dir


daltonbohning:
Within `mfu_io` there is a function `lookup_insert_dir` that uses a hash table to speedup the lookups of DAOS container directories. 

1. Currently, this is only used when only the src is DAOS or only the dst is DAOS. We should make some modifications that allow it to be used when both the src and dst are DAOS.

2. The `fprintf(stderr, ...)` calls should be updated to `MFU_LOG(MFU_LOGG_ERR, ...)`.

3. We need to find the proper way to add a corresponding `d_hash_table_destroy` call to handle cleanup.

4. We should investigate the possibility of using the hash table(s) elsewhere. Any of the `daos_*` IO functions could potentially be candidates.

Mirrored DAOS Jira ticket:
https://jira.hpdd.intel.com/browse/DAOS-5722

daltonbohning:
@dsikich

### Rough Proposal
**Step-by-step explanation**
1. There is a 1:1 relationship between a DAOS container and a `dfs` mount (the way we are using it, anyway).
2. Each container should have its own `d_hash_table` to avoid conflicts.
3. The `dfs` is stored with the `mfu_file` record.
4. We need to pass the `mfu_file` to `lookup_insert_dir` to get the `dfs`.
5. We can add a new entry `mfu_file_t->dir_hash` to hold the hash for that corresponding `dfs`.
6. When we call `dfs_mount`, we can also set the corresponding `mfu_file_t->dir_hash`.
7. We can create a wrapper function for this, for simplicity and reusability.
8. When we call `dfs_umount`, we can cleanup the `mfu_file_t->dir_hash`.
9. We can create a wrapper function for this as well, to keep the creation and destruction of the hash in sync.
10. At this point, each `mfu_file_t` would "automatically" have a `mfu_file_t->dir_hash`.
11. Now to utilize it...
12. All `dfs_lookup` calls within `mfu_io` are _similar_ (permissions, etc.) except for one that also passes `mode`, but ignore this for now.
13. We can replace _**all**_ `dfs_lookup` calls with a call to `lookup_insert_dir`.
14. `lookup_insert_dir` would get/insert into the corresponding `mfu_file->dir_hash`.
15. Now all lookups will use the hash.
16. This also has other (positive) implications.
17. The `dir_hash` will now be cleaned up properly (currently isn't) when we call `dfs_umount` (in our new wrapper function).
[...]

PR #386, opened 2020-09-23
daos: robust enhancements

### LINKED ISSUE #352: daos: add mfu_file_unlink and daos_unlink
The copy function might eventually call mfu_unlink on destination files if it detects an error after performing the copy.

https://github.com/hpc/mpifileutils/blob/a059f299c3b0d4f4409a0ce48360a1081645d1b4/src/common/mfu_flist_copy.c#L2125

To handle this, we'll want to add an mfu_file_unlink and daos_unlink.

This code is commented out now.  I forget why I left that commented, other than maybe I'm paranoid about deleting things.

### LINKED ISSUE #382: daos: mfu_io: call dfs_release for each dfs_lookup
Based on the documentation for `dfs_lookup`, "the object must be released with `dfs_release`":
https://github.com/daos-stack/daos/blob/0006b6472c7624b4f1e79505678613dcc6c3a50d/src/include/daos_fs.h#L189-L204

Within `mfu_io.c`, each call to `dfs_lookup` should have a matching `dfs_release` call once we are done with the returned object (usually `parent`).

Also, to be consistent with other forms of error checking, we should change each 
`if (parent != NULL) {// all good} else {// ERROR}`
to 
`if (parent == NULL) {// ERROR} else {// all good}`

### LINKED ISSUE #383: daos: mfu_io: daos_* functions return an int when no DAOS_SUPPORT
If `#ifdef DAOS_SUPPORT` is not active, some compilers might have an issue with the `daos_*` functions that return an int, since they would not return one.
We could add an `#else` case or perhaps an `int rc = 0;` at the top and an `return rc;` at the bottom.

daltonbohning:
Three commits, three issues.

>**daos: mfu_file_unlink and daos_unlink implementation**
>Added implementation for mfu_file_unlink and daos_unlink.
>Updated mfu_flist_copy.c and mfu_flist_walk.c to use the
>new mfu_file_unlink.
>
>The following files still have references to mfu_unlink,
>but I don't believe that can/should be updated:
>mfu_flist_remove.c
>dbcast.c
>dreln.c
>copy.c
>dbz2.c

closes #352 

>**daos: mfu_io: add dfs_release for each dfs_lookup**
>Each daos_* function that uses dfs_lookup now also
>has a corresponding dfs_release, except for "open"
>functions, since this is handled in "close".

One exception: calls made in `lookup_insert_dir` do not have a corresponding `dfs_release`, since this will require more effort to fix. New issue #384 created to handle this.

closes #382 

>**daos: mfu_io: return 0 or NULL if not DAOS_SUPPORT**
>For daos_* functions, when DAOS_SUPPORT is not defined,
>the functions will just return 0 or NULL to alleviate
>compiler warnings/errors.

I opted for the `#else` approach here, since I felt it was more explicit and less error prone than the `int rc = 0; #ifdef DAOS_SUPPORT...#endif return rc`
 approach.

closes #383

### DIFF
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -1227,6 +1227,67 @@ int mfu_file_ftruncate(mfu_file_t* mfu_file, off_t length)
     }
 }
 
+/* unlink a file */
+int mfu_file_unlink(const char* file, mfu_file_t* mfu_file)
+{
+    if (mfu_file->type == POSIX) {
+        int rc = mfu_unlink(file);
+        return rc;
+    } else if (mfu_file->type == DAOS) {
+        int rc = daos_unlink(file, mfu_file);
+        return rc;
+    } else {
+        MFU_ABORT(-1, "File type not known, type=%d",
+                  mfu_file->type);
+    } 
+}
+
+/* emulates unlink on a DAOS file or symlink */
+int daos_unlink(const char* file, mfu_file_t* mfu_file)
+{
+#ifdef DAOS_SUPPORT
+    char* name     = NULL;
+    char* dir_name = NULL;
+    parse_filename(file, &name, &dir_name);
+    assert(dir_name);
+
+    /* Need to lookup parent directory in DFS */
+    dfs_obj_t* parent = NULL;
+    mode_t mode;
+    int rc = dfs_lookup(mfu_file->dfs, dir_name, O_RDWR, &parent, &mode, NULL);
+    if (parent == NULL) {
+        MFU_LOG(MFU_LOG_ERR, "dfs_lookup %s failed", dir_name);
+        errno = ENOENT;
+        rc = -1;
+    } 
+    else {
+        /* only allow regular files and symlinks */
+        if (!S_ISREG(mode) && !S_ISLNK(mode)) {
+            MFU_LOG(MFU_LOG_ERR, "Invalid entry type (not a file or symlink)");
+            errno = EINVAL;
+            rc = -1;
+        }
+        else {
+            /* remove the file */
+            rc = dfs_remove(mfu_file->dfs, parent, name, false, NULL);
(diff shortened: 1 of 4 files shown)


dsikich:
@daltonbohning I liked the #else approach better here too.

dsikich (approved):

src/common/mfu_io.c
```diff
@@ -282,24 +285,28 @@ int daos_chmod(const char *path, mode_t mode, mfu_file_t* mfu_file)
 
     dfs_obj_t* parent = NULL;
     int rc = dfs_lookup(mfu_file->dfs, dir_name, O_RDWR, &parent, NULL, NULL);
-    if (parent != NULL) {
+    if (parent = NULL) {
```
adammoody, line 288:
```suggestion
    if (parent == NULL) {
```

src/common/mfu_io.c
```diff
@@ -1183,12 +1216,14 @@ int daos_ftruncate(mfu_file_t* mfu_file, off_t length)
     daos_off_t offset = (daos_off_t) length;
     int rc = dfs_punch(mfu_file->dfs, mfu_file->obj, offset, DFS_MAX_FSIZE);
     if (rc) {
-        MFU_LOG(MFU_LOG_ERR, "dfs_punch failed (%d)",
+        MFU_LOG(MFU_LOG_ERR, "dfs_punch failed (%d %s)",
```
adammoody, line 1219: Nice catch!

daltonbohning: Only because I was trying to copy/paste :)

src/common/mfu_io.c
daltonbohning: Thanks!

adammoody:
Looks good to me.  Thanks!

PR #388, opened 2020-09-25
mfu: support aarch64

tkameyama:
aarch64 kernel is not found SYS_getdents.
If SYS_getdents is not defined, use sys_getdents64.

Signed-off-by: Toyohisa Kameyama <kameyama@riken.jp>


adammoody:
Thank you, @tkameyama .

PR #389, opened 2020-09-28
daos: enhance use of lookup hash

### LINKED ISSUE #384: daos: mfu_io: enhance lookup_insert_dir

### LINKED ISSUE #387: daos: mfu_io: fix daos_unlink type check
I realize I misused `dfs_lookup` when I wrote the `daos_unlink` implementation.
When checking `if (!S_ISREG(mode) && !S_ISLNK(mode))`, this mode actually corresponds to the _**parent**_ of the file to be unlinked, not the file itself.
We would need either (1) another call to `dfs_lookup` on the file itself, (2) some other `dfs` api call that gives the mode of the file, or (3) we could just allow `daos_unlink` to also work on directories.
Preferably, (3) should only be considered as a last option.

daltonbohning:
**_Question_** @adammoody
As the DAOS-related code grows, would you prefer we move DAOS-specific functions to some new `daos_util.[h,c]`? E.g. hash functions, setup functions.
(But keep the `daos_*` IO functions in `mfu_io.c`)

lookup_insert_dir and associated functions have been
refactored and expanded.

*mfu_param_path.h*
- Added dfs_hash to mfu_file.
- Removed only_daos from mfu_file, since it is no longer needed

*Naming convention*
- Changed mfu_* to daos_* since they are only used by DAOS.
- Changed lookup_insert_dir to daos_lookup_hash.

*daos_lookup_hash and associated functions*
- Gets the hash from mfu_file instead of a global.
- Uses a pointer to the name instead of static allocation,
  which saves space.
- Updated error messages to use MFU_LOG instead of stderr directly.
- Added a _new and _delete function for the object handle struct.
- Properly deallocate and free each entry.

*daos_ IO functions*
With the excpetion of daos_opendir and daos_closedir,
all daos_ IO functions now use daos_hash_lookup to
get directory entries.

*mfu_util*
- Added daos_mount as a wrapper for dfs_mount.
- Added daos_umount as a wrapper for dfs_umount.
  This also destroys the hash.

*dcp.c*
Now uses the new daos_mount and daos_umount functions.

closes #384 
closes #387

### DIFF
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -25,81 +25,136 @@
 static int mpi_rank;
 
 #ifdef DAOS_SUPPORT
-struct d_hash_table *dir_hash;
-
-struct mfu_dir_hdl {
-        d_list_t	entry;
-        dfs_obj_t	*oh;
-        char		name[PATH_MAX];
+/* Handle for a hash table entry */
+struct daos_dir_hdl {
+    d_list_t    entry;
+    dfs_obj_t*  oh;
+    char*       name;
 };
 
-static inline struct mfu_dir_hdl* hdl_obj(d_list_t *rlink)
+/* Return a newly allocated daos_dir_hdl structure */
+static struct daos_dir_hdl* daos_dir_hdl_new(void)
 {
-        return container_of(rlink, struct mfu_dir_hdl, entry);
+    struct daos_dir_hdl* hdl = (struct daos_dir_hdl*) MFU_MALLOC(sizeof(struct daos_dir_hdl));
+    hdl->oh = NULL;
+    hdl->name = NULL;
+
+    return hdl;
 }
 
-static bool key_cmp(struct d_hash_table *htable, d_list_t *rlink,
-	const void *key, unsigned int ksize)
+/* free a daos_dir_hdl structure */
+static void daos_dir_hdl_delete(struct daos_dir_hdl** phdl)
 {
-        struct mfu_dir_hdl *hdl = hdl_obj(rlink);
-
-        return (strcmp(hdl->name, (const char *)key) == 0);
-}
-
-static void rec_free(struct d_hash_table *htable, d_list_t *rlink)
-{
-        struct mfu_dir_hdl *hdl = hdl_obj(rlink);
-
-        assert(d_hash_rec_unlinked(&hdl->entry));
-        dfs_release(hdl->oh);
-        free(hdl);
-}
-
-static d_hash_table_ops_t hdl_hash_ops = {
-        .hop_key_cmp	= key_cmp,
-        .hop_rec_free	= rec_free
-};
-
-static dfs_obj_t* lookup_insert_dir(const char *name, mfu_file_t* mfu_file)
-{
-        struct mfu_dir_hdl *hdl;
-        d_list_t *rlink;
-        int rc;
-
-        /* TODO: need to make sure we have a corresponding d_hash_table_destroy */
-	if (dir_hash == NULL) {
-	    rc = d_hash_table_create(0, 16, NULL, &hdl_hash_ops, &dir_hash);
-	    if (rc) {
-		    fprintf(stderr, "Failed to initialize dir hashtable");
-		    return NULL;
-	    }
-	}
-
-        rlink = d_hash_rec_find(dir_hash, name, strlen(name));
-        if (rlink != NULL) {
-                hdl = hdl_obj(rlink);
-                return hdl->oh;
+    if (phdl != NULL) {
+        struct daos_dir_hdl* hdl = *phdl;
+        if (hdl->oh != NULL) {
+            dfs_release(hdl->oh);
         }
+        mfu_free(&hdl->name);
+        mfu_free(phdl);
+    }
+}
 
-        hdl = calloc(1, sizeof(struct mfu_dir_hdl));
-        if (hdl == NULL)
-		return NULL;
+/* Get the daos_dir_hdl from its entry */
+static inline struct daos_dir_hdl* hdl_obj(d_list_t* rlink)
+{
+    return container_of(rlink, struct daos_dir_hdl, entry);
+}
 
-        strncpy(hdl->name, name, PATH_MAX-1);
-        hdl->name[PATH_MAX-1] = '\0';
+/* Simple string comparison of hdl->name as the key */
+static bool key_cmp(struct d_hash_table* htable, d_list_t* rlink, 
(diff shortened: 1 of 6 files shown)


src/common/mfu_io.c
```diff
@@ -25,81 +25,136 @@
...
+static void daos_dir_hdl_delete(struct daos_dir_hdl** phdl)
 {
-        struct mfu_dir_hdl *hdl = hdl_obj(rlink);
+    if (phdl != NULL) {
+        struct daos_dir_hdl* hdl = *phdl;
+        if (hdl->oh != NULL) {
+            dfs_release(hdl->oh);
+        }
+        mfu_free(&hdl->name);
+        mfu_free(&hdl);
```
adammoody, line 54: Let's pass the caller's argument here rather than the address of the local variable.
```suggestion
        mfu_free(phdl);
```
The difference is that this will set the caller's pointer to NULL.

src/common/mfu_util.c
```diff
@@ -195,6 +197,45 @@ int daos_connect(
...
+  daos_handle_t* poh,
+  daos_handle_t* coh)
+{
+    int rc = 0;
+
+    /* Mount dfs */
+    rc = dfs_mount(*poh, *coh, O_RDWR, &mfu_file->dfs);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to mount DAOS filesystem (DFS): "
+                MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS));
```
adammoody, line 212: I haven't checked these macros, but do we need corresponding printf-type codes up in the error string, like %d and %s?

src/common/mfu_util.c
```diff
@@ -195,6 +197,45 @@ int daos_connect(
...
+}
+
+int daos_umount(
+  mfu_file_t* mfu_file)
+{
+    int rc = 0;
+    int tmp_rc;
+    
+    /* Unmount dfs */
+    tmp_rc = dfs_umount(mfu_file->dfs);
```
adammoody, line 226: Just a style thing here in trying to declare locals where they are first used.
```suggestion
    
    /* Unmount dfs */
    int tmp_rc = dfs_umount(mfu_file->dfs);
```

adammoody:
Thanks @daltonbohning .  This includes a bunch of nice additions, and thanks for the detailed PR comments.  This all looks good to me other than a few comments.

Yes, I think we as grow the DAOS code, it'll make sense to create a daos_util file to hold the various DAOS-specific functions.

src/common/mfu_util.c
daltonbohning: Ah, yes. It has been hammered in my brain to do it the other way. But I actually do prefer this way.

src/common/mfu_util.c
daltonbohning: I modeled this after the way DAOS does it, albeit simplified:
```
#define MFU_ERRF "%s(%d)"
#define MFU_ERRP(rc) "MFU_ERR", rc
```
Means that:
```
MFU_LOG(MFU_LOG_ERR, "Failed to mount DAOS filesystem (DFS): "
                MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS));
```
Evaluates to:
```
MFU_LOG(MFU_LOG_ERR, "Failed to mount DAOS filesystem (DFS): "
                "%s(%d)", "MFU_ERR", -MFU_ERR_DAOS);
```
Simplified:
```
MFU_LOG(MFU_LOG_ERR, "Failed to mount DAOS filesystem (DFS): %s(%d)"
                "MFU_ERR", -MFU_ERR_DAOS);
```

In the future, if desired, we could change:
```
#define MFU_ERRP(rc) "MFU_ERR", rc
```
To instead show the string corresponding to that rc value, instead of just always "MFU_ERR". DAOS does this, but as you can imagine, associating a string with an rc value, and the rc value with the same string, through macros, will take a little bit more work. DAOS handles this by have macros that "register" each error code with a string.

src/common/mfu_io.c
daltonbohning: Ah, yes. Thank you!

daltonbohning:
Thanks @adammoody! I'll get these fixed tomorrow, and figure out when/how to best abstract to `daos_util`

src/common/mfu_util.c
adammoody: Ah, nice.  Yes, that's all fine then.

Thanks for the explanation.

dsikich:
@daltonbohning looks like one of the builds is failing

daltonbohning:
@dsikich Hmm. It looks like it was only one of the builds, not both. It seems the `wget lwgrp` failed (network issue?)
I'll rerun this to see if it was just a fluke in the network.

$ lwgrp=lwgrp-1.0.2
0.07s$ wget https://github.com/llnl/lwgrp/releases/download/v1.0.2/${lwgrp}.tar.gz
--2020-09-29 16:10:06--  https://github.com/llnl/lwgrp/releases/download/v1.0.2/lwgrp-1.0.2.tar.gz
Resolving github.com (github.com)... 140.82.112.3
Connecting to github.com (github.com)|140.82.112.3|:443... connected.
HTTP request sent, awaiting response... 429 too many requests
2020-09-29 16:10:06 ERROR 429: too many requests.
The command "wget https://github.com/llnl/lwgrp/releases/download/v1.0.2/${lwgrp}.tar.gz" failed and exited with 8 during .

daltonbohning:
@dsikich Looks like it built this time! Do you see anything in the code that might be an issue either now or in the future?
(I have ran all the current functional tests against this, with no failures)

dsikich (approved):

dsikich:
@daltonbohning no, I don't see anything that stands out to me. Thanks!

daltonbohning:
@dsikich Thanks!

Issue #400, opened 2020-10-02
directory permissions incorrectly set before the subtree has been fully copied


akesandgren:
Using version 0.10.1

When I have a directory in the source tree which isn't writable, the target directory is created without write permission.
This is wrong.
The target dir must be created with write permission so that the files and subdirs can be copied/created.
Directory permissions must not be set/preserved until the subtree has finished copying

```
constant:
total 56
dr-xr-x---+ 3 ake folk 4096 Apr  3  2017 boundaryData/
-r--r-x---+ 1 ake folk 1546 Jul 20  2017 chemistryProperties*
-r--r-x---+ 1 ake folk 1109 Jul 20  2017 combustionProperties*

constant/boundaryData:
total 316
dr-xr-x---+ 5490 ake folk 319488 Jul 20  2017 inlet1/
constant/boundaryData/inlet1:
total 22096
dr-xr-x---+ 2 ake folk  4096 Jul 20  2017 0.0/
dr-xr-x---+ 2 ake folk  4096 Apr  3  2017 0.0001038898/
dr-xr-x---+ 2 ake folk  4096 Apr  3  2017 0.0001082185/
```

A normal cp -r or rsync -a has no problem with this, it only sets directory permissions AFTER finishing the subtree.
dcp --preserve and dsync both fail with permission denied when doing mkdir in the target tree.

Note that this only happens when running larger jobs on larger trees.

Small trees with just a few dirs have no problem.

akesandgren:
Never mind, looks like ACL problems against our new Lustre server...

akesandgren:
Problem seem to be in mfu_copy_xattrs when called from mfu_create_directory.
The xattrs contain system.posix_acl_default and system.posix_acl_access.
When chmod -w has been done on the src dir, the system.posix_acl_access xattr contains user::r-x resulting in write-protected target.

I suggest filtering out any system.posix_acl xattrs in mfu_copy_xattrs, like in #401

daltonbohning:
I think part of the issue here is that `mfu_copy_xattrs` is called immediately after creating each directory/file, to accommodate Lustre striping params:
https://github.com/daltonbohning/mpifileutils/blob/4ec784108d066abe5060ebb197c1dba83e88bd7d/src/common/mfu_flist_copy.c#L933-L942

In contrast, ownership, permissions, and timestamps are copied _after_ copying the files, and starting from the deepest level, to handle similar issues with standard permissions.

[...]

PR #401, opened 2020-10-03
Fix problems with copying xattrs that contain ACLs causing write-protected dirs

akesandgren:
Fixes issue https://github.com/hpc/mpifileutils/issues/400

Signed-off-by: Åke Sandgren <ake.sandgren@hpc2n.umu.se>

### DIFF
--- a/src/common/mfu_flist_copy.c
+++ b/src/common/mfu_flist_copy.c
@@ -324,14 +324,16 @@ static int mfu_copy_xattrs(
 
             /* set attribute on destination object */
             if(got_val) {
-                errno = 0;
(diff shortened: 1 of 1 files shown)


akesandgren:
It's probably better to not copy xattrs by default as in #418

ofaaland:
@akesandgren it looks from the comments in #418 like the problem motivating this patch was resolved by other patches in tag 0.11.1, and we can close this.  Is that correct?  Thanks

akesandgren:
Yes 0.11.1 does solve the problem, closing

PR #403, opened 2020-10-05
memory leaks and DAOS enhancements

### LINKED ISSUE #356: dsync silently skipping directories that can't be opened (EACCES)
Hello,

For directories that we are denied permission to, dsync does not generate an error (i.e. it's silently skipped).

I think it's as simple as updating walk_readdir_process_dir() and walk_stat_process_dir() to add the appropriate MFU_LOG message to the part that says:

    if (! dirp) {
        /* TODO: print error */
    }

(at least, that works for my use-case)

Thanks

Craig.

### LINKED ISSUE #392: mfu_flist_chunk.c: memory leak
None of these MALLOCs have a corresponding `mfu_free`:

https://github.com/hpc/mpifileutils/blob/66c5273035396c34998c32e9d7918bce9aeca58f/src/common/mfu_flist_chunk.c#L95-L96

https://github.com/hpc/mpifileutils/blob/66c5273035396c34998c32e9d7918bce9aeca58f/src/common/mfu_flist_chunk.c#L127-L131

**NOTE:** A pointer to each `elem` is stored in both `heads` and `tails`
https://github.com/hpc/mpifileutils/blob/66c5273035396c34998c32e9d7918bce9aeca58f/src/common/mfu_flist_chunk.c#L190

https://github.com/hpc/mpifileutils/blob/66c5273035396c34998c32e9d7918bce9aeca58f/src/common/mfu_flist_chunk.c#L252

https://github.com/hpc/mpifileutils/blob/66c5273035396c34998c32e9d7918bce9aeca58f/src/common/mfu_flist_chunk.c#L263-L264

https://github.com/hpc/mpifileutils/blob/66c5273035396c34998c32e9d7918bce9aeca58f/src/common/mfu_flist_chunk.c#L270-L271

https://github.com/hpc/mpifileutils/blob/66c5273035396c34998c32e9d7918bce9aeca58f/src/common/mfu_flist_chunk.c#L300

https://github.com/hpc/mpifileutils/blob/66c5273035396c34998c32e9d7918bce9aeca58f/src/common/mfu_flist_chunk.c#L329

\
\
The only MALLOC in this function that should _**not**_ be freed is:
https://github.com/hpc/mpifileutils/blob/66c5273035396c34998c32e9d7918bce9aeca58f/src/common/mfu_flist_chunk.c#L371
Since the `free` is done in `mfu_file_chunk_list_free`

### LINKED ISSUE #393: mfu_flist_copy.c: memory leak
This `MALLOC` does not have a corresponding `mfu_free`:
https://github.com/hpc/mpifileutils/blob/66c5273035396c34998c32e9d7918bce9aeca58f/src/common/mfu_flist_copy.c#L2042

### LINKED ISSUE #398: daos: mfu_file_open daos discrepancies
Within `mfu_copy_file` -> `mfu_copy_open_file`, errors are not captured properly and logged for `mfu_file->type == DAOS`:
https://github.com/hpc/mpifileutils/blob/871d8f7d96ae01d1bf66c7e9751884f3824ee368/src/common/mfu_flist_copy.c#L1961-L1967
https://github.com/hpc/mpifileutils/blob/871d8f7d96ae01d1bf66c7e9751884f3824ee368/src/common/mfu_flist_copy.c#L1969-L1975

If the type is DAOS, we should check if `mfu_file->obj` is NULL, which would be indication of failure. We also need to make sure the NULL check is sufficient to avoid accidentally using the previously cached file.

Alternatively, we may decide to make `mfu_file_open` return an error, since both `mfu_open` and `dao_open` return errors:
https://github.com/daltonbohning/mpifileutils/blob/871d8f7d96ae01d1bf66c7e9751884f3824ee368/src/common/mfu_io.c#L793
Then, we could also make `mfu_copy_open_file` propogate the error from `mfu_file_open`.
This approach needs further investigation, but either way, DAOS errors should be captured properly and logged.

### LINKED ISSUE #399: daos: daos_symlink and daos_readlink implementation
Within `mfu_io` we need `daos_symlink`, `daos_readlink`, and the corresponding wrapper functions `mfu_file_symlink`, `mfu_file_readlink`.
These can be implemented with `dfs_open` and `dfs_get_symlink_value`.

Then, references to `mfu_readlink` should be updated appropriately.

daltonbohning:
**_daos: added daos_symlink and daos_readlink_**
- Added `mfu_file_readlink`
- Added `daos_readlink`
- Added `mfu_file_symlink`
- Added `daos_symlink`
- Updated corresponding DAOS entry points
- Fixed bug with `PATH_MAX` check

closes #399 

**_mfu_copy_files: fixed memory leak with vals_**
Within `mfu_copy_files`, added `mfu_free(&vals)`

closes #393 

**_mfu_file_chunk_list_alloc: fixed memory leaks_**
Witin mfu_file_chunk_list_alloc:
Added mfu_free for each of the following:
- `sendlist`
- `recvlist`
- `heads`
  - including nested linked list elements
  - does not call `mfu_file_chunk_list_free`, because that also frees the `name`, which should not be freed here, because it points to the `mfu_flist` entry which should still be valid upon exiting
- `tails`
- `counts`
-`bytes`
- `sendbufs`
  - including each index
- `recvranklist`
- `request`
- `status`
- `send_counts`
- `recv_counts`
- `recvbuf`

closes #392 

**_mfu_copy_open_file: uniformly propagate errors_**
- Changed `mfu_copy_open_file` to `return -1` on error.
- Changed callers to check for the error instead of checking if `fd<0`
  - Allows the detail of the error checks to be done in one place
  - Allows easier compatibility with DAOS

closes #398

**_mfu_flist_walk: added more error messages_**
Added a few error messages to places with TODO

closes #356

### DIFF
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -126,8 +126,8 @@ static dfs_obj_t* daos_hash_lookup(const char* name, mfu_file_t* mfu_file)
 
     /* Allocate space for name, up to PATH_MAX,
      * leaving 1 extra for the null terminator */
-    size_t name_len = strnlen(name, PATH_MAX-1);
-    if (name_len >= PATH_MAX-1) {
+    size_t name_len = strnlen(name, PATH_MAX);
+    if (name_len > PATH_MAX-1) {
         MFU_LOG(MFU_LOG_ERR, "name is too long");
         daos_dir_hdl_delete(&hdl);
         return NULL;
@@ -628,6 +628,54 @@ retry:
  * Links
  ****************************/
 
+/* emulates readlink with dfs_lookup, dfs_get_symlink_value */
+ssize_t daos_readlink(const char* path, char* buf, size_t bufsize, mfu_file_t* mfu_file)
+{
+#ifdef DAOS_SUPPORT
+    char* name     = NULL;
+    char* dir_name = NULL;
+    parse_filename(path, &name, &dir_name);
+    assert(dir_name);
+
+    daos_size_t got_size = (daos_size_t) bufsize;
+
+    /* Lookup the parent directory first, since it is likely cached */
+    dfs_obj_t* parent = daos_hash_lookup(dir_name, mfu_file);
+    if (parent == NULL) {
+        errno = ENOENT;
+        got_size = -1;
+    } else { 
+        /* Lookup the symlink within the parent */
+        dfs_obj_t* sym_obj;
+        int rc = dfs_lookup_rel(mfu_file->dfs, parent, name, O_RDWR, &sym_obj, NULL, NULL);
+        if (sym_obj == NULL) {
+            MFU_LOG(MFU_LOG_ERR, "dfs_lookup_rel %s failed", path);
+            errno = ENOENT;
+            got_size = -1;
+        } else {
+            /* Read the symlink value. This also makes sure it is S_IFLNK */
+            rc = dfs_get_symlink_value(sym_obj, buf, &got_size);
+            if (rc) {
+                errno = rc;
+                got_size = -1;
+            } else {
+                /* got_size includes the NULL terminator, but mfu_file_readlink
+                * expects that it does not */
+                got_size--;
+            }
+
+            /* Release the symlink */
+            dfs_release(sym_obj);
+        }
+    }
+
+    return (ssize_t) got_size;
+
+#else
+    return (ssize_t) 0;
+#endif
+}
+
 /* call readlink, retry a few times on EINTR or EIO */
 ssize_t mfu_readlink(const char* path, char* buf, size_t bufsize)
 {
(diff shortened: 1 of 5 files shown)


src/common/mfu_io.c
```diff
@@ -628,6 +628,54 @@ int mfu_remove(const char* path)
...
+    assert(dir_name);
+
+    int rc = 0;
+    daos_size_t got_size = (daos_size_t) bufsize;
+
+    /* Lookup the parent directory first, since it is likely cached */
+    dfs_obj_t* parent = daos_hash_lookup(dir_name, mfu_file);
+    if (parent == NULL) {
+        errno = ENOENT;
+        rc = -1;
```
adammoody, line 647: Looks like the code below is return ``got_size``.  Do you mean to set that to -1 instead?

src/common/mfu_io.c
```diff
@@ -628,6 +628,54 @@ int mfu_remove(const char* path)
...
+    } else { 
+        /* Lookup the symlink within the parent */
+        dfs_obj_t* sym_obj;
+        rc = dfs_lookup_rel(mfu_file->dfs, parent, name, O_RDWR, &sym_obj, NULL, NULL);
+        if (sym_obj == NULL) {
+            MFU_LOG(MFU_LOG_ERR, "dfs_lookup %s failed", path);
+            errno = ENOENT;
```
adammoody, line 654: Do we need to set the return code here?

adammoody:
Thanks, @daltonbohning .  I had a couple questions, but this looks good.  Thanks a lot for finding and fixing those memory leaks!

src/common/mfu_io.c
daltonbohning: Yes, thank you. Gotta love copy-paste!

src/common/mfu_io.c
daltonbohning: Yes, thank you!

daltonbohning:
Thanks @adammoody And you're welcome for the memory leaks!

dsikich (approved):
Thanks!

daltonbohning:
@dsikich Thank you!

PR #405, opened 2020-10-06
dereference: rough draft

daltonbohning:
This is a work in progress for implementing --dereference
@adammoody I would like your input on the way I have approached this.
Basically, the crux of the solution is calling `stat()` instead of `lstat()` in `mfu_flist_walk.c`, and then adding corresponding:
- `mfu_file_stat`
- `mfu_stat`
- `daos_stat`

I have tested this with:
- symlinks to files
- symlinks to directories
- symlinks to symlinks
- symlinks to directories with other symlinks to ...

And so far I don't see any issues with it. The end result is that _**structure**_ of the destination is exactly the same as the _**structure**_ of the source, but the _**symlinks**_ from the source are just replaced with the underlying files/directories. I have also compared this to `cp -r --dereference` and this is how it is done.

My concern is really that this just seems too simple, so I would like any input you might have.

As a side note, this does not work for DAOS, but only because our `dfs` library does not have support for opening symlinks. I'm not sure when that support will be added (in progress), but once this is tried-and-true, I don't mind pushing without full DAOS support for now, since other users could benefit from this.

### DIFF
--- a/src/common/mfu_flist_walk.c
+++ b/src/common/mfu_flist_walk.c
@@ -518,10 +518,17 @@ static void walk_stat_process(CIRCLE_handle* handle)
 
     /* stat item */
     struct stat st;
-    int status = mfu_file_lstat(path, &st, mfu_file);
+    int status;
+    if (true) { /* TODO if --dereference*/
+        /* if symlink, stat the symlink value */
+        status = mfu_file_stat(path, &st, mfu_file);
+    } else {
+        /* if symlink, stat the symlink itself */
+        status = mfu_file_lstat(path, &st, mfu_file);
+    }
     if (status != 0) {
-        /* print error */
-        return;
+        MFU_LOG(MFU_LOG_ERR, "Failed to stat: '%s' (errno=%d %s)",
+                path, errno, strerror(errno));
     }
 
     /* increment our item count */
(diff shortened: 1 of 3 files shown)


dsikich:
@daltonbohning did Mohomad say he was going to add support for symlinks to DFS? I know he mentioned it as a possibility in the meeting on Monday but just want to confirm?

daltonbohning:
@dsikich He is going to add support for symlinks with relative path values (opposed to absolute paths).
But only today did I discover that opening a symlink is not really support at all. For example, `dfs_open` calls `open_symlink`, but `open_symlink` explicitly returns `ENOTSUP` unless it is being created.
And also, POSIX `open` and `opendir` will follow symlinks and open the file/dir pointed to, but `dfs` errors if this is attempted.
I've asked Momomad about:
- `open_file`-  If symlink, follow and open file
- `dfs_readdir` - If symlink, follow and open dir
- `open_symlink` - Replace `ENOTSUP` with an implementation

dsikich:
@daltonbohning I see, ok good you noticed that.

adammoody:
Nice!  Maybe it is just that simple.  That would be great.  Thanks for starting on this.

I'm busy for a couple days, but then I'll look this over.

daltonbohning:
@adammoody No rush. There are still some finer details I need to hammer out, so it's not fully ready yet anyway.

daltonbohning:
I'm realizing I still need quite a bit more work with the DAOS portion, so I'm just going to close this PR to stop the notification spam you're probably receiving.

PR #406, opened 2020-10-09
dcp and dsync: add --dereference and --no-dereference

### LINKED ISSUE #29: dcp: add --dereference option as in cp to copy original files instead of links
Some have requested an option to copy the source file the link points to rather than the link.

### LINKED ISSUE #404: daos: xattr support
**Wrapper `xattr` IO functions:**
- `mfu_file_llistxattr`
- `mfu_file_lgetxattr`
- `mfu_file_lsetxattr`

**Corresponding `daos_*` IO functions:**
- `daos_llistxattr`
- `daos_lgetxattr`
- `daos_lsetxattr`

**Corresponding POSIX IO functions:**
- `mfu_llistxattr`
- `mfu_lgetxattr`
- `mfu_lsetxattr`

**Update raw `xattr` functions**
- `mfu_flist_copy.c`
- Others TBD

**Error messages**
I noticed a few of the error messages for the `xattr` calls in `mfu_flist_copy.c` refer to the wrong function. For example, this message is for a `lgetxattr` call. We might as well correct this in the process.
https://github.com/hpc/mpifileutils/blob/4ec784108d066abe5060ebb197c1dba83e88bd7d/src/common/mfu_flist_copy.c#L309

daltonbohning:
### Support DAOS --preserve
Added new IO functions:
- `llistxattr`
  - `mfu_file_llistxattr`
  - `mfu_llistxattr`
  - `daos_llistxattr`
- `lgetxattr`
  - `mfu_file_lgetxattr`
  - `mfu_lgetxattr`
  - `daos_lgetxattr`
- `lsetxattr`
  - `mfu_file_lsetxattr`
  - `mfu_lsetxattr`
  - `daos_lsetxattr`
- `lchown`
  - `mfu_file_lchown`
  - `mfu_lchown`
  - `daos_lchown`
    - DFS has owner and group permissions only at the container level
      so this is just a placeholder to avoid errors.
- `utimensat`
  - `mfu_file_utimensat`
  - `mfu_utimensat`
  - `daos_utimensat`

#closes #404

### Implement --dereference and --no-dereference
**`mfu_walk_opts`**
- Added _dereference_ option
  - default 0

**`mfu_copy_opts`**
- Added _no-dereference_ option
  - default 0

**`mfu_flist_walk.c` -> `walk_stat_process`**
- Added `DEREFERENCE` global, similar to existing globals
- If `DEREFERENCE` call `stat`
  - Effectively follows symlinks to files/dirs pointed to
- Otherwise call `lstat`
- Added error message for failed `stat`/`lstat`

**`mfu_io`**
- Altered `daos_hash_lookup` to set `errno` on error
  - Adjusted callers to account for this
- Fixed missing `parent` lookup in `daos_access`
- Added `mfu_file_faccessat`, `mfu_faccessat`, `daos_faccessat`
  - Parameters allow for *not* following symlinks
- Renamed `daos_stat` to `daos_lstat` to be more accurate
- Added `mfu_file_stat`, `mfu_stat`, `daos_stat`
  - Similar to `lstat`, but follows symlinks
  - `daos_stat` is a WIP, pending `DFS` library changes

**`mfu_util.h`**
- Explicitly initialize some DAOS structs to silence erroneous
   valgrind errors.

**`mfu_param_path.c` -> `mfu_param_path_check_copy`**
- Added `no_dereference` parameter
- If `no_dereference`, call `faccessat(..., AT_SYMLINK_NOFOLLOW)`
  - Essentially allows broken links to be copied
- Otherwise, call `access`

**`dcp.c`**
- Added `--dereference`
  - "-L"
- Added `--no-dereference`
  - "-l" (short name hidden)

**`dsync.c`**
- Added `--dereference`
  - "-L"
- Added `--no-dereference`
  - "-x" (short name hidden)

**`dwalk.c`**
- Added `--dereference`

#closes #29 


daltonbohning:
**Questions**
1. I added `--dereference` and `--no-dereference` to `dcp` and `dsync`.
    To be similar to `cp`, I used `-L` for `--dereference`.
    But `cp` uses `-P` for `--no-dereference`. While neither `dcp` nor `dsync` show the short `-P` to the user, 
    it is still being used for `--progress`. Do we want to change `--progress` to use a different short option,
    so `--no-dereference` can use `-P` like `cp` does?

2. I found that `--dereference` made most sense to put into `mfu_walk_opts`,
    whereas `--no-dereference` made most sense to put into `mfu_copy_opts`.
    At first I found this strange, but after comparing `dcp` to `cp` in terms of functionality
    for these options, the two utilities are very similar.
    But still, any push back here or other ideas?

3. I didn't thoroughly test `--dereference` with `dsync`, since I'm not 100% familiar with the utility. However, I did notice some strange behavior when running with `--dereference` and then **not** with `--dereference`, and vice versa. For example, if you first run `dsync --dereference s d`, and then run `dsync s d`, any files or directories that were dereferenced in the first run are not overwritten with the link in the second run. And vice versa, links would not be overwritten with the dereferenced counterparts.
Thoughts here?

**Notes**
1. The new `daos_stat` function looks somewhat horrendous, because it kind of is. It works, and covers as many edges cases as I could think of, but it will be tweaked and simplified once `DFS` gets some proper support for traversing symlinks.

2. I tested `dcp` with varying structures. I.e. links to files, links to dirs, links to links, links to directories with links to...
But if you happen to have time (with POSIX at least, not DAOS), please feel free to run some complex `--dereference` test cases

daltonbohning:
I started adding DAOS support for `xattrs` and now I see that I need to add `--dereference` support in place of `mfu_copy_xattrs` -> `lgetxattr`. In the current state, the `xattrs` for the **_link itself_** will be copied to the dereferenced file/dir. This mostly involve switch between `l*xattr` (link interrogation) and `*xattr` (link dereferencing), similar to `lstat` and `stat`

adammoody:
Thanks @daltonbohning .  This all sounds good.

The short option for ``--progress`` is not exposed to the user, so we should be good to change that which would free up ``-P`` to be used for something else.  The progress option is used across most of the tools, so let's try to find a replacement letter that works across them all if possible.

Another thing we can keep in mind for the future.  I can imagine at some point we may end up wanting to allow the user to have finer control with something like the ``--preserve=`` option where one can specify subsets of things to be copied:
>
>
>        --preserve[=ATTR_LIST]
>              preserve the specified attributes (default: mode,ownership,timestamps), if possible  addi-
>              tional attributes: context, links, xattr, all

adammoody:
I like the idea of adding ``--dereference`` to other tools if it could be useful, but I also haven't had time to dive in to see what implications there might be with that.

daltonbohning:
> The short option for `--progress` is not exposed to the user, so we should be good to change that which would free up `-P` to be used for something else. The progress option is used across most of the tools, so let's try to find a replacement letter that works across them all if possible.

Great, I'll do just that!

> Another thing we can keep in mind for the future. I can imagine at some point we may end up wanting to allow the user to have finer control with something like the `--preserve=` option where one can specify subsets of things to be copied:
> 
> ```
> --preserve[=ATTR_LIST]
>       preserve the specified attributes (default: mode,ownership,timestamps), if possible  addi-
>       tional attributes: context, links, xattr, all
> ```

I agree. This could potentially be helpful for DAOS as well. I created a _wishlist_ issue for this, as a reminder: #410. We (DAOS team) will soon be getting more feedback on DAOS-related requirements. If this happens to be requested, I don't mind implementing this.

> I like the idea of adding --dereference to other tools if it could be useful, but I also haven't had time to dive in to see what implications there might be with that.

No worries.  Just something we can keep in mind

daltonbohning:
`'R'` wasn't being used for any options, so I changed all `--progress` to use `R` as the short option. **R** for p**R**ogress

daltonbohning:
@adammoody @dsikich 
I think I _**finally**_ have this in a stable state. At your convenience, please let me know if everything looks good.

dsikich:
@daltonbohning great! I can take a look tomorrow or Monday most likely

src/common/mfu_flist_copy.c
```diff
@@ -379,7 +396,7 @@ static int mfu_copy_ownership(
     gid_t gid = (gid_t) mfu_flist_file_get_gid(flist, idx);
 
     /* note that we use lchown to change ownership of link itself, it path happens to be a link */
-    if(mfu_lchown(dest_path, uid, gid) != 0) {
+    if(mfu_file_lchown(dest_path, uid, gid, mfu_dst_file) != 0) {
```
adammoody, line 399: Do we also need a dereference/no-dereference option for chown?

daltonbohning: I found that we don't have to worry about dereference/no-dereference for the _**destination**_, since the destination is created as a mirror of the source, and when we stat/lstat the source, we save the type (file/dir/link).
For example, if we use `--dereference` on a source symbolic link, `stat` will be called, which will dereference the link and store `struct stat` info for the file/dir (not the link). Then, we create the destination based on this `struct stat` info. Then, when we call `lchown` on the _**destination**_, it is okay to not dereference the link again, since if we originally dereferenced the source link then the destination wouldn't even be a link at all. And if the destination is a link, then we must not have dereferenced links, so `lchown` again is correct.

daltonbohning: But then again... please let me know if there are holes in this that I'm not seeing.

daltonbohning: Your comment got me thinking... And I realized that I didn't account for `dcp --input` (I haven't used that option at all). This function will need to handle dereferencing:
https://github.com/hpc/mpifileutils/blob/4ec784108d066abe5060ebb197c1dba83e88bd7d/src/common/mfu_flist_walk.c#L763-L769

adammoody: > 
> 
> I found that we don't have to worry about dereference/no-dereference for the _**destination**_, since the destination is created as a mirror of the source, and when we stat/lstat the source, we save the type (file/dir/link).
> For example, if we use `--dereference` on a source symbolic link, `stat` will be called, which will dereference the link and store `struct stat` info for the file/dir (not the link). Then, we create the destination based on this `struct stat` info. Then, when we call `lchown` on the _**destination**_, it is okay to not dereference the link again, since if we originally dereferenced the source link then the destination wouldn't even be a link at all. And if the destination is a link, then we must not have dereferenced links, so `lchown` again is correct.

Yes, thanks for the explanation.  I also just realized that under closer inspection.  You're right in that the destination should be the right type already, so it's fine to always use the "l" variant of the calls for destination items.  It only makes a difference on source items.

adammoody: Ah yes, good catch on ``mfu_flist_stat``.

Not all tools calls ``flist_stat`` after reading an input file.  If someone reads an input file as we do not call ``flist_stat``, then we may end up with a source list that does still have symlinks, e.g., if they ran dwalk without dereference and saved that to a file with ``--output`` which they then read back in to another tool.  We'll want to work through that case, though nothing jumps to mind right away.

adammoody:
This generally looks good to me.  Really nice work, @daltonbohning .

I'll look at the dsync case you mentioned above.

Would you mind also documenting these new options in the rst files for the tools, like?

https://github.com/hpc/mpifileutils/blob/master/doc/rst/dcp.1.rst

We use those rst files to generate man pages, and they get linked from our readthedocs.  We try to keep the same ordering of options that we use in the ``print_usage`` function, but we can list more detail in the description of the rst file.

I feel like it would help end users if we distinguish between ``-L``, ``-P``, and using neither a bit more.

adammoody:
Can you describe an example set up where you found that dsync was not overwriting items?

adammoody:
> 
> 
> Can you describe an example set up where you found that dsync was not overwriting items?

Oh, wait.  I think I found a case.  I have a directory that contains a regular file "foo" and a link to that same file "bar":
```
dir1/
  for
  bar -> foo
```
I found that if I run ``dsync --dereference dir1 dir2``, and ``dir2`` does not exist, then I get the expected behavior in that it make a copy of dir1 but it replaces bar with a physical copy of foo:
```
dir2/
  for
  bar (copy of foo)
```
If I then run ``dsync dir1/ dir2/``, while dir2 still exists, it replaces bar with a symlink to foo, which is also expected:
```
dir2/
  for
  bar -> foo
```
However, if I try ``dsync --dereference dir1 dir2``, I'd expect this to replace the bar symlink with a physical copy of foo again, but it does not.  I'm guessing that's because, it's also dereferencing bar in the destination, so it sees that bar as a file rather than a symlink.

daltonbohning:
After more testing, I'm finding that this isn't necessarily related to `--dereference`.
I have this directory structure:
```
dir1/
    file
dir2/
    file -> something
```
And these tests:
```
dcp dir1/file dst
ls -l
-rw-rw-r-- 1 dbohninx dbohninx    6 Oct 19 15:25 dst
dcp dir2/file dst
...
Original link exists, skip the creation:
...
ls -l
-rw-rw-r-- 1 dbohninx dbohninx    6 Oct 19 15:25 dst
```
This is because when creating the file, we ignore `EEXIST`:
https://github.com/daltonbohning/mpifileutils/blob/1e48169d87d8f63b03ada7897d9a12f533ae05d9/src/common/mfu_flist_copy.c#L1129-L1143
And we do the same for directories and files. For example, continuing with the test above, `dcp dir1 dst` doesn't result in `dst` becoming a directory, and the copy into `dst` fails, because it already exists as a file.

More generally, I think the copy operation needs to handle when the destination exists, but it's not the same type as the source.
Thoughts?

daltonbohning:
> Would you mind also documenting these new options in the rst files for the tools, like?
> 
> https://github.com/hpc/mpifileutils/blob/master/doc/rst/dcp.1.rst
> 
> We use those rst files to generate man pages, and they get linked from our readthedocs. We try to keep the same ordering of options that we use in the `print_usage` function, but we can list more detail in the description of the rst file.
> 
> I feel like it would help end users if we distinguish between `-L`, `-P`, and using neither a bit more.

I added some documentation for this. I added a little more detail that `cp` has, but it is by no means a complete account. FOr some of the specifics, I actually just had to run `cp` and do my own testing. For example, something I came across just this morning was that `cp` doesn't seem to explicitly handle when `--dereference` and `--no-dereference` are used in conjunction. It seems that whichever is specified _**last**_ overwrites the other. I made `dcp` and `dsync` behave similarly, but how much detail do you think should go in the documentation about specifics such as this?

src/common/mfu_flist_copy.c
daltonbohning: I can't get `--input` to work, due to some other issues.
- https://github.com/daltonbohning/mpifileutils/blob/c409987e1a572bc3e64337add489980942f8fe58/src/dcp/dcp.c#L933-L941
  - I introduced this, without realizing that `--input` allows input from a file. This is easy to fix, and I will
    - `if ((argc-optind) < 2 && !inputname)`

- https://github.com/daltonbohning/mpifileutils/blob/c409987e1a572bc3e64337add489980942f8fe58/src/dcp/dcp.c#L1000-L1011
  - This essentially requires both a `src` and `dst` to be provided, even when using `--input`. I don't believe I originally introduced this, so it seems it must have already been non-functioning. (not sure about pre-DAOS)
  - This is easy to fix, and I will: `if (!valid and !inputname)`

- https://github.com/daltonbohning/mpifileutils/blob/c409987e1a572bc3e64337add489980942f8fe58/src/common/mfu_flist_walk.c#L766
  - We are trying to reference `CURRENT_PFILE`, but that might global not always be initialized.
  - I'll fix this as well.

In short... I think some of our DAOS changes broke the `--inputname` functionality, so I will correct that.
And then I will get `--inputname` to work correctly with `--dereference`

daltonbohning: Okay, I corrected that issue and also adjusted `mfu_flist_stat` to accept `dereference` parameter, and also try to automatically dereference symbolic links that were dereferenced in `dwalk`'s output. The idea is that if we come across something that _**is not**_ a link, then we go ahead and try to dereference that, since `dwalk` might output `S_ISREG` for a symlink path, which means that the symlink should actually be dereferenced.

I will need to look at other places the `dwalk`'s output is used as input, and make sure these places also correctly handle when the path was dereferenced by `dwalk`.

daltonbohning: These are places that might need consideration to handle `dwalk --dereference` input.
- https://github.com/daltonbohning/mpifileutils/blob/4f53ca409712a2ba898dd14cd9f97d6c46dbcdde/src/common/mfu_flist_chmod.c#L1010
  - I think `chmod` doesn't matter, since `chmod` on a link dereferences the link anyway.
- https://github.com/daltonbohning/mpifileutils/blob/copy-dereference/src/common/mfu_flist_remove.c
  - Maybe somewhere in here, but this is more complex
  - `dsh`, `dsync`, and `drm` call `mfu_flist_unlink`
  - Maybe we just don't want to allow deleting in this way, since it could be dangerous to dereference and delete something accidentally.
- https://github.com/daltonbohning/mpifileutils/blob/copy-dereference/src/dreln/dreln.c
  - I think `dreln` is actually okay, since it manually filters out only links. If you pass it an input that has dereferenced links, then it just won't do anything at all.
- https://github.com/daltonbohning/mpifileutils/blob/4f53ca409712a2ba898dd14cd9f97d6c46dbcdde/src/common/mfu_param_path.c#L632
  - `mfu_param_path_set` / `mfu_param_path_set_all`
  - `dfind`, among others, call this to get `stat` info, among other things. This one seems a little complex to get working entirely, since the `stat` info is not always used entirely. For example, even `dcp` calls `mfu_param_path_set_all` on the source paths, but then `mfu_param_path_check_copy` still behaves as expected.
  - I'm not sure of the other implications of this for `dfind`

If this is too much to get into at this time, we could either leave it as is, which means `--input` might not always be dereferenced (at least not until those tools get explicit `--dereference` support), or I could back-out `--dereference` from `dwalk` until we are better prepared. I've really only working with `dcp`, so I'm not knowledgeable about the other tools.

adammoody:
> 
> 
> 
> More generally, I think the copy operation needs to handle when the destination exists, but it's not the same type as the source.
> Thoughts?

Yes, dcp is not very robust in dealing with items that already exist.  I think we had it ignore EEXIST errors so that someone could attempt a second dcp copy of a large directory if their first attempt had failed part way through.  However, that was all before dsync existed.  Really, dsync is better for these kinds of copies (and recopies).  For dcp, we probably could now have it return an error on EEXIST or look to see if it makes sense to emulate the behavior of cp if we can.

doc/rst/dwalk.1.rst
```diff
@@ -70,6 +70,11 @@ OPTIONS
 
    Print files to the screen.
 
+.. option:: -L, --dereference
+
+   Dereference symbolic links and copy the target file or directory
```
adammoody, line 75:
```suggestion
   Dereference symbolic links and walk the target file or directory
```

adammoody:
> 
> 
> > Would you mind also documenting these new options in the rst files for the tools, like?
> > https://github.com/hpc/mpifileutils/blob/master/doc/rst/dcp.1.rst
> > We use those rst files to generate man pages, and they get linked from our readthedocs. We try to keep the same ordering of options that we use in the `print_usage` function, but we can list more detail in the description of the rst file.
> > I feel like it would help end users if we distinguish between `-L`, `-P`, and using neither a bit more.
> 
> I added some documentation for this. I added a little more detail that `cp` has, but it is by no means a complete account. For some of the specifics, I actually just had to run `cp` and do my own testing. For example, something I came across just this morning was that `cp` doesn't seem to explicitly handle when `--dereference` and `--no-dereference` are used in conjunction. It seems that whichever is specified _**last**_ overwrites the other. I made `dcp` and `dsync` behave similarly, but how much detail do you think should go in the documentation about specifics such as this?

Thanks for digging into cp to figure that out.  The amount of detail you have now is good.

It looks to me like setting ``no_dereference`` currently affects whether ``param_path`` will deference a symlink during the access call.  Those would typically be paths the user listed on the command line.  Are there any other places?

adammoody:
Browsing at the rsync man page, it has these related options:
```
        -l, --links                 copy symlinks as symlinks
        -L, --copy-links            transform symlink into referent file/dir
            --copy-unsafe-links     only "unsafe" symlinks are transformed
            --safe-links            ignore symlinks that point outside the tree
            --munge-links           munge symlinks to make them safer
        -k, --copy-dirlinks         transform symlink to dir into referent dir
        -K, --keep-dirlinks         treat symlinked dir on receiver as dir
```
Interesting they have a way to change behavior distinguishing between links that point within the tree and those pointing to something external to the tree.  I could see that coming up for our users, too.  Anyway, let's not complicate this current PR with that added complexity.

daltonbohning:
> Thanks for digging into cp to figure that out. The amount of detail you have now is good.
> 
The source code for `cp` actually uses multiple flags for its `dereference` option:
https://github.com/coreutils/coreutils/blob/master/src/cp.c
(Search for `dereference`)
And then, digging into [copy.c](https://github.com/coreutils/coreutils/blob/master/src/copy.c), it gets more complex.
I wonder if maybe instead of the underlying `dereference` and `no-dereference` flags to copy_opts and walk_opts, we should use a single flag with values like: DEREF_NEVER, DEREF_ALWAYS, etc. But that would be a little more complex.


> It looks to me like setting `no_dereference` currently affects whether `param_path` will deference a symlink during the access call. Those would typically be paths that the user listed on the command line. Are there any other places?

Correct. This is so far the only place I found that needed this. `cp`, by default, does this:
```
if (x.dereference == DEREF_UNDEFINED)
{
  if (x.recursive && ! x.hard_link)
    x.dereference = DEREF_NEVER;
  else
    x.dereference = DEREF_ALWAYS;
}
```
Whereas `dcp`, by default, essentially never dereferenced symlinks at all before adding `--dereference`. The `l*` version of calls were always being made, so it's a little different from `cp`.
And, `dcp` doesn't have a `-r` flag; the behavior is that `dcp` always uses `-r`.


> I am wondering if we might change behavior so that `no_dereference` is just always on by default, and then only provide a `dereference` option to the user. However, I'm still not entirely sure why the cp command itself provides both of those as options. It seems like it must have at least three different behaviors (default, dereference, no-dereference).

Correct, as I explained above. Initially, I added `--no-dereference` to be synonymous with `cp`, but `dcp` is slightly different anyway, so maybe we don't need it. But another (small) reason is that currently there is not a way to copy "broken" symlinks that are specified at the command line. If a broken link appears in a directory, `dcp` will copy that link just fine. But if you try to do:
`dcp broken_link dst`, then `dcp` tries to call `access(broken_link)`, which fails. For DAOS in particular, this is useful because copying between different filesystems means that links aren't always valid in a different filesystem.


Overall, `cp` has a somewhat complex handling of dereferencing, and this is made more complex in that `dcp` has some slightly different defaults. Any and all criticism is welcome here, so thank you for the feedback!

daltonbohning:
> ... For dcp, we probably could now have it return an error on EEXIST or look to see if it makes sense to emulate the behavior of cp if we can.
> 
> We can change this particular behavior of dcp in a future PR, since that's separate from symlink dereferencing.

I'll create an issue for this to keep track of it.


> I think the dsync change is related to the new dereference option, though. It may be that we can just disable any dereference step when walking the destination (only dereference on source), but I haven't tested that. If you find other strange behavior from dsync, let me know in case we uncover other details to consider.

Confirmed, this is only an issue when using `--dereference`. I'll look into this.

dsikich:
@daltonbohning this looks like there have been a lot of updates recently so I was waiting to look over it again. Is it in a final state now?

daltonbohning:
> @daltonbohning this looks like there have been a lot of updates recently so I was waiting to look over it again. Is it in a final state now?

I think the DAOS-specific changes are "final", such as the new I/O functions, but the more general behavior of `--dereference` and `--no-dereference` aren't 100% yet, pending a few discussion points above.

dsikich:
@daltonbohning yeah, those DAOS I/O functions look good. I think this PR has become a bit big so I don't know if you want to separate those into another PR, and have those open questions stay pending in this PR? I'm fine with either way, but depends on if you think it will help to have some of it merged sooner.

daltonbohning:
In hindsight, I probably should have made multiple PRs. What happened is that when I added `--dereference` to the `xattr` copying, I realized we didn't yet have DAOS `xattr` functions, so I added those as well so I could fully make sure `--dereference` was working as expected. I'm okay with just a single PR, unless @adammoody thinks we should split it up. (The first two commits could be one PR and everything after could be another)

adammoody:
It's fine by me if you'd like to split this into multiple PRs.  I'm also good with one, so it's up to you.

adammoody:
I'll also take another look at things later this week.

daltonbohning:
> It's fine by me if you'd like to split this into multiple PRs. I'm also good with one, so it's up to you.

I'll keep it as one PR for now, unless we find that the `--dereference` portion still needs a lot more work.

> I'll also take another look at things later this week.

Sure, I understand.

adammoody:
Thanks, @daltonbohning .  I haven't had a chance to look at this in full detail yet, but it's solid enough that we could merge it and fix up anything that we find later.

If we decide to change ``dcp`` to work in ``--no-deferenece`` mode by default, then we could drop that option to simplify the user interface.  We'll try to make that call before we stamp the next release.

daltonbohning:
> Thanks, @daltonbohning . I haven't had a chance to look at this in full detail yet, but it's solid enough that we could merge it and fix up anything that we find later.

I think really the biggest outstanding issue is that `--input` doesn't always work with output from `dwalk --dereference`.
(Sometimes a dereferenced symlink is treated as a symlink)
For reference, this is that thread (lost in the noise):
https://github.com/hpc/mpifileutils/pull/406#discussion_r508098547

The easiest way to avoid this issue for now would be to just turn off `--dereference` in `dwalk`.
Or we can just let it be for now.

daltonbohning:
@adammoody I know you confirmed merging this as-is but I want to make sure this outstanding issue is addressed (or at least noted) beforehand.
There are two directions:
1. Keep `--dereference` option in `dwalk`
    - In other tools, this means that sometimes `--input` won't always use the dereferenced paths
2. Remove `--dereference` option in `dwalk`
    - `--input` works as usual

daltonbohning:
Sorry, I was caught up in other things and forgot to merge this.

PR #408, opened 2020-10-15
add --preserve flag to dsync to allow permissions to be copied

sjtbham:
dsync is missing the -p flag, this means that you cannot use it to preserve file permissions.

This patch adds the -p flag to copy_flags similar to dcp

### DIFF
--- a/src/dsync/dsync.c
+++ b/src/dsync/dsync.c
@@ -2947,7 +2947,6 @@ int main(int argc, char **argv)
             mfu_progress_timeout = atoi(optarg);
             break;
         case 'p':
-            case 'p':
             copy_opts->preserve = true;
             if(rank == 0) {
                 MFU_LOG(MFU_LOG_INFO, "Preserving file attributes.");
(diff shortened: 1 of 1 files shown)


src/dsync/dsync.c
```diff
@@ -2945,6 +2946,13 @@ int main(int argc, char **argv)
         case 'P':
             mfu_progress_timeout = atoi(optarg);
             break;
+        case 'p':
+            case 'p':
+            copy_opts->preserve = true;
```
daltonbohning, line 2951:
```suggestion
        case 'p':
            copy_opts->preserve = true;
```

sjtbham: stupid copy paste mistake, resolved :-)

daltonbohning:
The Travis CI build will fail unless you have a signoff message in your final commit: `git commit -s` or `git commit --amend -s`

adammoody:
Thanks @sjtbham .  I'm a bit confused though, because ``dsync`` should be set up to *always* copy permission bits:

https://github.com/hpc/mpifileutils/blob/master/src/dsync/dsync.c#L2843-L2844

Are you finding that it's not copying permissions for you in some case?

sjtbham:
Yes ... but I think it might be something else going on now, not this ... I will close this.

PR #413, opened 2020-10-24
daos: gracefully handle failed fiemap

daltonbohning:
**mfu_io**
- Added `mfu_file_truncate`
- Added `daos_truncate`

**mfu_flist_copy.c**
- `mfu_copy_file_fiemap`
  - Skip for DAOS src
    - Not yet supported
  - Skip if `ioctl` fails with `ENOTSUP`
    - DAOS DFUSE-mounted path does not support this
    - _**it seems that `cp` also silently ignores this**_
  - Change `goto fail` to `goto fail_fiemap` for clarity
  - Fixed some memory leaks for error conditions
    - `goto fail_fiemap` instead of `goto fail_normal_copy`
- Changed an occurrence of `mfu_truncate` to `mfu_file_truncate`

### DIFF
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -1360,6 +1360,21 @@ ssize_t daos_pwrite(const char* file, const void* buf, size_t size, off_t offset
 }
 
 /* truncate a file */
+int mfu_file_truncate(const char* file, off_t length, mfu_file_t* mfu_file)
+{
+    if (mfu_file->type == POSIX) {
+        int rc = mfu_truncate(file, length);
+        return rc;
+    } else if (mfu_file->type == DAOS) {
+        int rc = daos_truncate(file, length, mfu_file);
+        return rc;
+    } else {
+        MFU_ABORT(-1, "File type not known, type=%d",
+                  mfu_file->type);
+    }
+}
+
+/* truncate a file */
 int mfu_truncate(const char* file, off_t length)
 {
     int rc;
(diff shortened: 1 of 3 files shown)


dsikich:
@daltonbohning in your note above it says DAOs does not support the --sparse file option that uses the fiemap operation, and I'm not totally sure by what you mean when it says that a DAOS DFUSE mounted path also does not support this? Do you mean even if you have not built with "DAOS SUPPORT", have a dfuse mounted path, and run a --sparse option when attempting to copy? I want to make sure I have that right.

daltonbohning:
@dsikich 
Neither the DFS library nor the DFUSE library have adequate calls for `--sparse` support. I believe the DAOS core library has recently gained some support for sparse files, but this has not yet been propagated to DFS or DFUSE.

dsikich:
@daltonbohning ok, yeah I didn't think there was support yet. It is possible we may have to support sparse files in the future though.

daltonbohning:
@dsikich Yes, I think so in the future. The intent here was to prevent `dcp` from throwing errors in the meantime. Basically, if the functions are not supported for the `source`, the sparse file is just copied in its entirety. Luckily, the destination doesn't necessarily need to support this, because _**creating**_ a sparse file essentially just means writing to only some parts of the file and letting the filesystem handle how to index that.

dsikich (approved):
@daltonbohning I think this looks good until DAOS can support sparse files.

src/common/mfu_io.c
```diff
@@ -1380,7 +1395,63 @@ int mfu_truncate(const char* file, off_t length)
...
+                errno = rc;
+                rc = -1;
+            }
+
+            /* close the obj */
+            int tmp_rc = dfs_release(mfu_file->obj);
+            if (!rc && tmp_rc) {
+                MFU_LOG(MFU_LOG_ERR, "dfs_release %s failed (%d %s)",
+                        file, rc, strerror(rc));
+                errno = rc;
```
adammoody, line 1440: Do you want to use tmp_rc here?

daltonbohning: Yes... It seems I will never learn my lesson with copy-pasting

adammoody (approved):
Thanks, @daltonbohning .

daltonbohning:
Thanks @adammoody !

PR #418, opened 2020-11-03
Separate xattrs from general preserve attributes (only intended as an example)

akesandgren:
This is a quick mockup for separating xattrs from normal preserve, in rsync preserving xattrs is off by default and only enabled if explicitly requested. dsync should behave the same (and should preferably behave as rsync regarding ACLs too)

I know the short 'X' option clashes with dcp option daos-prefix, but I left it as 'X' since that is what rsync uses for this.

### DIFF
--- a/src/dcp/dcp.c
+++ b/src/dcp/dcp.c
@@ -633,6 +633,7 @@ void print_usage(void)
     printf("  -s, --direct             - open files with O_DIRECT\n");
     printf("  -S, --sparse             - create sparse files when possible\n");
     printf("      --progress <N>       - print progress every N seconds\n");
+    printf("  -X, --xattrs             - preserve extended attributes\n");
     printf("  -v, --verbose            - verbose output\n");
     printf("  -q, --quiet              - quiet output\n");
     printf("  -h, --help               - print usage\n");
@@ -698,6 +699,7 @@ int main(int argc, char** argv)
         {"direct"               , no_argument      , 0, 's'},
         {"sparse"               , no_argument      , 0, 'S'},
         {"progress"             , required_argument, 0, 'P'},
+        {"xattrs"               , no_argument      , 0, 'X'},
         {"verbose"              , no_argument      , 0, 'v'},
         {"quiet"                , no_argument      , 0, 'q'},
         {"help"                 , no_argument      , 0, 'h'},
@@ -867,6 +869,12 @@ int main(int argc, char** argv)
                 break;
             case 'P':
                 mfu_progress_timeout = atoi(optarg);
+                break;
+            case 'X':
+                mfu_copy_opts->preserve_xattrs = true;
+                if(rank == 0) {
+                    MFU_LOG(MFU_LOG_INFO, "Preserving extended attributes.");
+                }
                 break;
             case 'v':
                 mfu_debug_level = MFU_LOG_VERBOSE;
--- a/src/dsync/dsync.c
+++ b/src/dsync/dsync.c
@@ -2842,7 +2842,6 @@ int main(int argc, char **argv)
 
     /* By default, sync option will preserve most attributes. */
     copy_opts->preserve = true;
-    copy_opts->preserve_xattrs = true;
 
     /* flag to check for sync option */
     copy_opts->do_sync = 1;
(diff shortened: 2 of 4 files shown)


src/dcp/dcp.c
```diff
@@ -868,6 +870,12 @@ int main(int argc, char** argv)
             case 'P':
                 mfu_progress_timeout = atoi(optarg);
                 break;
+            case 'X':
+                mfu_copy_opts->preserve_xattrs = true;
+                if(rank == 0) {
+                    MFU_LOG(MFU_LOG_INFO, "Preserving extended attributes.");
+                }
+                break;
```
daltonbohning, line 878: Since the short option `X` for `--daos-prefix` isn't user-facing, it's okay to reuse `X` for something else. But I do think we need to change `--daos-prefix` to use _**some**_ other character (it doesn't really matter). Otherwise, with DAOS support enabled these options will clash.

src/dsync/dsync.c
```diff
@@ -2840,8 +2840,9 @@ int main(int argc, char **argv)
     /* verbose by default */
     mfu_debug_level = MFU_LOG_VERBOSE;
 
-    /* By default, sync option will preserve all attributes. */
+    /* By default, sync option will preserve most attributes. */
     copy_opts->preserve = true;
+    copy_opts->preserve_xattrs = true;
```
daltonbohning, line 2845:
```suggestion
    copy_opts->preserve_xattrs = false;
```
Do you want this _**off**_ by default?

src/dsync/dsync.c
```diff
@@ -2945,6 +2947,12 @@ int main(int argc, char **argv)
         case 'P':
             mfu_progress_timeout = atoi(optarg);
             break;
+        case 'X':
+            copy_opts->preserve_xattrs = true;
+            if(rank == 0) {
+                MFU_LOG(MFU_LOG_INFO, "Preserving extended attributes.");
+            }
+            break;
```
daltonbohning, line 2955: Same as for `dcp`; it's okay to repurpose `X` here, but Ithink we need to use _**some**_ other option for `--daos-prefix`

daltonbohning:
I realize this is a mockup, but I left comments anyway, in case this becomes _**real**_

src/dsync/dsync.c
akesandgren: Yes I do want if off by default, that's thw whole point of this PR.
Copying xattrs by default messes up to many things.

And since rsync -a (which is the most commonly used flag) does not copy xattrs (nor ACLs btw) so should dsync.
I think dsync should mimic rsync as much as possible since that is the "gold standard" at the moment.

My use case for this is migrating a Lustre file system without PFL to a new shiny system with PFL and I do not want  the layout to be copied over.

Frankly I'm a bit surprised to see that dsync did not try to mimic rsync in the first place when it comes to options and behavior.

daltonbohning: Yes, I understand the use case, but it looks like `copy_opts->preserve_xattrs = true;` turns this _**on**_ by default, not _**off**_.
From what I've gathered from conversations with Adam Moody, it generally seems desirable to mimic the serial versions of the tools, but considering simplicity and development effort, this just hasn't yet been done in all places.

akesandgren: oops, now I see what you mean, that's a line that shouldn't be there....
I made a bobo when moving the patch onto master, my original is based in 0.10.1...
I'll drop that.

akesandgren: fixed

ofaaland:
@akesandgren check out this commit which might do what you need:
* 96a9fe7 dsync and dcp: allow user to control copying of xattrs

akesandgren:
> @akesandgren check out this commit which might do what you need:
> 
> * [96a9fe7](https://github.com/hpc/mpifileutils/commit/96a9fe7d2f49a9d6b3e28941bd51f33bef8af8ce) dsync and dcp: allow user to control copying of xattrs

Yeah, that looks good, will try when I need it again.

akesandgren:
Closing this since 0.11.1 seem to do what I had in mind

PR #419, opened 2020-11-03
Fix compile error with GPFS_SUPPORT

Phhere:
There is a syntax error when GPFS_SUPPORT is enabled
``` c
if(type != MFU_TYPE_LINK) {
```
was not closed correctly before endif


Phhere:
Commits are now signed

adammoody:
Thanks @Phhere , this fix was included in https://github.com/hpc/mpifileutils/pull/430

PR #423, opened 2020-11-09
Print name of path that is deemed too long.

akesandgren:
Makes it easier to see where the problem is...

### DIFF
--- a/src/common/mfu_flist_walk.c
+++ b/src/common/mfu_flist_walk.c
@@ -488,7 +488,7 @@ static void walk_stat_process_dir(char* dir, CIRCLE_handle* handle)
                 }
                 else {
                     /* name is too long */
-                    MFU_LOG(MFU_LOG_ERR, "Path name is too long: %lu chars exceeds limit %lu", len, sizeof(newpath));
+                    MFU_LOG(MFU_LOG_ERR, "Path name is too long: %lu chars exceeds limit %lu, (%s/%s)", len, sizeof(newpath), dir, name);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


daltonbohning (commented):
Within this file, there are three total instances of the error `Path name is too long`.
Maybe we should update the other 2 instances to use this new format as well?

akesandgren:
Probably, I picked the one that triggered on my run and didn't see any others at the moment.

src/common/mfu_flist_walk.c
```diff
@@ -488,7 +488,7 @@ static void walk_stat_process_dir(char* dir, CIRCLE_handle* handle)
                 }
                 else {
                     /* name is too long */
-                    MFU_LOG(MFU_LOG_ERR, "Path name is too long: %lu chars exceeds limit %lu", len, sizeof(newpath));
+                    MFU_LOG(MFU_LOG_ERR, "Path name is too long: %lu chars exceeds limit %lu, (%s/%s)", len, sizeof(newpath), dir, name);
```
adammoody, line 491: This is good, but let's tweak the format just a bit.  I try to use a pseudo-convention to always contain printed path names in single quotes set off by a colon.
```suggestion
                    MFU_LOG(MFU_LOG_ERR, "Path name is too long, %lu chars exceeds limit %lu: '%s/%s'", len, sizeof(newpath), dir, name);
```

adammoody:
@akesandgren , would you mind also picking up the other instances in this one commit?

src/common/mfu_flist_walk.c
akesandgren: @adammoody  done.

adammoody:
Thanks @akesandgren

PR #424, opened 2020-11-11
update libmfu.so to version 1.0.0

daltonbohning:
The major version of libmfu.so needs to be updated
since there are new parameters to these functions:
- `mfu_flist.h` -> `mfu_flist_stat`
- `mfu_param_path.h` -> `mfu_param_path_check_copy`

The new parameters are not backwards-compatible;
hence, the major version is updated.


daltonbohning:
Closing, since #427 will take care of this

PR #428, opened 2020-12-03
daos: add support for copying daos containers at obj level

dsikich:
This adds support for copying any type of daos container
in parallel.

    * add obj_id_lo and obj_id_hi to elem_t struct
    * add total_oids field to flist
    * add a pack and unpack call for obj_id_lo and obj_id_hi
    * create mfu_flist_global_oid_size to return total oids
    * add boolean is_posix_copy for non-posix copies
    * adds daos_obj_copy function and helper functions

This copy starts by retrieving all obj ids on rank 0,
then it uses the mfu_flist_spread function to distribute
all obj ids evenly across ranks. After the obj ids are
distributed, the obj ids are copied using the local
flist.

Also, daos will split the anchor used to retrive all
obj ids and enumerate the dkeys, akeys, and records.
There is work that still needs to be done from the daos
level to split the anchor for the object iterator table
used to retrieve obj ids. In the case that a large
record extent needs to be copied, daos already chunks
the record extents using their index ranges. Each record
is of a fixed size, where multiple records is a record
extent.

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -28,7 +28,8 @@ static int daos_check_args(
     int rank,
     char** argpaths,
     daos_args_t* da,
-    int* flag_daos_args)
+    int* flag_daos_args,
+    bool* is_posix_copy)
 {
     char* src_path = argpaths[0];
     char* dst_path = argpaths[1];
@@ -40,6 +41,12 @@ static int daos_check_args(
     bool have_dst_pool  = daos_uuid_valid(da->dst_pool_uuid);
     bool have_dst_cont  = daos_uuid_valid(da->dst_cont_uuid);
     bool have_prefix    = da->dfs_prefix != NULL;
+
+    /* var to determine if this is a non-dfs copy */
+    if (!have_src_path && (have_src_pool && have_dst_pool) &&
+  	(have_src_cont && have_dst_cont)) {
+        *is_posix_copy = false;
+    }
 
     /* Determine whether any DAOS arguments are supplied. 
      * If not, then there is nothing to check. */
@@ -468,11 +475,12 @@ void daos_args_delete(daos_args_t** pda)
 }
 
 int daos_setup(
-  int rank,
-  char** argpaths,
-  daos_args_t* da,
-  mfu_file_t* mfu_src_file,
-  mfu_file_t* mfu_dst_file)
+    int rank,
+    char** argpaths,
+    daos_args_t* da,
+    mfu_file_t* mfu_src_file,
+    mfu_file_t* mfu_dst_file,
+    bool* is_posix_copy)
 {
     int tmp_rc;
 
@@ -484,7 +492,7 @@ int daos_setup(
 
     /* Make sure we have the required DAOS arguments (if any).
      * Safe to return here, since all processes have the same values. */
-    tmp_rc = daos_check_args(rank, argpaths, da, &flag_daos_args);
+    tmp_rc = daos_check_args(rank, argpaths, da, &flag_daos_args, is_posix_copy);
     if (tmp_rc != 0) {
         if (rank == 0) {
             MFU_LOG(MFU_LOG_ERR, "Invalid DAOS args: "
@@ -517,7 +525,7 @@ int daos_setup(
 
     /* Re-check the required DAOS arguments (if any) */
     if (!local_daos_error) {
-        tmp_rc = daos_check_args(rank, argpaths, da, &flag_daos_args);
+        tmp_rc = daos_check_args(rank, argpaths, da, &flag_daos_args, is_posix_copy);
         if (tmp_rc != 0) {
             MFU_LOG(MFU_LOG_ERR, "Invalid DAOS args: "
                     MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS_INVAL_ARG));
@@ -571,7 +579,7 @@ int daos_setup(
         }
     }
 
-    if (!local_daos_error && mfu_src_file->type == DAOS) {
+    if (!local_daos_error && mfu_src_file->type == DAOS && *is_posix_copy) {
         /* DFS is mounted for the source container */
         tmp_rc = daos_mount(mfu_src_file, &da->src_poh, &da->src_coh);
         if (tmp_rc != 0) {
@@ -579,7 +587,7 @@ int daos_setup(
         }
     }
 
-    if (!local_daos_error && mfu_dst_file->type == DAOS) {
+    if (!local_daos_error && mfu_dst_file->type == DAOS && *is_posix_copy) {
         /* DFS is mounted for the destination container */
         if (same_pool) {
             tmp_rc = daos_mount(mfu_dst_file, &da->src_poh, &da->dst_coh);
@@ -602,9 +610,10 @@ int daos_setup(
 }
 
 int daos_cleanup(
-  daos_args_t* da,
-  mfu_file_t* mfu_src_file,
-  mfu_file_t* mfu_dst_file)
+    daos_args_t* da,
+    mfu_file_t* mfu_src_file,
+    mfu_file_t* mfu_dst_file,
+    bool* is_posix_copy)
 {
     /* assume success until failure */
     int rc = 0;
@@ -615,13 +624,14 @@ int daos_cleanup(
             && uuid_compare(da->src_pool_uuid, da->dst_pool_uuid) == 0) {
         same_pool = true;
     }
-
     if (mfu_src_file->type == DAOS) {
-        tmp_rc = daos_umount(mfu_src_file);
-        MPI_Barrier(MPI_COMM_WORLD);
-        if (tmp_rc != 0) {
-            rc = 1;
-        }
+        if (*is_posix_copy) {
+            tmp_rc = daos_umount(mfu_src_file);
+            MPI_Barrier(MPI_COMM_WORLD);
+            if (tmp_rc != 0) {
+                rc = 1;
+            }
+	}
 
         /* Close the container */
         tmp_rc = daos_cont_close(da->src_coh, NULL);
@@ -631,13 +641,14 @@ int daos_cleanup(
             rc = 1;
         }
     }
-
     if (mfu_dst_file->type == DAOS) {
-        tmp_rc = daos_umount(mfu_dst_file);
-        MPI_Barrier(MPI_COMM_WORLD);
-        if (tmp_rc != 0) {
-            rc = 1;
-        }
+        if (*is_posix_copy) {
+            tmp_rc = daos_umount(mfu_dst_file);
+            MPI_Barrier(MPI_COMM_WORLD);
+            if (tmp_rc != 0) {
+                rc = 1;
+            }
+	}
 
         /* Close the container */
         tmp_rc = daos_cont_close(da->dst_coh, NULL);
@@ -674,6 +685,402 @@ int daos_cleanup(
 
     /* Free DAOS args */
     daos_args_delete(&da);
+
+    return rc;
+}
+
+static int daos_copy_recx_single(daos_key_t *dkey,
+                            daos_handle_t *src_oh,
+                            daos_handle_t *dst_oh,
+                            daos_iod_t *iod)
+{
+    /* if iod_type is single value just fetch iod size from source
+     * and update in destination object */
+    int buf_len = (int)(*iod).iod_size;
+    char buf[buf_len];
+    d_sg_list_t sgl;
+    d_iov_t iov;
+    int rc;
+
+    /* set sgl values */
+    sgl.sg_nr     = 1;
+    sgl.sg_nr_out = 0;
+    sgl.sg_iovs   = &iov;
+    d_iov_set(&iov, buf, buf_len);
+    rc = daos_obj_fetch(*src_oh, DAOS_TX_NONE, 0, dkey, 1, iod, &sgl, NULL, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS object fetch returned with errors: ", MFU_ERRF,
+	        MFU_ERRP(-MFU_ERR_DAOS));
+        return 1;
+    }
+    rc = daos_obj_update(*dst_oh, DAOS_TX_NONE, 0, dkey, 1, iod, &sgl, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS object update returned with errors: ", MFU_ERRF,
+	        MFU_ERRP(-MFU_ERR_DAOS));
+        return 1;
+    }
+    return rc;
+}
+
+static int daos_copy_recx_array(daos_key_t *dkey,
+                           daos_key_t *akey,
+                           daos_handle_t *src_oh,
+                           daos_handle_t *dst_oh,
+                           daos_iod_t *iod)
+{
+    daos_anchor_t recx_anchor = {0}; 
+    int rc;
+    int i;
+    while (!daos_anchor_is_eof(&recx_anchor)) {
+        daos_epoch_range_t	eprs[5];
+        daos_recx_t		recxs[5];
+        daos_size_t		size;
+
+        /* list all recx for this dkey/akey */
+        uint32_t number = 5;
+        rc = daos_obj_list_recx(*src_oh, DAOS_TX_NONE, dkey, akey,
+	                        &size, &number, recxs, eprs,
+				&recx_anchor, true, NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS daos_obj_list_recx returned with errors: ", MFU_ERRF,
+	            MFU_ERRP(-MFU_ERR_DAOS));
+            return 1;
+        }
+
+        /* if no recx is returned for this dkey/akey move on */
+        if (number == 0) 
+            continue;
+
+        for (i = 0; i < number; i++) {
+            uint64_t    buf_len = recxs[i].rx_nr;
+            char        buf[buf_len];
+            d_sg_list_t sgl;
+            d_iov_t     iov;
+
+            /* set iod values */
+            (*iod).iod_type  = DAOS_IOD_ARRAY;
+            (*iod).iod_size  = 1;
+            (*iod).iod_nr    = 1;
+            (*iod).iod_recxs = &recxs[i];
+
+            /* set sgl values */
+            sgl.sg_nr     = 1;
+            sgl.sg_nr_out = 0;
+            sgl.sg_iovs   = &iov;
+
+            /* fetch recx values from source */
+            d_iov_set(&iov, buf, buf_len);	
+            rc = daos_obj_fetch(*src_oh, DAOS_TX_NONE, 0, dkey, 1, iod,
+                                &sgl, NULL, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "DAOS object fetch returned with errors: ", MFU_ERRF,
+	                MFU_ERRP(-MFU_ERR_DAOS));
+                return 1;
+            }
+
+            /* update fetched recx values and place in destination object */
+            rc = daos_obj_update(*dst_oh, DAOS_TX_NONE, 0, dkey, 1, iod,
+                                 &sgl, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "DAOS object update returned with errors: ", MFU_ERRF,
+	                MFU_ERRP(-MFU_ERR_DAOS));
+                return 1;
+            }
+	
+        }
+    }
+    return rc;
+}
+
+static int daos_copy_list_keys(daos_handle_t *src_oh,
+                          daos_handle_t *dst_oh)
+{
+    /* loop to enumerate dkeys */
+    daos_anchor_t dkey_anchor = {0}; 
+    int rc;
+    while (!daos_anchor_is_eof(&dkey_anchor)) {
+        d_sg_list_t     dkey_sgl;
+        d_iov_t         dkey_iov;
+        daos_key_desc_t dkey_kds[ENUM_DESC_NR]       = {0};
+        uint32_t        dkey_number                  = ENUM_DESC_NR;
+        char            dkey_enum_buf[ENUM_DESC_BUF] = {0};
+        char            dkey[ENUM_KEY_BUF]           = {0};
+
+        dkey_sgl.sg_nr     = 1;
+        dkey_sgl.sg_nr_out = 0;
+        dkey_sgl.sg_iovs   = &dkey_iov;
+
+        d_iov_set(&dkey_iov, dkey_enum_buf, ENUM_DESC_BUF);
+
+        /* get dkeys */
+        rc = daos_obj_list_dkey(*src_oh, DAOS_TX_NONE, &dkey_number, dkey_kds,
+                                &dkey_sgl, &dkey_anchor, NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS daos_obj_list_dkey returned with errors: ", MFU_ERRF,
+	            MFU_ERRP(-MFU_ERR_DAOS));
+            return 1;
+        }
+
+        /* if no dkeys were returned move on */
+        if (dkey_number == 0)
+            continue;
+
+        char* dkey_ptr;
+        int   i;
+
+        /* parse out individual dkeys based on key length and numver of dkeys returned */
+        for (dkey_ptr = dkey_enum_buf, i = 0; i < dkey_number; i++) {
+
+            /* Print enumerated dkeys */
+            daos_key_t diov;
+            snprintf(dkey, dkey_kds[i].kd_key_len + 1, "%s", dkey_ptr);
+            d_iov_set(&diov, (void*)dkey, dkey_kds[i].kd_key_len);
+            dkey_ptr += dkey_kds[i].kd_key_len;
+
+            /* loop to enumerate akeys */
+            daos_anchor_t akey_anchor = {0}; 
+            while (!daos_anchor_is_eof(&akey_anchor)) {
+                d_sg_list_t     akey_sgl;
+                d_iov_t         akey_iov;
+                daos_key_desc_t akey_kds[ENUM_DESC_NR]       = {0};
+                uint32_t        akey_number                  = ENUM_DESC_NR;
+                char            akey_enum_buf[ENUM_DESC_BUF] = {0};
+                char            akey[ENUM_KEY_BUF]           = {0};
+
+                akey_sgl.sg_nr     = 1;
+                akey_sgl.sg_nr_out = 0;
+                akey_sgl.sg_iovs   = &akey_iov;
+
+                d_iov_set(&akey_iov, akey_enum_buf, ENUM_DESC_BUF);
+
+                /* get akeys */
+                rc = daos_obj_list_akey(*src_oh, DAOS_TX_NONE, &diov, &akey_number, akey_kds,
+                                        &akey_sgl, &akey_anchor, NULL);
+                if (rc != 0) {
+                    MFU_LOG(MFU_LOG_ERR, "DAOS daos_obj_list_akey returned with errors: ", MFU_ERRF,
+	                    MFU_ERRP(-MFU_ERR_DAOS));
+                    return 1;
+                }
+
+                /* if no akeys returned move on */
+                if (akey_number == 0)
+                    continue;
+
+                int j;
+                char* akey_ptr;
+
+                /* parse out individual akeys based on key length and numver of dkeys returned */
+                for (akey_ptr = akey_enum_buf, j = 0; j < akey_number; j++) {
+                    daos_key_t aiov;
+                    daos_iod_t iod;
+                    daos_recx_t recx;
+                    snprintf(akey, akey_kds[j].kd_key_len + 1, "%s", akey_ptr);
+                    d_iov_set(&aiov, (void*)akey, akey_kds[j].kd_key_len);
+
+                    /* set iod values */
+                    iod.iod_nr    = 1;
+                    iod.iod_type  = DAOS_IOD_SINGLE;
+                    iod.iod_size  = DAOS_REC_ANY;
+                    iod.iod_recxs = NULL;
+
+                    d_iov_set(&iod.iod_name, (void*)akey, strlen(akey));
+
+                    /* Do a fetch (with NULL sgl) of single value type, and if that
+                     * returns iod_size == 0, then a single value does not exist. */
+                    rc = daos_obj_fetch(*src_oh, DAOS_TX_NONE, 0, &diov, 1, &iod, NULL, NULL, NULL);
+                    if (rc != 0) {
+                        MFU_LOG(MFU_LOG_ERR, "DAOS daos_obj_fetch returned with errors: ", MFU_ERRF,
+	                        MFU_ERRP(-MFU_ERR_DAOS));
+                        return 1;
+                    }
+
+                    /* if iod_size == 0 then this is a DAOS_IOD_ARRAY type */
+                    if ((int)iod.iod_size == 0) {
+                        rc = daos_copy_recx_array(&diov, &aiov, src_oh, dst_oh, &iod);
+                        if (rc != 0) {
+                            MFU_LOG(MFU_LOG_ERR, "DAOS daos_copy_recx_array returned with errors: ", MFU_ERRF,
+	                            MFU_ERRP(-MFU_ERR_DAOS));
+                            return 1;
+                        }
+                    } else {
+                        rc = daos_copy_recx_single(&diov, src_oh, dst_oh, &iod);
+                        if (rc != 0) {
+                            MFU_LOG(MFU_LOG_ERR, "DAOS daos_copy_recx_single returned with errors: ", MFU_ERRF,
+	                            MFU_ERRP(-MFU_ERR_DAOS));
+                            return 1;
+                        }
+                    }
+
+                    /* advance to next akey returned */	
+                    akey_ptr += akey_kds[j].kd_key_len;
+                }
+            }
+        }
+    }
+    return rc;
+}
+
+static int daos_obj_copy(
+    daos_args_t* da,
+    mfu_flist bflist)
+{
+    int rc = 0;
+
+    flist_t* flist = (flist_t*) bflist;
+
+    uint64_t i;
+    const elem_t* p = flist->list_head;
+    for (i = 0; i < flist->list_count; i++) {
+        /* open DAOS object based on oid[i] to get obj handle */
+        daos_handle_t oh;
+        daos_obj_id_t oid;
+        oid.lo = p->obj_id_lo;	
+        oid.hi = p->obj_id_hi;	
+        rc = daos_obj_open(da->src_coh, oid, 0, &oh, NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS object open returned with errors: ", MFU_ERRF,
+	            MFU_ERRP(-MFU_ERR_DAOS));
+	    return 1;
+        }
+
+        /* open handle of object in dst container */
+        daos_handle_t dst_oh;
+        rc = daos_obj_open(da->dst_coh, oid, 0, &dst_oh, NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS object open returned with errors: ", MFU_ERRF,
+	            MFU_ERRP(-MFU_ERR_DAOS));
+            /* make sure to close the source if opening dst fails */
+            daos_obj_close(oh, NULL);
+	    return 1;
+        }
+        rc = daos_copy_list_keys(&oh, &dst_oh);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS copy list keys returned with errors: ", MFU_ERRF,
+                    MFU_ERRP(-MFU_ERR_DAOS));
+            /* cleanup object handles on failure */
+            daos_obj_close(oh, NULL);
+            daos_obj_close(dst_oh, NULL);
+	    return 1;
+        }
+
+        /* close source and destination object */
+        daos_obj_close(oh, NULL);
+        daos_obj_close(dst_oh, NULL);
+        p = p->next;
+    }
+
+    return rc;
+}
+
+static int daos_obj_list_oids(daos_args_t* da, daos_epoch_t* epoch, mfu_flist bflist) {
+    /* List objects in src container to be copied to 
+     * destination container */
+    static const int     OID_ARR_SIZE = 50;
+    daos_obj_id_t        oids[OID_ARR_SIZE];
+    daos_anchor_t        anchor;
+    uint32_t             oids_nr;
+    daos_handle_t        toh;
+    uint32_t             oids_total = 0;
+    int                  rc = 0;
+
+    /* create snapshot to pass to object iterator table */
+    rc = daos_cont_create_snap_opt(da->src_coh, epoch, NULL,
+    				   DAOS_SNAP_OPT_CR | DAOS_SNAP_OPT_OIT,
+				   NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS failed to create snapshot: ", MFU_ERRF,
+                MFU_ERRP(-MFU_ERR_DAOS));
+        return 1;
+    }
+
+    /* open object iterator table */
+    rc = daos_oit_open(da->src_coh, *epoch, &toh, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS failed to open oit: ", MFU_ERRF,
+                MFU_ERRP(-MFU_ERR_DAOS));
+        return 1;
+    }
+    memset(&anchor, 0, sizeof(anchor));
+    flist_t* flist = (flist_t*) bflist;
+
+    /* list and store all object ids in flist for this epoch */
+    while (1) {
+        oids_nr = OID_ARR_SIZE;
+        rc = daos_oit_list(toh, oids, &oids_nr, &anchor, NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS daos_oit_list returned with errors: ", MFU_ERRF,
+	            MFU_ERRP(-MFU_ERR_DAOS));
+            daos_oit_close(toh, NULL);
+            return 1;
+        }
+	int i;
+	/* create element in flist for each obj id retrived */
+ 	for (i = 0; i < oids_nr; i++) {
+            uint64_t idx = mfu_flist_file_create(bflist);
+            mfu_flist_file_set_oid(bflist, idx, oids[i]);
+	}
+	oids_total = oids_nr + oids_total;
+ 	if (daos_anchor_is_eof(&anchor)) {
+ 		break;
+ 	}
+    }
+
+    /* close object iterator */
+    rc = daos_oit_close(toh, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS failed to close oit: ", MFU_ERRF,
+                MFU_ERRP(-MFU_ERR_DAOS));
+    }
+    return rc;
+}
+
+int mfu_flist_walk_daos(
+    daos_args_t* da,
+    daos_epoch_t* epoch,
+    mfu_flist flist)
+{
+    /* assume we'll succeed */
+    int rc = 0;
+
+    /* get our rank */
+    int rank;
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+    /* have rank 0 do the work of listing the objects */
+    if (rank == 0) {
+        rc = daos_obj_list_oids(da, epoch, flist);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS failed to list oids: ",
+                MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS));
+        }
+    }
+
+    /* summarize list since we added items to it */
+    mfu_flist_summarize(flist);
+
+    /* broadcast return code from rank 0 so everyone knows whether walk succeeded */
+    MPI_Bcast(&rc, 1, MPI_INT, 0, MPI_COMM_WORLD);
+
+    return rc;
+}
+
+int mfu_flist_copy_daos(
+    daos_args_t* da,
+    mfu_flist flist)
+{
+    /* copy object ids listed in flist to destination in daos args */
+    int rc = daos_obj_copy(da, flist);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS object copy failed: ",
+            MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS));
+    }
+
+    /* wait until all procs are done copying,
+     * and determine whether everyone succeeded */
+    if (! mfu_alltrue(rc == 0, MPI_COMM_WORLD)) {
+        /* someone failed, so return failure on all ranks */
+        rc = 1;
+    }
 
     return rc;
 }
--- a/src/common/mfu_daos.h
+++ b/src/common/mfu_daos.h
@@ -1,6 +1,12 @@
 #include "mfu.h"
 
 #include <daos.h>
+#include "mfu_flist_internal.h"
+
+#define ENUM_KEY_BUF		32 /* size of each dkey/akey */
+#define ENUM_LARGE_KEY_BUF	(512 * 1024) /* 512k large key */
+#define ENUM_DESC_NR		5 /* number of keys/records returned by enum */
+#define ENUM_DESC_BUF		512 /* all keys/records returned by enum */
 
 enum handleType {
     POOL_HANDLE,
@@ -38,7 +44,8 @@ int daos_setup(
   char** argpaths,
   daos_args_t* da,
   mfu_file_t* mfu_src_file,
-  mfu_file_t* mfu_dst_file
+  mfu_file_t* mfu_dst_file,
+  bool* is_posix_copy
 );
 
 /* Unmount DFS.
@@ -48,5 +55,20 @@ int daos_setup(
 int daos_cleanup(
   daos_args_t* da,
   mfu_file_t* mfu_src_file,
-  mfu_file_t* mfu_dst_file
+  mfu_file_t* mfu_dst_file,
+  bool* is_posix_copy
+);
+
+/* walk objects in daos and insert to given flist */
+int mfu_flist_walk_daos(
+    daos_args_t* da,
+    daos_epoch_t* epoch,
+    mfu_flist flist
+);
+
+/* copy objects in flist to destination listed in daos args,
+ * copies DAOS data at object level (non-posix) */
+int mfu_flist_copy_daos(
+    daos_args_t* da,
+    mfu_flist flist
 );
--- a/src/common/mfu_flist.c
+++ b/src/common/mfu_flist.c
@@ -128,6 +128,13 @@ static size_t list_elem_pack2_size(int detail, uint64_t chars, const elem_t* ele
     else {
         size = 2 * 4 + chars + 1 * 4;
     }
+
+    #ifdef DAOS_SUPPORT
+    /* add space for obj_id_lo and obj_id_hi if 
+     * using DAOS */
+    size += 2 * 8;
+    #endif
+
     return size;
 }
 
@@ -146,8 +153,16 @@ static size_t list_elem_pack2(void* buf, int detail, uint64_t chars, const elem_
 
     /* copy in file name */
     char* file = elem->file;
-    strcpy(ptr, file);
+    if (file != NULL) {
+        strcpy(ptr, file);
+    }
     ptr += chars;
+
+#ifdef DAOS_SUPPORT
+    /* copy in values for obj ids */
+    mfu_pack_uint64(&ptr, elem->obj_id_lo);
+    mfu_pack_uint64(&ptr, elem->obj_id_hi);
+#endif
 
     if (detail) {
         /* copy in fields */
@@ -198,6 +213,12 @@ static size_t list_elem_unpack2(const void* buf, elem_t* elem)
 
     elem->detail = (int) detail;
 
+#ifdef DAOS_SUPPORT
+    /* unpack obj ids */
+    mfu_unpack_uint64(&ptr, &elem->obj_id_lo);
+    mfu_unpack_uint64(&ptr, &elem->obj_id_hi);
+#endif
+
     if (detail) {
         /* extract fields */
         mfu_unpack_uint64(&ptr, &elem->mode);
@@ -210,7 +231,6 @@ static size_t list_elem_unpack2(const void* buf, elem_t* elem)
         mfu_unpack_uint64(&ptr, &elem->ctime);
         mfu_unpack_uint64(&ptr, &elem->ctime_nsec);
         mfu_unpack_uint64(&ptr, &elem->size);
-
         /* use mode to set file type */
         elem->type = mfu_flist_mode_to_filetype((mode_t)elem->mode);
     }
@@ -433,10 +453,12 @@ static void list_compute_summary(flist_t* flist)
     uint64_t max_name = 0;
     elem_t* current = flist->list_head;
     while (current != NULL) {
-        uint64_t len = (uint64_t)(strlen(current->file) + 1);
-        if (len > max_name) {
-            max_name = len;
-        }
+        if (current->file != NULL) {
+            uint64_t len = (uint64_t)(strlen(current->file) + 1);
+            if (len > max_name) {
+                max_name = len;
+            }
+	}
 
         int depth = current->depth;
         if (depth < min_depth || min_depth == -1) {
@@ -615,7 +637,7 @@ void mfu_flist_array_free(int levels, mfu_flist** outlists)
     return;
 }
 
-/* return number of files across all procs */
+/* return number of items across all procs */
 uint64_t mfu_flist_global_size(mfu_flist bflist)
 {
     flist_t* flist = (flist_t*) bflist;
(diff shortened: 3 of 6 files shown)


dsikich:
@daltonbohning @adammoody this probably shouldn't be merged before the object iterator PR is merged in DAOS, since this code builds off of that API. I just wanted to submit the PR now and start a review because there is quite a bit of code here, and that PR should land soon I would think.

src/dcp/dcp.c
```diff
@@ -1000,77 +1305,110 @@ int main(int argc, char** argv)
...
-    mfu_param_path_free_all(numpaths, paths);
+	/* free newlist that was created for non-posix copy */
+	mfu_flist_free(&newlist);
+    }
 
-    /* free memory allocated to hold params */
-    mfu_free(&paths);
+    if (rc == 0 && rank == 0) {
+        printf("Successfully copied to DAOS Destination Container.\n");
+    }
```
daltonbohning, line 1400: For consistency, do we want a message like this for DAOS POSIX copy as well?

src/dcp/dcp.c
```diff
@@ -81,7 +87,8 @@ static int daos_check_args(
     int rank,
     char** argpaths,
     daos_args_t* da,
-    int* flag_daos_args)
+    int* flag_daos_args,
+    bool* is_posix_copy)
```
daltonbohning, line 91: To be consistent with some other tools that have DAOS integration, I wonder if we want to use something like `API in (POSIX, DAOS)` instead of `is_posix_copy`
For example, `ior` has these options:
```
-a POSIX # Assumes path is POSIX
-a DAOS # Uses the daos_ API
-a DFS # Uses the dfs_ API
```
For our purposes here, we can probably just use:
```
api = POSIX # POSIX paths, or POSIX containers with dfs_ API
api = DAOS # DAOS container with daos_ api
```

daltonbohning (commented):
@dsikich I still need to test this and do a more thorough review, but this is a start. (I need to get familiar with the daos object api)

src/dcp/dcp.c
```diff
@@ -346,6 +359,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
     return 0;
 }
 
+static int daos_copy_recx_single(daos_key_t *dkey,
```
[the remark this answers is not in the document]
dsikich: @daltonbohning the daos utility that copies non-posix containers has most of the same functions (copy_recx_single, etc.), the only difference is that the daos_obj_copy uses an flist, whereas in the serial daos utility I don't

src/dcp/dcp.c
```diff
@@ -346,6 +359,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+            d_iov_t     iov;
+
+            /* set iod values */
+            (*iod).iod_type  = DAOS_IOD_ARRAY;
+            (*iod).iod_size  = 1;
+            (*iod).iod_nr    = 1;
+            (*iod).iod_recxs = &recxs[i];
+
+            /* set sgl values */
+            sgl.sg_nr     = 1;
```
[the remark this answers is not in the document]
dsikich: @daltonbohning it doesn't quite work like that since sgl just tells you how many segments will be filled, but you can specify the number you want to fetch in the API call. I can look into doing it that way.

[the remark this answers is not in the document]
daltonbohning: @dsikich Anyway, it would just be a performance optimization if it's possible, so we can look into it for a future iteration if needed.

src/common/mfu_flist.c
```diff
@@ -123,10 +123,10 @@ static size_t list_elem_pack2_size(int detail, uint64_t chars, const elem_t* ele
 {
     size_t size;
     if (detail) {
-        size = 2 * 4 + chars + 0 * 4 + 10 * 8;
+        size = 8 * 8 * 2 * 4 + chars + 0 * 4 + 10 * 8;
```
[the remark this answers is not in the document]
dsikich: added

src/common/mfu_flist.c
```diff
@@ -149,6 +149,10 @@ static size_t list_elem_pack2(void* buf, int detail, uint64_t chars, const elem_
     strcpy(ptr, file);
     ptr += chars;
 
+    /* copy in values for obj ids */
+    mfu_pack_uint64(&ptr, elem->obj_id_lo);
+    mfu_pack_uint64(&ptr, elem->obj_id_hi);
+
```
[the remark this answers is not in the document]
dsikich: wrapped these

src/common/mfu_flist.c
```diff
@@ -1448,13 +1465,23 @@ static int map_spread(mfu_flist flist, uint64_t idx, int ranks, const void* args
...
+    } else {
+        global_size = mfu_flist_global_size(flist);
+    }
 
     /* get whole number of items on each rank */
     uint64_t items_per_rank = global_size / (uint64_t)ranks;
 
     /* if global list size is not divisible by the number of ranks
      * then we need to use the remainder */
+    //uint64_t remainder = (global_size) - (items_per_rank * (uint64_t)ranks);
```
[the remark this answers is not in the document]
dsikich: oops, removed

src/dcp/dcp.c
```diff
@@ -346,6 +359,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+                    /* set iod values */
+                    iod.iod_nr   = 1;
+                    iod.iod_type = DAOS_IOD_SINGLE;
+                    iod.iod_size = DAOS_REC_ANY;
+
+                    d_iov_set(&iod.iod_name, (void*)akey, strlen(akey));
+
+                    /* Do a fetch (with NULL sgl) of single value type, and if that
+		     * returns iod_size == 0, then a single value does not exist. */
+                    rc = daos_obj_fetch(*src_oh, DAOS_TX_NONE, 0, &diov, 1, &iod, NULL, NULL, NULL);
```
daltonbohning, line 535: Error check for fetch?

src/dcp/dcp.c
```diff
@@ -346,6 +359,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+
+                    /* if iod_size == 0 then this is a DAOS_IOD_ARRAY type */
+                    if ((int)iod.iod_size == 0) {
+                        rc = daos_copy_recx_array(&diov, &aiov, src_oh, dst_oh, &iod);
+                    } else {
+                        rc = daos_copy_recx_single(&diov, src_oh, dst_oh, &iod);
```
daltonbohning, line 541: I think the copy functions are logging errors here when `rc` is not 0,  but do we need to also check `rc` here as well?
Or maybe it makes sense to let the copy functions just return an error, without logging, and then only log an error here for the `rc`?
Thoughts?

src/dcp/dcp.c
```diff
@@ -346,6 +359,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+    const elem_t* p = flist->list_head;
+    for (i = 0; i < flist->list_count; i++) {
+        /* open DAOS object based on oid[i] to get obj handle */
+        daos_handle_t oh;
+        daos_obj_id_t oid;
+        oid.lo = p->obj_id_lo;	
+        oid.hi = p->obj_id_hi;	
+        rc = daos_obj_open(da->src_coh, oid, 0, &oh, NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS object open returned with errors: ", MFU_ERRF,
```
daltonbohning, line 566: Should we continue if this fails?
Also, I think much of the error handling is probably similar to the DAOS PR you submitted, so I'll stop pointing those out for now, since it's probably redundant. I imagine we'll want a way to centralize these functions anyway. (at least the ones we can)

adammoody:
Which if any of the object ids, dkeys, and akeys could be encoded into strings and decoded back into values?

If there are ways to do that, that opens up the possibility to insert those items as libcircle work elements.

Even if we are stuck with treating them as a string of bytes, we could perhaps use something like uuencode or base64 formats to code them to a string.

src/dcp/dcp.c
```diff
@@ -346,6 +359,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+                                &sgl, &dkey_anchor, NULL);
+        if (rc)
+            return -1;       
+
+        /* if no dkeys were returned move on */
+        if (dkey_number == 0)
+            continue;
+
+        char* ptr;
+        int   rc;
```
daltonbohning, line 478:
```suggestion
```
`rc` is already defined above in this function

src/dcp/dcp.c
```diff
@@ -346,6 +359,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+            /* Print enumerated dkeys */
+            daos_key_t diov;
+            snprintf(dkey, dkey_kds[j].kd_key_len + 1, "%s", ptr);
+            d_iov_set(&diov, (void*)dkey, dkey_kds[j].kd_key_len);
+            ptr += dkey_kds[j].kd_key_len;
+
+            /* loop to enumerate akeys */
+            daos_anchor_t akey_anchor = {0}; 
+            while (!daos_anchor_is_eof(&akey_anchor)) {
+                d_sg_list_t     sgl;
```
daltonbohning, line 493: `sgl` is already defined above in this function.
If we can't reuse it reliably, should we use a different name?

src/dcp/dcp.c
```diff
@@ -346,6 +359,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+                                        &sgl, &akey_anchor, NULL);
+                if (rc)
+                    return -1;       
+
+                /* if no akeys returned move on */
+                if (akey_number == 0)
+                    continue;
+
+                int j;
+                char* ptr;
```
daltonbohning, line 517: `j` and `ptr` are already defined above in this function

daltonbohning (commented):
I got some warnings about shadowed declarations.
Also, I'm getting
```
dcp.c:(.text+0x29f2): undefined reference to `daos_cont_open_oit'
dcp.c:(.text+0x2b59): undefined reference to `daos_cont_list_oit'
dcp.c:(.text+0x2c20): undefined reference to `daos_cont_close_oit'
```
I've built against the latest DAOS master. Wasn't the oit merged?

dsikich:
> I got some warnings about shadowed declarations.
> Also, I'm getting
> 
> ```
> dcp.c:(.text+0x29f2): undefined reference to `daos_cont_open_oit'
> dcp.c:(.text+0x2b59): undefined reference to `daos_cont_list_oit'
> dcp.c:(.text+0x2c20): undefined reference to `daos_cont_close_oit'
> ```
> 
> I've built against the latest DAOS master. Wasn't the oit merged?
> 
> Update: It seems the API in DAOS master is different than what this is using, probably updated after you created this PR.
> This PR:
> 
> * daos_cont_open_oit
> * daos_cont_close_oit
> * daos_cont_list_oit
> 
> In master:
> 
> * daos_oit_open
> * daos_oit_close
> * daos_oit_list

@daltonbohning I think what happened here is the names of the API for the oit table changed, and when I pushed this PR it was using the old names, and so I need to update the API calls in mpifileutils

daltonbohning:
> @daltonbohning I think what happened here is the names of the API for the oit table changed, and when I pushed this PR it was using the old names, and so I need to update the API calls in mpifileutils

@dsikich  When you update this, could you also pull in the latest master changes as well? (removal of svcl). Then, I can use the DAOS test suite to verify different datasets.
In the meantime, I'll do both of these things locally.

daltonbohning:
@dsikich This doesn't seem to be working as I would expect:
```
# The source
> ls -l dfuse/$pool1/$p1cont1/
total 1
-rw-rw-r-- 1 dbohninx dbohninx 6 Dec 15 22:01 file1

> dcp --daos-src-pool $pool1 --daos-src-cont $p1cont1 --daos-dst-pool $pool2 --daos-dst-cont $p2cont2
Successfully copied to DAOS Destination Container.

# The destination
> ls -l dfuse/$pool2/$p2cont2/
total 0
# Should have file1 here

# Using list-obj
> daos cont list-obj --pool $pool1 --cont $p1cont1
1155473290706288642.0
1152922423729848321.0
1152921848204230656.0

> daos cont list-obj --pool $pool2 --cont $p2cont2
1152921848204230656.0
# should have two more objs
```

dsikich:
> @dsikich This doesn't seem to be working as I would expect:
> 
> ```
> # The source
> > ls -l dfuse/$pool1/$p1cont1/
> total 1
> -rw-rw-r-- 1 dbohninx dbohninx 6 Dec 15 22:01 file1
> 
> > dcp --daos-src-pool $pool1 --daos-src-cont $p1cont1 --daos-dst-pool $pool2 --daos-dst-cont $p2cont2
> Successfully copied to DAOS Destination Container.
> 
> # The destination
> > ls -l dfuse/$pool2/$p2cont2/
> total 0
> # Should have file1 here
> 
> # Using list-obj
> > daos cont list-obj --pool $pool1 --cont $p1cont1
> 1155473290706288642.0
> 1152922423729848321.0
> 1152921848204230656.0
> 
> > daos cont list-obj --pool $pool2 --cont $p2cont2
> 1152921848204230656.0
> # should have two more objs
> ```

@daltonbohning alright maybe there is a bug copying across two different pools.

daltonbohning:
> @daltonbohning alright maybe there is a bug copying across two different pools.

It seems to also be the case when using the same pool. Maybe something in the oit API changed?
```
# Copy to same pool
> dcp --daos-src-pool $pool1 --daos-src-cont $p1cont1 --daos-dst-pool $pool1 --daos-dst-cont $p1cont2
Successfully copied to DAOS Destination Container.

# The destination
> ls -l  dfuse/$pool1/$p1cont2/
total 0
```

dsikich:
> > @daltonbohning alright maybe there is a bug copying across two different pools.
> 
> It seems to also be the case when using the same pool. Maybe something in the oit API changed?
> 
> ```
> # Copy to same pool
> > dcp --daos-src-pool $pool1 --daos-src-cont $p1cont1 --daos-dst-pool $pool1 --daos-dst-cont $p1cont2
> Successfully copied to DAOS Destination Container.
> 
> # The destination
> > ls -l  dfuse/$pool1/$p1cont2/
> total 0
> ```

@daltonbohning yeah, maybe because I had it successfully copying data when I was running. I will take a look at this again this week and make sure everything is updated, don't worry about testing it right now, something must be off here after the update.

dsikich:
> > > @daltonbohning alright maybe there is a bug copying across two different pools.
> > 
> > 
> > It seems to also be the case when using the same pool. Maybe something in the oit API changed?
> > ```
> > # Copy to same pool
> > > dcp --daos-src-pool $pool1 --daos-src-cont $p1cont1 --daos-dst-pool $pool1 --daos-dst-cont $p1cont2
> > Successfully copied to DAOS Destination Container.
> > 
> > # The destination
> > > ls -l  dfuse/$pool1/$p1cont2/
> > total 0
> > ```
> 
> @daltonbohning yeah, maybe because I had it successfully copying data when I was running. I will take a look at this again this week and make sure everything is updated, don't worry about testing it right now, something must be off here after the update.

@daltonbohning just reiterating here to keep track OIT API has been updated so that issue at least can be marked as resolved.

daltonbohning:
@dsikich #432 Abstracted the DAOS code into a separate `daos_util.[h,c]`. I think it would be a good idea to continue with this, and move these new functions into that file.

dsikich:
> Which if any of the object ids, dkeys, and akeys could be encoded into strings and decoded back into values?
> 
> If there are ways to do that, that opens up the possibility to insert those items as libcircle work elements.
> 
> Even if we are stuck with treating them as a string of bytes, we could perhaps use something like uuencode or base64 formats to code them to a string.

@adammoody @daltonbohning I'm not sure using libcircle here would provide any benefit. I may still have to think about it some more. From what I understand libcircle works best on a tree-like recursive data structure, so I am trying to wrap my mind around how this would work efficiently for DAOS key enumeration. Each obj id has multiple dkeys, then each dkey can have multiple akeys. So I would have to enumerate all of them at each level within/using libcirlce. I suppose we can enumerate all of the obj -> dkey -> akey using libcircle and store it, but I am not totally clear on all of the details yet. I think I would just have to try it, and I probably won't get to that for a few weeks.

daltonbohning:
> > Which if any of the object ids, dkeys, and akeys could be encoded into strings and decoded back into values?
> > If there are ways to do that, that opens up the possibility to insert those items as libcircle work elements.
> > Even if we are stuck with treating them as a string of bytes, we could perhaps use something like uuencode or base64 formats to code them to a string.
> 
> @adammoody @daltonbohning I'm not sure using libcircle here would provide any benefit. I may still have to think about it some more. From what I understand libcircle works best on a tree-like recursive data structure, so I am trying to wrap my mind around how this would work efficiently for DAOS key enumeration. Each obj id has multiple dkeys, then each dkey can have multiple akeys. So I would have to enumerate all of them at each level within/using libcirlce. I suppose we can enumerate all of the obj -> dkey -> akey using libcircle and store it, but I am not totally clear on all of the details yet. I think I would just have to try it, and I probably won't get to that for a few weeks.

I think rank 0 could insert each `obj id` (`oid`) as libcircle work elements, and then each process can work on the `oids` at the same time. Each `oid` could be enumerated to get the `dkeys`, which could then also be inserted as libcircle work elements. And each `dkey` could be enumerated to get the `akeys`. At this point, each process would have a partial list of `oid + dkey + akey`, which could then be distributed to all processes.
Though, I'm not sure if this would be more efficient than just distributing the `oids` and letting each process directly enumerate the `dkeys`, `akeys`, and perform the copy.
Like you said, @dsikich, I think you might have to do it and run it against some different datasets to see how it performs.

dsikich:
> > > Which if any of the object ids, dkeys, and akeys could be encoded into strings and decoded back into values?
> > > If there are ways to do that, that opens up the possibility to insert those items as libcircle work elements.
> > > Even if we are stuck with treating them as a string of bytes, we could perhaps use something like uuencode or base64 formats to code them to a string.
> > 
> > 
> > @adammoody @daltonbohning I'm not sure using libcircle here would provide any benefit. I may still have to think about it some more. From what I understand libcircle works best on a tree-like recursive data structure, so I am trying to wrap my mind around how this would work efficiently for DAOS key enumeration. Each obj id has multiple dkeys, then each dkey can have multiple akeys. So I would have to enumerate all of them at each level within/using libcirlce. I suppose we can enumerate all of the obj -> dkey -> akey using libcircle and store it, but I am not totally clear on all of the details yet. I think I would just have to try it, and I probably won't get to that for a few weeks.
> 
> I think rank 0 could insert each `obj id` (`oid`) as libcircle work elements, and then each process can work on the `oids` at the same time. Each `oid` could be enumerated to get the `dkeys`, which could then also be inserted as libcircle work elements. And each `dkey` could be enumerated to get the `akeys`. At this point, each process would have a partial list of `oid + dkey + akey`, which could then be distributed to all processes.
> Though, I'm not sure if this would be more efficient than just distributing the `oids` and letting each process directly enumerate the `dkeys`, `akeys`, and perform the copy.
> Like you said, @dsikich, I think you might have to do it and run it against some different datasets to see how it performs.

@daltonbohning yeah, I can't say it would be better without comparing them. In that case sometimes one rank could get stuck with an obj with more data, and to redistribute the data evenly we would have to wait for all processes to finish enumerating. I think I would like to get this version solid, then we add an update that will allow us to add in some other algorithms for performance testings? @adammoody does that sound good?

src/dcp/dcp.c
```diff
@@ -943,77 +1250,110 @@ int main(int argc, char** argv)
...
-    mfu_free(&paths);
+    if (rc == 0 && rank == 0) {
+        printf("Successfully copied to DAOS Destination Container.\n");
+    }
 
-    /* free the input file name */
-    mfu_free(&inputname);
+    /* Cleanup DAOS-related variables, etc. */
+    daos_cleanup(daos_args, mfu_src_file, mfu_dst_file, &is_posix_copy);
+#endif
```
daltonbohning, line 1349: @dsikich Do we want a sanity check in case `is_posix_copy` is false? This should never happen, but just in case some other function modifies the variable accidentally.
```suggestion
#else
    MFU_LOG(MFU_LOG_ERR, "Expected POSIX copy.");
    rc = 1;
#endif
```

src/dcp/dcp.c
```diff
@@ -299,6 +312,293 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS object open returned with errors: ", MFU_ERRF,
+	            MFU_ERRP(-MFU_ERR_DAOS));
+        }
+
+        /* open handle of object in dst container */
+        daos_handle_t dst_oh;
+        rc = daos_obj_open(da->dst_coh, oid, 0, &dst_oh, NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS object open returned with errors: ", MFU_ERRF,
```
daltonbohning, line 529: @dsikich Should we return on error?

src/dcp/dcp.c
```diff
@@ -299,6 +312,293 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+	            MFU_ERRP(-MFU_ERR_DAOS));
+        }
+        rc = daos_copy_list_keys(&oh, &dst_oh);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS copy list keys returned with errors: ", MFU_ERRF,
```
daltonbohning, line 534: @dsikich Should we return on error?

src/dcp/dcp.c
```diff
@@ -299,6 +312,293 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+                    iod.iod_nr    = 1;
+                    iod.iod_type  = DAOS_IOD_SINGLE;
+                    iod.iod_size  = DAOS_REC_ANY;
+		    iod.iod_recxs = NULL;
+
+                    d_iov_set(&iod.iod_name, (void*)akey, strlen(akey));
+
+                    /* Do a fetch (with NULL sgl) of single value type, and if that
+		     * returns iod_size == 0, then a single value does not exist. */
+                    rc = daos_obj_fetch(*src_oh, DAOS_TX_NONE, 0, &diov, 1, &iod, NULL, NULL, NULL);
```
daltonbohning, line 489: @dsikich Error check for fetch?

src/dcp/dcp.c
```diff
@@ -346,6 +359,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+            sgl.sg_iovs   = &iov;
+
+            /* fetch recx values from source */
+            d_iov_set(&iov, buf, buf_len);	
+            rc = daos_obj_fetch(*src_oh, DAOS_TX_NONE, 0, dkey, 1, iod,
+                                &sgl, NULL, NULL);
+
+            /* update fetched recx values and place in destination object */
+            rc = daos_obj_update(*dst_oh, DAOS_TX_NONE, 0, dkey, 1, iod,
+                                 &sgl, NULL);
```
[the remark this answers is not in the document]
daltonbohning: @dsikich

src/dcp/dcp.c
```diff
@@ -299,6 +312,293 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+        dkey_sgl.sg_nr_out = 0;
+        dkey_sgl.sg_iovs   = &dkey_iov;
+
+        d_iov_set(&dkey_iov, dkey_enum_buf, ENUM_DESC_BUF);
+
+        /* get dkeys */
+        rc = daos_obj_list_dkey(*src_oh, DAOS_TX_NONE, &dkey_number, dkey_kds,
+                                &dkey_sgl, &dkey_anchor, NULL);
+        if (rc)
+            return -1;
```
daltonbohning, line 424: @dsikich Should we log the `rc` here so there is an indication of what went wrong?

src/dcp/dcp.c
```diff
@@ -299,6 +312,293 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+                akey_sgl.sg_nr_out = 0;
+                akey_sgl.sg_iovs   = &akey_iov;
+
+                d_iov_set(&akey_iov, akey_enum_buf, ENUM_DESC_BUF);
+
+                /* get akeys */
+                rc = daos_obj_list_akey(*src_oh, DAOS_TX_NONE, &diov, &akey_number, akey_kds,
+                                        &akey_sgl, &akey_anchor, NULL);
+                if (rc)
+                    return -1;
```
daltonbohning, line 462: @dsikich Similar to above, should we log the `rc`?

src/dcp/dcp.c
```diff
@@ -299,6 +312,293 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+    int rc;
+    int i;
+    while (!daos_anchor_is_eof(&recx_anchor)) {
+        daos_epoch_range_t	eprs[5];
+        daos_recx_t		recxs[5];
+        daos_size_t		size;
+
+        /* list all recx for this dkey/akey */
+        uint32_t number = 5;
+        rc = daos_obj_list_recx(*src_oh, DAOS_TX_NONE, dkey, akey,
```
daltonbohning, line 362: @dsikich Return on error?

src/dcp/dcp.c
```diff
@@ -346,6 +359,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS object fetch returned with errors: ", MFU_ERRF,
+	        MFU_ERRP(-MFU_ERR_DAOS));
+    }
+    rc = daos_obj_update(*dst_oh, DAOS_TX_NONE, 0, dkey, 1, iod, &sgl, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS object fetch returned with errors: ", MFU_ERRF,
+	        MFU_ERRP(-MFU_ERR_DAOS));
+    }
+    return rc;
```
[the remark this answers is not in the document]
daltonbohning: @dsikich

src/dcp/dcp.c
daltonbohning: @dsikich @adammoody Thoughts here?

daltonbohning (commented):
@dsikich After adding the barrier as mentioned below, I was able to successfully copy a POSIX structure with ~100 files and a structure with ~1500 directories.
Other than that, my comments are mostly about refined error checking.

src/dcp/dcp.c
```diff
@@ -943,77 +1250,110 @@ int main(int argc, char** argv)
...
-        /* hit some sort of error during copy */
-        rc = 1;
-    }
+        /* evenly spread obj ids among each rank */
+        mfu_flist_summarize(flist);
+        mfu_flist newlist = mfu_flist_spread(flist, &is_posix_copy);
+
+        /* perform copy after oids are spread evenly across all ranks */
+        flist_t* local_flist = (flist_t*) newlist;
+        rc = daos_obj_copy(daos_args, local_flist);
```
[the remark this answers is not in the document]
dsikich: @daltonbohning added

src/dcp/dcp.c
```diff
@@ -943,77 +1250,110 @@ int main(int argc, char** argv)
...
 
-    /* free the path parameters */
-    mfu_param_path_free_all(numpaths, paths);
+	/* free newlist that was created for non-posix copy */
+	mfu_flist_free(&newlist);
+    }
 
-    /* free memory allocated to hold params */
-    mfu_free(&paths);
+    if (rc == 0 && rank == 0) {
```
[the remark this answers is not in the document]
dsikich: @daltonbohning should it be mfu_alltrue(rc, MPI_COMM_WORLD)? Not sure the reason behind explicitly setting rc equal to zero to call the function there?

src/dcp/dcp.c
dsikich: @daltonbohning do you mean return 1 here? I guess I am used to the old way where an error is logged.

src/common/mfu_flist.c
```diff
@@ -123,10 +123,10 @@ static size_t list_elem_pack2_size(int detail, uint64_t chars, const elem_t* ele
 {
     size_t size;
     if (detail) {
-        size = 2 * 4 + chars + 0 * 4 + 10 * 8;
+        size = 2 * 4 + chars + 2 * 8 + 0 * 4 + 10 * 8;
     }
     else {
-        size = 2 * 4 + chars + 1 * 4;
+        size = 2 * 4 + chars + 2 * 8 + 1 * 4;
```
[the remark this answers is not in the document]
daltonbohning: @dsikich What do you think here?

src/dcp/dcp.c
daltonbohning: @dsikich I guess either return 1, or the `rc` from `daos_obj_open`. Since, if we fail to open the object, then I would think we should _**NOT**_ try to copy it.

src/common/mfu_flist.c
daltonbohning, line 129: Since `list_elem_pack2` is wrapped with `#ifdef DAOS_SUPPORT`, maybe we should wrap the size as well?
Maybe we could just keep the size calculation as it was originally, and just do:
```
#ifdef DAOS_SUPPORT
size += 2 * 8; // for obj_id_lo and obj_id_hi
#endif
```

@dsikich What do you think here?

src/dcp/dcp.c
```diff
@@ -442,77 +445,123 @@ int main(int argc, char** argv)
...
 
-    /* free the input file name */
-    mfu_free(&inputname);
+// Synchronize rc across all processes
+if (! mfu_alltrue(rc == 0) {
+    rc = 1;
+}
+
+// Rank 0 prints success if needed
+if (rc == 0 && rank == 0) {
```
daltonbohning, line 548: @dsikich Thanks!
Minor: formatting:
```suggestion
    // Synchronize rc across all processes
    if (! mfu_alltrue(rc == 0) {
        rc = 1;
    }

    // Rank 0 prints success if needed
    if (rc == 0 && rank == 0) {
```

src/dcp/dcp.c
dsikich: @daltonbohning yeah, I just still see some code that is returning -1 in some places like daos_connect, but I am updating the obj_open to return 1.

src/common/mfu_daos.c
```diff
@@ -28,7 +28,8 @@ static int daos_check_args(
     int rank,
     char** argpaths,
     daos_args_t* da,
-    int* flag_daos_args)
+    int* flag_daos_args,
+    bool* is_posix_copy)
```
daltonbohning, line 32: @dsikich @adammoody This is mostly a consideration for the future...

To make mpifileutils more extensible to other filesystems in the future, and to be consistent with some other tools that have DAOS integration, I wonder if we want to use something like `API in (POSIX, DAOS)` instead of `is_posix_copy`
For example, `ior` has these options:
```
-a POSIX # Assumes path is POSIX
-a DAOS # Uses the daos_ API
-a DFS # Uses the dfs_ API
```
For our purposes here, we can probably just use:
```
api = POSIX # POSIX paths, or POSIX containers with dfs_ API
api = DAOS  # DAOS container with daos_ api
```
But again, we don't necessarily have to worry about this right now.

src/dcp/dcp.c
daltonbohning: @dsikich Ah, I see what you mean. Yeah, I think we haven't been strictly consistent with 1 vs -1, etc., although the `daos_*` functions generally return a meaningful error code, so we could either propagate that or just use a generic error code like 1 or -1.

src/common/mfu_flist.c
dsikich: @daltonbohning that is fine, it keeps it more clearly separated

dsikich:
@daltonbohning added a new commit with the extra error checking/logging

src/dcp/dcp.c
```diff
@@ -442,77 +445,123 @@ int main(int argc, char** argv)
...
-        mfu_flist input_flist = mfu_flist_new();
-        mfu_flist_read_cache(inputname, input_flist);
-
-        skip_args.numpaths = numpaths_src;
-        skip_args.paths = paths;
-        mfu_flist_stat(input_flist, flist, input_flist_skip, (void *)&skip_args,
-                       walk_opts->dereference, mfu_src_file);
-        mfu_flist_free(&input_flist);
+        /* wait here until until all procs are done copying,
+	 * then destroy snapshot */
```
daltonbohning, line 527: Minor format (spaces vs tabs):
```suggestion
        /* wait here until until all procs are done copying,
         * then destroy snapshot */
```

src/dcp/dcp.c
```diff
@@ -442,77 +445,123 @@ int main(int argc, char** argv)
...
     }
 
-    /* copy flist into destination */ 
-    rc = mfu_flist_copy(flist, numpaths_src, paths,
-                            destpath, mfu_copy_opts, mfu_src_file,
-                            mfu_dst_file);
-    if (rc < 0) {
-        /* hit some sort of error during copy */
+    // Synchronize rc across all processes
+    if (!mfu_alltrue(rc, MPI_COMM_WORLD) == 0) {
```
daltonbohning, line 543: This doesn't handle all cases.
Consider:
Rank 0; rc = 0
Rank 1; rc = 1

Then, `mfu_alltrue(rc, MPI_COMM_WORLD)`
would perform a logical AND on the list [0,1]
which would be false. So `mfu_alltrue` would return false.
And then `!mfu_alltrue = !false = true`.
But `true != 0`, so the `rc` is not set to 1 for all processes, but it should be.

The idea behind passing `rc == 0` is that the result of `rc == 0` would be true/false.
And `mfu_alltrue` would return true if and only if all processes have a true `rc == 0`.
And if ANY processes has a false `rc == 0`, then `mfu_alltrue` returns false:
```suggestion
    if (!mfu_alltrue(rc == 0, MPI_COMM_WORLD)) {
```

As an example, consider this:
https://github.com/hpc/mpifileutils/blob/871d8f7d96ae01d1bf66c7e9751884f3824ee368/src/common/mfu_decompress_bz2_libcircle.c#L193-L200

src/common/mfu_daos.c
```diff
@@ -677,3 +688,326 @@ int daos_cleanup(
...
+    /* set sgl values */
+    sgl.sg_nr     = 1;
+    sgl.sg_nr_out = 0;
+    sgl.sg_iovs   = &iov;
+    d_iov_set(&iov, buf, buf_len);
+    rc = daos_obj_fetch(*src_oh, DAOS_TX_NONE, 0, dkey, 1, iod, &sgl, NULL, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS object fetch returned with errors: ", MFU_ERRF,
+	        MFU_ERRP(-MFU_ERR_DAOS));
+    }
```
daltonbohning, line 714: Should we skip update if fetch fails?

src/common/mfu_daos.c
```diff
@@ -677,3 +688,326 @@ int daos_cleanup(
...
+    rc = daos_obj_update(*dst_oh, DAOS_TX_NONE, 0, dkey, 1, iod, &sgl, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS object fetch returned with errors: ", MFU_ERRF,
```
daltonbohning, line 717:
```suggestion
        MFU_LOG(MFU_LOG_ERR, "DAOS object update returned with errors: ", MFU_ERRF,
```

src/common/mfu_daos.c
```diff
@@ -677,3 +688,326 @@ int daos_cleanup(
...
+            sgl.sg_iovs   = &iov;
+
+            /* fetch recx values from source */
+            d_iov_set(&iov, buf, buf_len);	
+            rc = daos_obj_fetch(*src_oh, DAOS_TX_NONE, 0, dkey, 1, iod,
+                                &sgl, NULL, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "DAOS object fetch returned with errors: ", MFU_ERRF,
+	                MFU_ERRP(-MFU_ERR_DAOS));
+            }
```
daltonbohning, line 775: Should we skip update if fetch failed?

src/common/mfu_daos.c
```diff
@@ -677,3 +688,326 @@ int daos_cleanup(
...
+        }
+
+        /* open handle of object in dst container */
+        daos_handle_t dst_oh;
+        rc = daos_obj_open(da->dst_coh, oid, 0, &dst_oh, NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS object open returned with errors: ", MFU_ERRF,
+	            MFU_ERRP(-MFU_ERR_DAOS));
+	    return 1;
+        }
```
daltonbohning, line 939: For an error on opening the destination, we need to make sure we close the source:
`daos_obj_close(oh, NULL);`

src/common/mfu_daos.c
```diff
@@ -677,3 +688,326 @@ int daos_cleanup(
...
+        rc = daos_copy_list_keys(&oh, &dst_oh);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS copy list keys returned with errors: ", MFU_ERRF,
+                    MFU_ERRP(-MFU_ERR_DAOS));
+	    return 1;
+        }
```
daltonbohning, line 945: For an error here, we need to make sure we close the source and destination:
```
daos_obj_close(oh, NULL);
daos_obj_close(dst_oh, NULL);
```

src/common/mfu_daos.c
```diff
@@ -677,3 +688,326 @@ int daos_cleanup(
...
+        MFU_LOG(MFU_LOG_ERR, "DAOS failed to create snapshot: ", MFU_ERRF,
+                MFU_ERRP(-MFU_ERR_DAOS));
+    }
+
+    /* open object iterator table */
+    rc = daos_oit_open(da->src_coh, *epoch, &toh, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS failed to open oit: ", MFU_ERRF,
+                MFU_ERRP(-MFU_ERR_DAOS));
+    }
```
daltonbohning, line 980: If `daos_oit_open` fails, should we skip `daos_oit_list` below?

src/common/mfu_daos.c
```diff
@@ -677,3 +688,326 @@ int daos_cleanup(
...
+    int rc;
+
+    /* create snapshot to pass to object iterator table */
+    rc = daos_cont_create_snap_opt(da->src_coh, epoch, NULL,
+    				   DAOS_SNAP_OPT_CR | DAOS_SNAP_OPT_OIT,
+				   NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS failed to create snapshot: ", MFU_ERRF,
+                MFU_ERRP(-MFU_ERR_DAOS));
+    }
```
daltonbohning, line 973: If `daos_cont_create_snap_opt` fails, should we skip `daos_oit_open`?

src/common/mfu_flist.c
```diff
@@ -1448,7 +1502,16 @@ static int map_spread(mfu_flist flist, uint64_t idx, int ranks, const void* args
...
+    uint64_t global_size;
+
+    /* global size will be total obj ids instead of total files,
+     * if a non-posix copy is being performed */
+    bool* is_posix_copy = (bool*)args;
+    if (*is_posix_copy) {
+        global_size = mfu_flist_global_oid_size(flist);
+    } else {
+        global_size = mfu_flist_global_size(flist);
+    }
```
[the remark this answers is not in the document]
dsikich: @daltonbohning ok I updated this to reflect how we do it in mfu_io by just putting it in the function body and returning 0 if DAOS_SUPPORT is not defined

src/common/mfu_daos.c
dsikich: @daltonbohning yes

src/dcp/dcp.c
dsikich: @daltonbohning I am just logging what function failed in these... I like it telling me which function failed. The function itself will return and log which error failed. So you get a message for each level. If that is too much logging I can do that? I don't have a strong opinion either way.

src/common/mfu_flist.c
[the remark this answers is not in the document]
daltonbohning: Yeah, I think that should work

src/common/mfu_flist.c
```diff
@@ -433,10 +454,12 @@ static void list_compute_summary(flist_t* flist)
...
-        uint64_t len = (uint64_t)(strlen(current->file) + 1);
-        if (len > max_name) {
-            max_name = len;
-        }
+        if (current->file != NULL) {
+            uint64_t len = (uint64_t)(strlen(current->file) + 1);
+            if (len > max_name) {
+                max_name = len;
+            }
+	}
```
daltonbohning, line 462: @dsikich Just a thought... I wonder if it could be feasible/useful to convert the oid to a string and use that as the name? This could tie into what we discussed before about being able to write oids to a file, similar to how posix filenames are written to a file in dwalk.
We can address this later when it becomes relevant, but something to keep in mind.

src/common/mfu_flist.c
```diff
@@ -1448,7 +1504,16 @@ static int map_spread(mfu_flist flist, uint64_t idx, int ranks, const void* args
...
+    uint64_t global_size;
+
+    /* global size will be total obj ids instead of total files,
+     * if a non-posix copy is being performed */
+    bool* is_posix_copy = (bool*)args;
+    if (*is_posix_copy) {
+        global_size = mfu_flist_global_oid_size(flist);
+    } else {
+        global_size = mfu_flist_global_size(flist);
+    }
```
daltonbohning, line 1516: @dsikich Minor: Just a thought...
What if `mfu_flist_global_size` did something like this?
```
uint64_t mfu_flist_global_size(mfu_flist bflist)
{
    flist_t* flist = (flist_t*) bflist;
#ifdef DAOS_SUPPORT
    uint64_t total_oids = flist->total_oids;
    if (total_oids > 0) {
        return total_oids;
    }
#endif
    uint64_t val = flist->total_files;
    return val;
}
```
The benefit would be that we wouldn't need `is_posix_copy` to be passed to `mfu_flist_spread`, which would simplify the function header.

src/common/mfu_daos.h
```diff
@@ -38,7 +44,8 @@ int daos_setup(
   char** argpaths,
   daos_args_t* da,
   mfu_file_t* mfu_src_file,
-  mfu_file_t* mfu_dst_file
+  mfu_file_t* mfu_dst_file,
+  bool* is_posix_copy
```
daltonbohning, line 48: @dsikich I still keep getting hung up on the `is_posix_copy` variable, mainly because it wouldn't be extensible if other filesystems are introduced.
We don't necessarily need to address this now, but I had another thought...
What if `mfu_file_t` had three types:
```
enum                 {POSIX, DFS, DAOS} type;
```
`DFS` would be what the existing `DAOS` is now (using the DFS API), and `DAOS` would become the object-level copy (using the DAOS API directly).
Of course, we would want to enforce that `DAOS` can only be used with another `DAOS`

daltonbohning (approved):
@dsikich I just had a few comments about future ideas and considerations.
We might need to revisit some minor things, but overall this appears to be working.

src/dcp/dcp.c
```diff
@@ -442,77 +445,139 @@ int main(int argc, char** argv)
...
+        daos_epoch_t epoch;
+        if (rank == 0) {
+            rc = daos_obj_list_oids(daos_args, &epoch, flist);
+	    if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "DAOS failed to list oids: ", MFU_ERRF,
+                        MFU_ERRP(-MFU_ERR_DAOS));
+                rc = 1;
+                
+            }
+        }
```
adammoody, line 519: looks like some indentation mismatch here, and an extra empty line we could drop

src/dcp/dcp.c
```diff
@@ -442,77 +445,139 @@ int main(int argc, char** argv)
...
+            mfu_flist input_flist = mfu_flist_new();
+            mfu_flist_read_cache(inputname, input_flist);
+
+            skip_args.numpaths = numpaths_src;
+            skip_args.paths = paths;
+            mfu_flist_stat(input_flist, flist, input_flist_skip, (void *)&skip_args,
+                           walk_opts->dereference, mfu_src_file);
+            mfu_flist_free(&input_flist);
+        }
+        /* copy flist into destination */
```
adammoody, line 488:
```suggestion
        }

        /* copy flist into destination */ 
```

src/common/mfu_flist.c
adammoody: @dsikich , yes, I was also looking at whether we could avoid adding the ``is_posix_copy`` parameter to those functions.  Does this suggestion work instead?

adammoody: Taking a little closer look, two more things comes to mind.

After adding elements to an flist, we need to call ``mfu_flist_summarize``, which is a global collective.  All ranks must call this function, even if they didn't change the list.  This internally executes a bunch of collectives to do things like sum up the total number of items in the list (e.g., it calls allreduce and then sets ``total_files``).  You'd want to call this just after the ``daos_obj_list_oids`` function.

By the way, this feels like a "walk" but for DAOS objects.  We could wrap this in a new ``mfu_flist_walk_daos`` function, which would look something like:

```
int mfu_flist_walk_daos(daos_args, epoch, flist) {
  if (rank == 0) {
    daos_obj_list_oids(daos_args, &epoch, flist);
  }
  mfu_flist_summarize(flist);
}
```

Also, I think you could probably just rely on the existing ``total_files`` field rather than creating a separate ``total_oids`` field in the mfu_flist struct.  Although it's called "files", it really is a count of "items".  We may as well just overload it to be the oid count in this case.  That might simplify things quite a bit.

dsikich: @adammoody if I use the total_files field to store the number of objects I think it would be better to rename it to "total_items"? So that someone looking at it doesn't thing these are files.

src/common/mfu_flist.c
dsikich: @daltonbohning I think we could. I was going to try and do this with the serialization but it ended up not being necessary. I don't like that every time you go to do a key lookup you will have to decode the string (the whole list of strings I imagine after reading it in). I think trying this would be a good idea though so that we can try out using libcircle, and see if it helps performance.

src/dcp/dcp.c
dsikich: @adammoody I keep forgetting to switch my vim configuration to not use tabs. DAOS only uses tabs, and going back and forth can be a pain :-)

src/common/mfu_daos.h
dsikich: @daltonbohning this was more along the lines of what I was thinking, so I like it better than the other option you proposed. Using the posix variable for this version version was to help me quickly separate the code that was only used for POSIX copies, but I agree it should fit into the framework we already have.

src/common/mfu_flist.c
dsikich: @adammoody there is a call to mfu_flist_summarize after listing the oids. I have daos_obj_list_oids, then I use a barrier to wait for rank 0 to retrieve all of the kids, then I call summarize, then mfu_flist_spread. Is this correct?

dsikich: @adammoody I think the barrier I put in is unnecessary before mfu_flist_summarize

dsikich: @adammoody I used @daltonbohning suggestion to remove the is_posix_copy variable from being passed to mfu_flist_spread by just checking if total_oids > 0.

src/common/mfu_daos.h
dsikich: @daltonbohning I do think looking through the code it is going to take a good amount of renaming and changes to do it this way, even if they are simple, so I think it would be better to improve it this way in another PR.

src/common/mfu_flist.c
adammoody: > 
> 
> @adammoody if I use the total_files field to store the number of objects I think it would be better to rename it to "total_items"? So that someone looking at it doesn't think these are files.

Yes, we can rename the field from total_files to total_items.  It's good by me if you want to do that now or in a separate PR.

adammoody: > 
> 
> @adammoody I think the barrier I put in is unnecessary before mfu_flist_summarize

That's right.  The mfu_flist_summarize includes synchronizing collectives, so it essentially has a barrier built in.

src/dcp/dcp.c
adammoody: Oh, right.  Yeah, that's annoying.  Sorry about that.  We could look at the astyle options to automate the conversion if you have to copy in a bunch of code.

src/common/mfu_daos.h
daltonbohning: @dsikich I'm okay with that, as long as @adammoody agrees.

src/common/mfu_flist.c
daltonbohning: @dsikich Yeah, we can worry about this separately when we look at performance, etc.

dsikich:
@adammoody is this good to merge? or is there more changes you'd like me to make first?

src/common/mfu_flist.c
```diff
@@ -1448,7 +1497,11 @@ static int map_spread(mfu_flist flist, uint64_t idx, int ranks, const void* args
     uint64_t global_idx = offset + idx;
 
     /* get global size of the list */
-    uint64_t global_size = mfu_flist_global_size(flist);
+    uint64_t global_size;
+
+    /* global size will be total obj ids instead of total files,
+     * if a non-posix copy is being performed */
+    global_size = mfu_flist_global_size(flist);
```
adammoody, line 1504:
```suggestion
    uint64_t global_size = mfu_flist_global_size(flist);
```

src/common/mfu_flist_chmod.c
```diff
@@ -1198,6 +1198,7 @@ void mfu_flist_chmod(
         double start = MPI_Wtime();
 
         /* spread items for this level evenly over all procs */
+        bool is_posix_copy = true;
```
adammoody, line 1201:
```suggestion
```

src/common/mfu_flist_remove.c
```diff
@@ -150,6 +150,7 @@ static void remove_spread(mfu_flist flist, uint64_t* rmcount)
 {
     /* evenly spread flist among processes,
      * execute direct delete, and free temp list */
+    bool is_posix_copy = true;
```
adammoody, line 153:
```suggestion
```

src/common/mfu_flist_remove.c
```diff
@@ -601,6 +602,7 @@ void mfu_flist_unlink(mfu_flist flist, bool traceless)
 
     /* if traceless, restore the stat of each item's pdir */
     if (traceless) {
+        bool is_posix_copy = true;
```
adammoody, line 605:
```suggestion
```

src/dcp/dcp.c
```diff
@@ -442,77 +445,136 @@ int main(int argc, char** argv)
...
+        /* copy flist into destination */ 
+        rc = mfu_flist_copy(flist, numpaths_src, paths,
+                            destpath, mfu_copy_opts, mfu_src_file,
+                            mfu_dst_file);
+        if (rc < 0) {
+            /* hit some sort of error during copy */
+            rc = 1;
         }
+        /* free the file list */
+        mfu_flist_free(&flist);
```
adammoody, line 498: Just adding a blank line here.
```suggestion

        /* free the file list */
        mfu_flist_free(&flist);
```

src/dcp/dcp.c
```diff
@@ -442,77 +445,136 @@ int main(int argc, char** argv)
...
-    mfu_flist_free(&flist);
-
-    /* free the path parameters */
-    mfu_param_path_free_all(numpaths, paths);
-
-    /* free memory allocated to hold params */
-    mfu_free(&paths);
+    // Rank 0 prints success if needed
+    if (rc == 0 && rank == 0 && !is_posix_copy) {
+        printf("Successfully copied to DAOS Destination Container.\n");
```
adammoody, line 563: Just converting to MFU_LOG and dropping the newline.
```suggestion
        MFU_LOG(MFU_LOG_INFO, "Successfully copied to DAOS Destination Container.");
```

dsikich:
@adammoody ok, thanks! I added those by committing your suggestions.

src/common/mfu_flist.c
```diff
@@ -615,10 +639,16 @@ void mfu_flist_array_free(int levels, mfu_flist** outlists)
...
-/* return number of files across all procs */
+/* return number of files or oids across all procs */
 uint64_t mfu_flist_global_size(mfu_flist bflist)
 {
     flist_t* flist = (flist_t*) bflist;
+#ifdef DAOS_SUPPORT
+    uint64_t total_oids = flist->total_oids;
+    if (total_oids > 0) {
+        return total_oids;
+    }
```
adammoody, line 650: Is the intent for this to fall through and return flist->total_files if oids == 0?

dsikich: @adammoody yes, but now that I think about it that doesn't necessary mean we should return total_files if that is the case. I could init total_oids to -1, and check for that instead?

dsikich: @adammoody since now that I think about it someone could be trying to perform a copy where there are no oids (or its empty).

src/dcp/dcp.c
```diff
@@ -442,77 +445,137 @@ int main(int argc, char** argv)
...
-
-    /* free the path parameters */
-    mfu_param_path_free_all(numpaths, paths);
-
-    /* free memory allocated to hold params */
-    mfu_free(&paths);
+    // Rank 0 prints success if needed
+    if (rc == 0 && rank == 0 && !is_posix_copy) {
+        MFU_LOG(MFU_LOG_INFO, "Successfully copied to DAOS Destination Container.");
+    }
```
adammoody, line 564: Should this go inside of the else branch?

src/dcp/dcp.c
```diff
@@ -442,77 +445,137 @@ int main(int argc, char** argv)
...
-    /* copy flist into destination */ 
-    rc = mfu_flist_copy(flist, numpaths_src, paths,
-                            destpath, mfu_copy_opts, mfu_src_file,
-                            mfu_dst_file);
-    if (rc < 0) {
-        /* hit some sort of error during copy */
+    // Synchronize rc across all processes
+    if (!mfu_alltrue(rc == 0, MPI_COMM_WORLD)) {
         rc = 1;
     }
```
adammoody, line 559: Also, can this one go inside the else branch for DAOS above?

src/dcp/dcp.c
```diff
@@ -442,77 +445,137 @@ int main(int argc, char** argv)
...
+    }
 
-    /* free the input file name */
-    mfu_free(&inputname);
+    /* Cleanup DAOS-related variables, etc. */
+    daos_cleanup(daos_args, mfu_src_file, mfu_dst_file, &is_posix_copy);
+#else
+    MFU_LOG(MFU_LOG_ERR, "Expected POSIX copy.");
+    rc = 1;
+#endif
```
adammoody, line 571: Is this code placed correctly is someone were to build dcp without DAOS enabled?  It'd hard to follow the #ifdef and if/else blocks, but it looks like it would always execute if DAOS_SUPPORT is not defined.

src/common/mfu_flist.c
dsikich: @adammoody well I actually think this is ok. Since if someone was actually trying to perform a posix copy it will fall through and return total_files. And if total_oids is zero, but the intent was to do a cont copy, then no copy will be performed (since they are both zero).

src/dcp/dcp.c
dsikich: @adammoody sure adding

src/dcp/dcp.c
dsikich: @adammoody added

src/dcp/dcp.c
dsikich: @adammoody yes, good catch. I just tested with/without DAOS_SUPPORT turned on and fixed this issue.

dsikich:
@adammoody Thanks Adam, I moved your updates to this PR. It didn't quite compile, but I made the necessary updates and added them. I also tested this with/without DAOS_SUPPORT on, and it looks like it is working. I also added the style updates you mentioned.

src/dcp/dcp.c
daltonbohning: It looks like this was updated to check the `rc` and log an error here as well. Thanks!

src/common/mfu_flist.c
daltonbohning: @dsikich That's what I was thinking as well, though the concept of `total_items` would ultimately be simpler.

daltonbohning (approved):
@dsikich @adammoody I think the only outstanding issue/suggestion on my end is whether to update `is_posix_copy` to something like `mfu_file->type == DFS` on this PR or a future PR.
Otherwise, this seems to be in order.

dsikich:
> @dsikich @adammoody I think the only outstanding issue/suggestion on my end is whether to update `is_posix_copy` to something like `mfu_file->type == DFS` on this PR or a future PR.
> Otherwise, this seems to be in order.

@daltonbohning @adammoody thanks, I would prefer another PR. I already looked at what all would need to be renamed/updated, and it is going to require quite a few changes. I think it will be easier to review in a separate PR, since it won't have all of the other stuff in it as well.

adammoody:
Thanks @dsikich and @daltonbohning .  This looks good to me.

Although, it looks like it may need some conflict resolution and a rebase before we can merge?

dsikich:
> Thanks @dsikich and @daltonbohning . This looks good to me.
> 
> Although, it looks like it may need some conflict resolution and a rebase before we can merge?

@adammoody @daltonbohning I fixed the merge conflicts with master. But some of the commits will need to be squashed.

daltonbohning:
@dsikich @adammoody It looks like this was "merged" into master instead of "squash and merge" or "rebase", which means there is a "merge commit" at HEAD. I believe this will cause travis to fail, since the "merge commit" is not signed off. (I made this mistake the first time I "merged" to master.)

I think there are 2 approaches here:
1. Ignore the failure, until the next "commit" is signed off.
2. rebase master to remove the "merge commit", and force push that to hpc/master. I believe this is as simple as:
```
git checkout master
git fetch
git pull
git rebase HEAD~1
git push -f
```
And to verify we didn't accidentally clobber something, this should return NO DIFFS:
```
git diff origin/master
```

dsikich:
@daltonbohning yeah its been a while since I merged to this repo, I just hit "merge". I removed the merge commit, thanks.

adammoody:
Thanks again @dsikich and @daltonbohning !  Have a good weekend.

PR #429, opened 2020-12-05
DAOS: Update I/O functions with O_NOFOLLOW

daltonbohning:
The DFS API has had support added for dereferencing
symlinks in dfs_lookup and dfs_lookup_rel.

- Removed daos_dereference_symlink
  - No longer needed
- Modified calls to dfs_lookup_rel to pass O_NOFOLLOW when needed
- Modified daos_open to use dfs_lookup_rel
- Modified daos_opendir to use dfs_lookup_rel

### DIFF
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -596,8 +596,8 @@ int daos_utimensat(int dirfd, const char* pathname, const struct timespec times[
     } else {
         /* Lookup the object */
         dfs_obj_t* obj;
-	int lookup_flags = O_RDWR;
-	if (flags & AT_SYMLINK_NOFOLLOW) {
+        int lookup_flags = O_RDWR;
+        if (flags & AT_SYMLINK_NOFOLLOW) {
             lookup_flags |= O_NOFOLLOW;
         }
         rc = dfs_lookup_rel(mfu_file->dfs, parent, name, lookup_flags, &obj, NULL, NULL);
@@ -2087,12 +2087,12 @@ DIR* daos_opendir(const char* dir, mfu_file_t* mfu_file)
             if (rc) {
                 MFU_LOG(MFU_LOG_ERR, "dfs_lookup_rel %s/%s failed", dir_name, name);
                 errno = rc;
-    	        goto err_dirp;
+                goto err_dirp;
             } else {
                 if (!S_ISDIR(mode)) {
                     errno = ENOTDIR;
                     goto err_dirp;
-    	        }
+                }
             }
         }
     }
@@ -2214,7 +2214,7 @@ struct dirent* daos_readdir(DIR* _dirp, mfu_file_t* mfu_file)
     while (!daos_anchor_is_eof(&dirp->anchor)) {
         rc = dfs_readdir(mfu_file->dfs, dirp->dir,
                          &dirp->anchor, &dirp->num_ents,
-	                 dirp->ents);
+                         dirp->ents);
         if (rc) {
             MFU_LOG(MFU_LOG_ERR, "dfs_readdir failed (%d %s)", rc, strerror(rc));
             dirp->num_ents = 0;
@@ -2222,9 +2222,9 @@ struct dirent* daos_readdir(DIR* _dirp, mfu_file_t* mfu_file)
             return NULL;
         }
         if (dirp->num_ents == 0) {
-	    continue;
+            continue;
         }
-	goto ret;
+        goto ret;
     }
     assert(daos_anchor_is_eof(&dirp->anchor));
     return NULL;
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -410,6 +410,15 @@ int daos_faccessat(int dirfd, const char* path, int amode, int flags, mfu_file_t
                     rc = -1;
                 }
             }
+
+            /* Release the obj */
+            int tmp_rc = dfs_release(obj);
+            if (tmp_rc && (rc != -1)) {
+                MFU_LOG(MFU_LOG_ERR, "dfs_release %s failed (%d %s)",
+                        name, tmp_rc, strerror(tmp_rc));
+                errno = tmp_rc;
+                rc = -1;
+            }
         }
     }
 
@@ -623,8 +632,8 @@ int daos_utimensat(int dirfd, const char* pathname, const struct timespec times[
             int tmp_rc = dfs_release(obj);
             if (tmp_rc && (rc != -1)) {
                 MFU_LOG(MFU_LOG_ERR, "dfs_release %s failed (%d %s)",
-                        pathname, rc, strerror(rc));
-                errno = rc;
+                        pathname, tmp_rc, strerror(tmp_rc));
+                errno = tmp_rc;
                 rc = -1;
             }
         }
@@ -1765,9 +1774,9 @@ int daos_truncate(const char* file, off_t length, mfu_file_t* mfu_file)
 
             /* close the obj */
             int tmp_rc = dfs_release(mfu_file->obj);
-            if (!rc && tmp_rc) {
+            if (tmp_rc && (rc != -1)) {
                 MFU_LOG(MFU_LOG_ERR, "dfs_release %s failed (%d %s)",
-                        file, rc, strerror(rc));
+                        file, tmp_rc, strerror(tmp_rc));
                 errno = tmp_rc;
                 rc = -1;
             }
@@ -2085,12 +2094,17 @@ DIR* daos_opendir(const char* dir, mfu_file_t* mfu_file)
             int rc = dfs_lookup_rel(mfu_file->dfs, parent, name, O_RDWR, &dirp->dir,
                                    &mode, NULL);
             if (rc) {
-                MFU_LOG(MFU_LOG_ERR, "dfs_lookup_rel %s/%s failed", dir_name, name);
+                MFU_LOG(MFU_LOG_ERR, "dfs_lookup_rel %s failed", dir);
                 errno = rc;
                 goto err_dirp;
             } else {
                 if (!S_ISDIR(mode)) {
                     errno = ENOTDIR;
+                    rc = dfs_release(dirp->dir);
+                    if (rc) {
+                        MFU_LOG(MFU_LOG_ERR, "dfs_release %s failed (%d %s)",
+                                dir, rc, strerror(rc));
+                    }
                     goto err_dirp;
                 }
             }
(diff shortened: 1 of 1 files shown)


daltonbohning:
@adammoody For what it's worth, these changes are only within `#ifdef DAOS_SUPPORT` blocks. Also, we have plans to move these `daos_*` I/O functions internally into DAOS, which would abstract their definitions from mpifileutils entirely.
So, if it's easier for you, it's okay if you just want to do a quick review for the sake of your own time.

src/common/mfu_io.c
```diff
@@ -646,23 +596,16 @@ int daos_utimensat(int dirfd, const char* pathname, const struct timespec times[
     } else {
         /* Lookup the object */
         dfs_obj_t* obj;
-        rc = dfs_lookup_rel(mfu_file->dfs, parent, name, O_RDWR, &obj, NULL, NULL);
+	int lookup_flags = O_RDWR;
+	if (flags & AT_SYMLINK_NOFOLLOW) {
```
adammoody, line 600: missing some spaces?

src/common/mfu_io.c
```diff
@@ -646,23 +596,16 @@ int daos_utimensat(int dirfd, const char* pathname, const struct timespec times[
...
+            lookup_flags |= O_NOFOLLOW;
+        }
+        rc = dfs_lookup_rel(mfu_file->dfs, parent, name, lookup_flags, &obj, NULL, NULL);
```
adammoody, line 603: I'm not familiar with DAOS, but if this call fails with an error, is it still safe to call ``dfs_release(obj)`` or would that cause a problem?

src/common/mfu_io.c
```diff
@@ -2101,14 +2060,53 @@ DIR* daos_opendir(const char* dir, mfu_file_t* mfu_file)
...
+        if (parent == NULL) {
+            goto err_dirp;
+        } else {
+            mode_t mode;
+            int rc = dfs_lookup_rel(mfu_file->dfs, parent, name, O_RDWR, &dirp->dir,
+                                   &mode, NULL);
+            if (rc) {
+                MFU_LOG(MFU_LOG_ERR, "dfs_lookup_rel %s/%s failed", dir_name, name);
+                errno = rc;
+    	        goto err_dirp;
```
adammoody, line 2090: spacing?

src/common/mfu_io.c
```diff
@@ -2101,14 +2060,53 @@ DIR* daos_opendir(const char* dir, mfu_file_t* mfu_file)
...
+            } else {
+                if (!S_ISDIR(mode)) {
+                    errno = ENOTDIR;
+                    goto err_dirp;
+    	        }
```
adammoody, line 2095: spacing

src/common/mfu_io.c
daltonbohning: Ah, it was tabs instead of spaces 👍

src/common/mfu_io.c
daltonbohning: If dfs_lookup_rel fails, then dfs_release would return an error. 
But it does look like we're only calling dfs_release when lookup did NOT fail, right?

src/common/mfu_io.c
daltonbohning: 👍 Another case of tabs instead of spaces. Thanks!

adammoody:
@daltonbohning , thanks for adding this in.  It's great to see updates to our code here to keep up with new DAOS improvements.  I had a chance to skim the changes, and it all looks good to me.

daltonbohning:
@adammoody Thanks for taking the time to review! Internally, we have tests to verify DAOS integration with mpifileutils, so hopefully we'll be able to easily keep up with any issues in the future.

src/common/mfu_io.c
adammoody: Ah, yes.  Sorry, I skimmed to quickly and missed the else branch here.

daltonbohning: 👍 No worries!

src/common/mfu_io.c
```diff
@@ -2101,14 +2060,53 @@ DIR* daos_opendir(const char* dir, mfu_file_t* mfu_file)
...
+            int rc = dfs_lookup_rel(mfu_file->dfs, parent, name, O_RDWR, &dirp->dir,
+                                   &mode, NULL);
+            if (rc) {
+                MFU_LOG(MFU_LOG_ERR, "dfs_lookup_rel %s/%s failed", dir_name, name);
+                errno = rc;
+                goto err_dirp;
+            } else {
+                if (!S_ISDIR(mode)) {
+                    errno = ENOTDIR;
+                    goto err_dirp;
```
adammoody, line 2094: Does this need a release of dirp->dir?

src/common/mfu_io.c
```diff
@@ -2315,7 +2313,8 @@ ssize_t daos_llistxattr(const char* path, char* list, size_t size, mfu_file_t* m
     else {
         /* lookup and open name */
         dfs_obj_t* obj;
-        int rc = dfs_lookup_rel(mfu_file->dfs, parent, name, O_RDWR, &obj, NULL, NULL);
+        int lookup_flags = O_RDWR | O_NOFOLLOW;
+        int rc = dfs_lookup_rel(mfu_file->dfs, parent, name, lookup_flags, &obj, NULL, NULL);
```
adammoody, line 2317: This might be the spot I noticed before, where it looks like dfs_release might be called after an error.

src/common/mfu_io.c
```diff
@@ -2486,7 +2476,8 @@ ssize_t daos_lgetxattr(const char* path, const char* name, void* value, size_t s
     else {
         /* lookup and open obj_name */
         dfs_obj_t* obj;
-        int rc = dfs_lookup_rel(mfu_file->dfs, parent, obj_name, O_RDWR, &obj, NULL, NULL);
+        int lookup_flags = O_RDWR | O_NOFOLLOW;
+        int rc = dfs_lookup_rel(mfu_file->dfs, parent, obj_name, lookup_flags, &obj, NULL, NULL);
```
adammoody, line 2480: And if the above is right, also here?

src/common/mfu_io.c
```diff
@@ -2657,7 +2639,8 @@ int daos_lsetxattr(const char* path, const char* name, const void* value, size_t
     else {
         /* lookup and open obj_name */
         dfs_obj_t* obj;
-        rc = dfs_lookup_rel(mfu_file->dfs, parent, obj_name, O_RDWR, &obj, NULL, NULL);
+        int lookup_flags = O_RDWR | O_NOFOLLOW;
+        rc = dfs_lookup_rel(mfu_file->dfs, parent, obj_name, lookup_flags, &obj, NULL, NULL);
```
adammoody, line 2643: and here?

src/common/mfu_io.c
```diff
@@ -444,7 +390,11 @@ int daos_faccessat(int dirfd, const char* path, int amode, int flags, mfu_file_t
         /* Get the mode of the object */
         mode_t mode;
         dfs_obj_t* obj;
-        rc = dfs_lookup_rel(mfu_file->dfs, parent, name, O_RDWR, &obj, &mode, NULL);
+        int lookup_flags = O_RDWR;
+        if (flags & AT_SYMLINK_NOFOLLOW) {
+            lookup_flags |= O_NOFOLLOW;
+        }
+        rc = dfs_lookup_rel(mfu_file->dfs, parent, name, lookup_flags, &obj, &mode, NULL);
```
adammoody, line 397: Do we need a release for this call?

src/common/mfu_io.c
daltonbohning: Ah, yes!

src/common/mfu_io.c
daltonbohning: The release for this checks `if (rc && (got_size != -1))`, so release isn't called after an error from dfs_lookup_rel.
But... release sometimes isn't called after an error from dfs_listxattr, but it SHOULD be, because that handle is still open.
Thanks for noticing this!

daltonbohning: Edit: No, you're right. It WAS being called after dfs_lookup failed 👍

src/common/mfu_io.c
daltonbohning: 👍 Yes, thank you

src/common/mfu_io.c
daltonbohning: 👍

src/common/mfu_io.c
daltonbohning: Sloppy of me...

adammoody (approved):
Looks good.  Thanks @daltonbohning

dsikich:
@daltonbohning I haven't followed this all that well, what is the purpose of using dfs_lookup_rel vs dfs_lookup if both have symlink support?

src/common/mfu_io.c
```diff
@@ -908,11 +840,14 @@ int daos_mknod(const char* path, mode_t mode, dev_t dev, mfu_file_t* mfu_file)
...
+                      dfs_mode, O_CREAT | O_EXCL,
                       0, 0, NULL, &(mfu_file->obj));
         if (rc) {
-            MFU_LOG(MFU_LOG_ERR, "dfs_open %s failed (%d %s)",
-                    name, rc, strerror(rc));
+            /* Avoid excessive logging */
+            if (rc != EEXIST) {
+                MFU_LOG(MFU_LOG_ERR, "dfs_open %s failed (%d %s)",
+                        name, rc, strerror(rc));
+            }
```
dsikich, line 850: @daltonbohning so are we not printing a message if the file exists already?

src/common/mfu_io.c
```diff
@@ -1218,12 +1159,39 @@ int daos_open(const char* file, int flags, mode_t mode, mfu_file_t* mfu_file)
...
+                errno = EEXIST;
+                rc = -1;
+            } else if (!S_ISREG(obj_mode)) {
+                /* ... but the obj isn't a file */
+                dfs_release(mfu_file->obj);
+                MFU_LOG(MFU_LOG_ERR, "Invalid entry type (not a file)");
+                errno = EINVAL;
+                rc = -1;
+            } else {
+                /* good */
```
dsikich, line 1179: @daltonbohning is the empty else block necessary?

dsikich:
@daltonbohning I think this is good other than the minor things I asked about.

dsikich (approved):

src/common/mfu_io.c
daltonbohning: @dsikich In the code that calls mfu_file_mknod, a message is logged if the file already exists. So the idea here is to just not print what would essentially be a duplicate message.
Once we move these I/O functions into daos, I imagine we might want to remove some of the errors in general, and just return a proper rc/errno and let the caller handle it. I.e. the posix equivalents don't always print messages like this anyway.

src/common/mfu_io.c
daltonbohning: @dsikich No, it's not necessary. I adapted this style from somewhere else in this repository... though now I'm not sure where.
I can remove it if you like, but I think we'll probably convert the cascading if/else statements into the "goto" style when we move them to daos, since that seems to be the standard in daos.

daltonbohning:
> @daltonbohning I haven't followed this all that well, what is the purpose of using dfs_lookup_rel vs dfs_lookup if both have symlink support?

@dsikich Just performance, really. There are two options:
1. Use dfs_lookup - This will traverse the entire path.
2. Lookup the dirname in the hash, and then use dfs_lookup_rel on the basename/leaf - This allows the parent to be cached.

In dfuse anyway, almost everything uses dfs_lookup_rel, and only uses dfs_lookup for the root "/", as a starting point.

src/common/mfu_io.c
dsikich: @daltonbohning if you prefer it I don't mind

PR #430, opened 2020-12-15
Dtar

adammoody:
Updates dtar to enable:
- parallel creation of PAX-format tarballs
- parallel extraction of general tarballs
- faster parallel extraction of tarballs that meet certain format restrictions

This currently only supports creating tarballs of directories, regular files, and symlinks (object types that can be represented in an ``mfu_flist``).  It can currently extract a wider set of item types, though that might also be reduced.

One can currently create tarballs in PAX-format without compression and without support for sparse encoding of file data.  This restriction enables direct computation of the starting offset for each entry before the archive is written.  Future work may relax these restrictions.

To support extraction of general tar formats, including compressed tarballs and those containing global header entries, every process must scan through the archive from the beginning of the file.

For a subset of tarballs, it should be possible to create an index file listing the starting offset for the header and data of each entry to enable faster parallel extraction.  Known restrictions for those tarballs include at least:
- no compression
- no use of global headers (all header info affecting an entry should be included in that entry's header)
- no sparse encoding of entry data (this might be supported if each process processes the list of sparse blocks starting from the entry header)


adilger:
Some thoughts I had while reading about this:
- rather than having all of the processes read the tarball to find file offsets (which causes N-times reading of the file data) `dtar` could queue the next starting file offset to the other processes, and processes which are idle would grab the next file from the tarball for extraction, like a producer-consumer.  It isn't clear whether having multiple producers would be faster than a single one, since each thread would end up having to do the same read work, and it is unlikely that IOPS are faster than MPI messages
- some compression formats allow parallel decompression and/or restart in the middle of the file.  In such cases, it might make sense to subdivide the file into N chunks and read until a restart marker is found, then start decompression looking for a tar file header.  At a minimum, the data decompression could be parallelized, even if finding the start of the next file can't be found.
- storing the index could be done in an xattr (up to 64KB for most filesystems today) so that it stays with the file during renames, and at least some transfers

adammoody:
@dsikich and @daltonbohning , I plan to check in this fist version of dtar without daos support.  I'm still learning about libarchive, but I think it may have callbacks that could let us integrate with daos.  We can do that in step 2.  A daos dtar would be awesome, but I've got my hands full with getting posix working.

I'd thinking we can plan to make a mpiFileUtils release in mid to late January to pull in your daos changes and hopefully at least a posix-enabled dtar.

adammoody:
Thanks for looking this over and contributing those ideas, @adilger .

daltonbohning:
@adammoody

> @dsikich and @daltonbohning , I plan to check in this fist version of dtar without daos support. I'm still learning about libarchive, but I think it may have callbacks that could let us integrate with daos. We can do that in step 2. A daos dtar would be awesome, but I've got my hands full with getting posix working.

I can see how this would be useful to daos - we are also working on some other serialization tools. I haven't yet heard mention specifically of a daos compressed tarball, but I'll look into it.
Anyway, if POSIX works, then dtar would technically work with daos mounted over dfuse, though not as performant.

> I'd thinking we can plan to make a mpiFileUtils release in mid to late January to pull in your daos changes and hopefully at least a posix-enabled dtar.

Great, thank for the heads up! Let me know if there's anything to be done on the daos end for this release, etc.

adilger:
One other comment to make here - when extracting xattrs from the tarball, they should preferentially be set *before* the body of the file is extracted. That allows backup and restore of Lustre layout xattrs (trusted.lov).

adammoody:
> 
> 
> One other comment to make here - when extracting xattrs from the tarball, they should preferentially be set _before_ the body of the file is extracted. That allows backup and restore of Lustre layout xattrs (trusted.lov).
> 
> Newer GNU tar does this correctly by using `mknod()` to create the file, then `setxattr()` to store the xattrs without opening the file first. That avoids triggering the MDS to assign the layout, which happens at `open()` time.

Yes, good point and thanks for the review, @adilger .  I'll change things to set the xattrs before writing file data, which is how we do it in dcp/dsync.

I had placed it after because I was following the pattern used in libarchive, which seems to write xattrs *after* writing the file data.  I should double check that, but if that's true, this means that the algorithms that use libarchive to create items and write file data will not have restored the Lustre xattrs in time.

adammoody:
I verified that libarchive does copy xattrs *after* copying the file data.  So the algorithms that use libarchive to create items and copy data won't restore Lustre striping parameters.

Potential fix: it may work to configure libarchive to *not* extract xattrs, and then insert code to manually copy xattrs between the call to archive_write_header() and archive_write_data_block().

adammoody:
This is not quite done, but it's close enough to be merged soon.  Removing the work-in-progress label.

PR #432, opened 2020-12-21
daos: abstract to mfu_daos_util

### LINKED ISSUE #391: daos: abstract to daos_util
As we add more DAOS-related code, it would be a good idea to abstract as much as possible to a new `daos_util.[h,c]`.
Since we might be adding DAOS support to `dsync`, it would be a good idea to do this first, so `dsync` can reuse the code.

There are also plans to move the daos_* IO functions internal to DAOS, so we an leave those within `mfu_io` for now.

Some considerations:
- `daos_*` hash and helper functions within `mfu_io.c`
  - Maybe these could be moved to `daos_util.c`
- `daos_*` functions within `mfu_util.c`
  - These could be moved to `daos_util.c`
- `daos_*` functions within `dcp.c`
  - Some of these can be moved to `daos_util.c`
  - Some of these can be re-written in a way that they can be moved to `daos_util.c`
  - Some of the `dcp`-specific DAOS functions should probably remain in `dcp.c`

daltonbohning:
Abstracted daos-specifc functions to new files:
- mfu_daos.h
- mfu_daos.c

cmake will now conditionally include these files.

Note: The daos functions in mfu_io have been left untouched, as this will eventually be moved internal to daos.

closes #391 


daltonbohning:
@adammoody This just puts all the `daos_` functions into their own file. I would like to perhaps change the prefix for the functions from `daos_` to `mfu_daos_`, since internally we are trying to reserve `daos_` for the core DAOS API. But @dsikich is on vacation, so I'd like to avoid obtrusive changes of that nature without discussion. For now, I think that just simply moving these functions is okay, and will allow me to R&D DAOS support in `dsync` (just preliminary at this point).

Question: Which of these do you prefer?
1. CMakeLists.txt conditionally includes `daos_util.[h,c]` when ENABLE_DAOS.
    - This is the option I chose
2. CMakeLists.txt _**always**_  includes `daos_util.[h,c]`, but the entire contents of `daos_util.[h,c]` are wrapped with `#IFDEF DAOS_SUPPORT`

adammoody:
Thanks, @daltonbohning .  Looks good to me.  Yes, I like your approach to just skip even compiling those files when DAOS is not enabled.

daltonbohning:
@adammoody Thanks for the quick review. Is this okay to merge?

adammoody (approved):
Good to merge.

daltonbohning:
Thanks!

PR #434, opened 2020-12-28
daos: improvements

### LINKED ISSUE #433: daos: mfu_param_path_set_all doesn't use daos
`mfu_param_path_set_all` should probably use the `mfu_file_*` operations to support DAOS.

https://github.com/hpc/mpifileutils/blob/063647045d61c9e5583d69bec18e3b9ffea82ae9/src/common/mfu_param_path.c#L632

daltonbohning:
- `dcp`
  - Added a missing daos_cleanup() call
- `mfu_param_path_set` and `mfu_param_path_set_all`
  - Added the argument mfu_file
- Generally changed some functions to use the mfu_file
version, since mfu_file_t was introduced.

Also broke DAOS type into DFS and DAOS

- Changed previous usage of DAOS to DFS
- Added "new" DAOS for non-DFS API
- Added --daos-api to dcp
- Removed is_posix_copy references
  - Use is_daos_copy locally in dcp
- Added compatibility checks for daos containers
- Refactored some DAOS functions
- Fixed memory leaks with setting argpaths

Closes #433 

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -633,6 +633,8 @@ daos_args_t* daos_args_new(void)
     /* By default, try to automatically determine the API */
     da->api = DAOS_API_AUTO;
 
+    da->epc = 0;
+
     return da;
 }
 
@@ -664,6 +666,18 @@ int daos_parse_api_str(
     }
 
     return rc;
+}
+
+int daos_parse_epc_str(
+    const char* epc_str,
+    daos_epoch_t* epc)
+{
+    *epc = strtoull(epc_str, NULL, 10);
+    if (*epc == 0 || (*epc == ULLONG_MAX && errno != 0)) {
+        return 1;
+    }
+
+    return 0;
 }
 
 int daos_setup(
@@ -1165,7 +1179,7 @@ static int daos_obj_copy(
     return rc;
 }
 
-static int daos_obj_list_oids(daos_args_t* da, daos_epoch_t* epoch, mfu_flist bflist) {
+static int daos_obj_list_oids(daos_args_t* da, mfu_flist bflist) {
     /* List objects in src container to be copied to 
      * destination container */
     static const int     OID_ARR_SIZE = 50;
@@ -1177,7 +1191,7 @@ static int daos_obj_list_oids(daos_args_t* da, daos_epoch_t* epoch, mfu_flist bf
     int                  rc = 0;
 
     /* create snapshot to pass to object iterator table */
-    rc = daos_cont_create_snap_opt(da->src_coh, epoch, NULL,
+    rc = daos_cont_create_snap_opt(da->src_coh, &da->epc, NULL,
     				   DAOS_SNAP_OPT_CR | DAOS_SNAP_OPT_OIT,
 				   NULL);
     if (rc != 0) {
@@ -1187,7 +1201,7 @@ static int daos_obj_list_oids(daos_args_t* da, daos_epoch_t* epoch, mfu_flist bf
     }
 
     /* open object iterator table */
-    rc = daos_oit_open(da->src_coh, *epoch, &toh, NULL);
+    rc = daos_oit_open(da->src_coh, da->epc, &toh, NULL);
     if (rc != 0) {
         MFU_LOG(MFU_LOG_ERR, "DAOS failed to open oit: ", MFU_ERRF,
                 MFU_ERRP(-MFU_ERR_DAOS));
@@ -1229,7 +1243,6 @@ static int daos_obj_list_oids(daos_args_t* da, daos_epoch_t* epoch, mfu_flist bf
 
 int mfu_flist_walk_daos(
     daos_args_t* da,
-    daos_epoch_t* epoch,
     mfu_flist flist)
 {
     /* assume we'll succeed */
@@ -1241,7 +1254,7 @@ int mfu_flist_walk_daos(
 
     /* have rank 0 do the work of listing the objects */
     if (rank == 0) {
-        rc = daos_obj_list_oids(da, epoch, flist);
+        rc = daos_obj_list_oids(da, flist);
         if (rc != 0) {
             MFU_LOG(MFU_LOG_ERR, "DAOS failed to list oids: ",
                 MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS));
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -28,8 +28,7 @@ static int daos_check_args(
     int rank,
     char** argpaths,
     daos_args_t* da,
-    int* flag_daos_args,
-    bool* is_posix_copy)
+    int* flag_daos_args)
 {
     char* src_path = argpaths[0];
     char* dst_path = argpaths[1];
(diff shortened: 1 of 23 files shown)


daltonbohning:
@adammoody I would like your input on this before I go any further.
I found myself needing to add an `mfu_file_t` argument to `mfu_param_path_set_all`.
Since many of the utilities call this function, I thought it might be good to make it optional, by passing NULL.
This could also be applied to `mfu_flist_walk_param_paths`.
Basically, the idea is that most utilities do NOT support DAOS, so maybe it's simpler to not have to allocate and free an `mfu_file_t` object _**just**_ so it can be passed to these functions. If the `mfu_file_t` param is passed as NULL, we can just create a default with `mfu_file_new()`.

daltonbohning:
@dsikich I would appreciate your input as well, if you have an opinion.

adammoody:
@daltonbohning , thanks.  This looks good to me.  I'm also fine with updating each tool to allocate a struct, but we can add that later.

However, you're starting to touch on a more general topic that I haven't had time to think about in a lot of detail yet.  Even before @dsikich started to add any DAOS support, there was a need to create a structure like this to record details about the file system at the source and destination paths.

Some of the mfu library functions can be implemented in many different ways, e.g., mfu_flist_unlink() has 5-6 different algorithms.  The fastest algorithm to do a particular task can be different across file systems, and it would be nice to know so that the library could select the right one.  For example, knowing that the source path is on Lustre and that the destination path is on NFS would be useful at times.  Or someone might even have Lustre on both ends, but have different versions of Lustre between the source and destination (e.g., migrating data from an older file system to a newer one).  

So eventually, it would be useful to be able to assign a file system type struct to each path that the user gives.  For example, someone might have a command like:

```dcp /lustre/dir1 /gpfs/dir2 /nfs/dir```

In this case, all three of these paths (2 source and 1 destination) are hosted on different file systems.

Going that route, I can envision that we may want an array of these structs, one for each source path.

We don't need to do this now either, but I wanted to bring it up just to plant the idea.  We'd want to first think that through before taking that step.  Perhaps that could even be buried in the param path struct and copied over to the mfu_file_t or something?

Anyway, we can figure out the more general case down the road.

**Comment by daltonbohning** [comment]
@adammoody Thanks for the input!
 
> Some of the mfu library functions can be implemented in many different ways, e.g., mfu_flist_unlink() has 5-6 different algorithms. The fastest algorithm to do a particular task can be different across file systems... 

I've started working on DAOS support in dsync, and I've come across mfu_flist_unlink() and the different algorithms, so this will soon have the mfu_file_t structure, which is at least a start.

> So eventually, it would be useful to be able to assign a file system type struct to each path that the user gives
> ...
> Going that route, I can envision that we may want an array of these structs, one for each source path.

Definitely. Currently, for DAOS, we are restricted to only one source for this very reason.

> We don't need to do this now either, but I wanted to bring it up just to plant the idea. We'd want to first think that through before taking that step. Perhaps that could even be buried in the param path struct and copied over to the mfu_file_t or something?

I'll need to think on this as well. I'll keep this in mind as I work through these changes, DAOS dsync support, etc.; although, I like the general idea of burying the mfu_file_t struct in an existing struct if possible, as that reduces the likelihood of non-backwards-compatible changes whenever a new function needs mfu_file_t. (since most of the code base is a public library)

In general, it seems that the mfu_file_t structure is a step toward future goals, so I think I'll go ahead and allocate the mfu_file_t structure in each tool that uses these functions, and I'll remove the "optional" NULL check for the mfu_file parameter, to avoid clutter. This is quite simple to do.

**Comment by daltonbohning** [comment]
@adammoody Okay, I added `mfu_file_t` to the corresponding tools that use the functions with `mfu_file_t` arguments.
Since `mfu_file_t` was introduced to these tools, I also updated some IO functions to use the `mfu_file_*` versions, to be consistent with usage. E.g. `mfu_stat` -> `mfu_file_stat`. I didn't update very instance in every tool, but I think I did get most of the ones in the `main` functions.
Once I have [DAOS support added to dsync](https://github.com/daos-stack/mpifileutils/tree/dbohninx-daos-dsync), even more of these such instances should be updated.

File: src/dcp/dcp.c [hunk 1]
```diff
@@ -518,8 +525,8 @@ int main(int argc, char** argv)
         /* destroy snapshot after copy */
         if (rank == 0) {
             daos_epoch_range_t epr;
-            epr.epr_lo = epoch;
-            epr.epr_hi = epoch;
+            epr.epr_lo = daos_args->epc;
+            epr.epr_hi = daos_args->epc;
```
dsikich, line 529: @daltonbohning this was my mistake, but we should add an if here to only destroy if the snapshot wasn't passed in by the user. I had to go back and add this to the daos utility myself after you brought it up.

dsikich:
@daltonbohning I suppose I am wondering why the AUTO enum is necessary? Would someone need to explicitly pass in --daos-api=AUTO if that is already the default behavior?

daltonbohning:
> @daltonbohning I suppose I am wondering why the AUTO enum is necessary? Would someone need to explicitly pass in --daos-api=AUTO if that is already the default behavior?

@dsikich I think it is necessary in the `enum` itself, but not at the command line. Not passing in `--daos-api` at all really is just AUTO, so I'll remove that from the command line so it's not confusing.

src/dcp/dcp.c
```diff
@@ -66,6 +66,11 @@ void print_usage(void)
     printf("\n");
     printf("Usage: dcp [options] source target\n");
     printf("       dcp [options] source ... target_dir\n");
+#ifdef DAOS_SUPPORT
+    printf("\n");
+    printf("DAOS paths can be specified as:\n");
+    printf("       daos://<pool>/<cont>[/<path> | <UNS path>\n");
```
dsikich, line 72: @daltonbohning I think there is a missing closing bracket in here.

src/dcp/dcp.c
```diff
@@ -429,7 +412,7 @@ int main(int argc, char** argv)
         numpaths_src = numpaths - 1;
     }
 
-    if (numpaths_src == 0 && is_posix_copy) {
+    if (numpaths_src == 0 && !is_daos_copy) {
```
dsikich, line 415: @daltonbohning if it is a DAOS copy, and we are using DFS, could numpaths_src here be zero? Would that be an error?

daltonbohning: @dsikich Oh yes, seg faults. Now that the obj-level copy uses the "paths" as well, we can move that numpaths check back to BEFORE daos_setup() and simplify this a bit.

daltonbohning: @dsikich I actually found a few things that could be simplified, in my latest commit

adammoody:
@daltonbohning , I just merged the dtar branch, which likely generated some merge conflicts for you.  Hopefully that's not too much trouble.

adammoody:
@daltonbohning and @dsikich , oh also this merge of dtar now requires libarchive 3.5.0.  You'll probably have to update that.  I can pass along the steps I use if it helps.

daltonbohning:
> @daltonbohning , I just merged the dtar branch, which likely generated some merge conflicts for you. Hopefully that's not too much trouble.

@adammoody Great! The conflicts shouldn't be too difficult, and I saw it coming anyway :)

daltonbohning:
> @daltonbohning and @dsikich , oh also this merge of dtar now requires libarchive 3.5.0. You'll probably have to update that. I can pass along the steps I use if it helps.

@adammoody Oh, that might be fun... Internally, we have a `.spec` to build mpifileutils, so hopefully I can just specify version 3.5.0. I'll take a look at this on Monday. Thanks for the heads up!

daltonbohning:
@adammoody I rebased this on master and I'm at a point where the changes are stable.

dsikich (approved):

src/common/mfu_daos.c
```diff
@@ -121,118 +104,285 @@ static bool daos_check_prefix(char* path, const char* dfs_prefix, char** suffix)
...
+    char* path,
+    uuid_t* p_uuid,
+    uuid_t* c_uuid)
+{
+    struct duns_attr_t  dattr = {0};
+    size_t              path_len = strlen(path);
+    int                 rc;
+
+    rc = duns_resolve_path(path, &dattr);
+    if (rc == 0) {
```
dsikich, line 135: @daltonbohning if the user is not using a UNS path, wouldn't duns_resolve_path always return a non-zero rc? If you are using the rel_path passed back by dattr then I guess I am confused how you are retrieving the rel_path for non-UNS paths.

daltonbohning: @dsikich You're right, `rc != 0` for "regular POSIX" paths. In that case, just don't do anything. Ignore `dattr` entirely for `rc != 0`, don't modify the path, don't use `rel_path`, don't set pool/cont uuids, etc. Then, this "POSIX" path will be handled normally by `dcp` and and the other internal functions.

The way I'm using `duns_resolve_path` here is basically as an "interceptor". I call `duns_resolve_path` on the path passed in. This means that `rc == 0` for daos:// and UNS paths, in which case we DO use rel_path and set the pool/cont uuids.

An improvement that could be made here is something like:
```
if (rc != 0 && strcnmp(path, "daos://", 8) == 0) {
    // the user explicitly specified a daos path, but it was somehow invalid
    // do NOT treat this as a POSIX path
}
```
I will add this.

dsikich: @daltonbohning ok, I see. I was looking at how you parsed the paths here, because I don't think my other PR for the DAOS filesystem copy is going to get through unless I update using this function as well (instead of all the manual parsing I'm using). I am planning to use how you re-worked it here. It should simplify and remove a lot of the comments made on the PR I'm trying to get through.

daltonbohning: @dsikich Sounds good! There are definitely some more improvements that can be made here with this parsing, but I think this is at least stable. My latest commit also includes another error check.

adammoody:
@daltonbohning , this looks good to me.  Thanks!

adammoody (approved):

PR #435, opened 2021-01-15
doc: update build steps to use libarchive-3.5.0

adammoody:

### DIFF
--- a/doc/rst/build.rst
+++ b/doc/rst/build.rst
@@ -46,8 +46,9 @@ Then to install the dependencies, run the following commands:
    mkdir deps
    cd deps
      wget https://github.com/hpc/libcircle/releases/download/v0.3/libcircle-0.3.0.tar.gz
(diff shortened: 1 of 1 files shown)


daltonbohning:
@adammoody libarchive-3.5.1 was released on December 25, 2020. It includes some compile fixes, for an issue I am running into on my system. Would it be okay if we bump libarchive to 3.5.1 instead of 3.5.0? I can submit a PR for this

daltonbohning:
> @adammoody libarchive-3.5.1 was released on December 25, 2020. It includes some compile fixes, for an issue I am running into on my system. Would it be okay if we bump libarchive to 3.5.1 instead of 3.5.0? I can submit a PR for this

Edit: It does seem that FindLibArchive.cmake will accept version 3.5.0, so I guess it's not necessary to require 3.5.1 or higher, but it would still reduce the likelihood of someone else trying to use 3.5.0 and running into similar compile issues.
I created #436 for this

PR #436, opened 2021-01-18
libarchive: update to version 3.5.1

daltonbohning:
libarchive 3.5.1 fixes some compile issues from 3.5.0.
I suggest using 3.5.1 in the documentation, etc. so other users don't run into compile issues that are fixed in 3.5.1.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,7 +84,7 @@ LIST(APPEND MFU_EXTERNAL_LIBS ${DTCMP_LIBRARIES})
 # libarchive 3.1.2 is available on some systems,
 # but pick a newer libarchive to avoid bug with files starting with "._",
 # which is misinterpretted as a MacOS extension on Linuxlibarchive 3.1.2
-FIND_PACKAGE(LibArchive 3.5.0 REQUIRED)
+FIND_PACKAGE(LibArchive 3.5.1 REQUIRED)
 INCLUDE_DIRECTORIES(${LibArchive_INCLUDE_DIRS})
 LIST(APPEND MFU_EXTERNAL_LIBS ${LibArchive_LIBRARIES})
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,8 @@ before_install:
  - pushd ${dtcmp}
  - ./configure --with-lwgrp=/usr/local && make && sudo make install
  - popd
- - libarchive=libarchive-3.5.0
- - wget https://github.com/libarchive/libarchive/releases/download/v3.5.0/${libarchive}.tar.gz
+ - libarchive=libarchive-3.5.1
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ LIST(APPEND MFU_EXTERNAL_LIBS ${DTCMP_LIBRARIES})
 # libarchive 3.1.2 is available on some systems,
 # but pick a newer libarchive to avoid bug with files starting with "._",
 # which is misinterpretted as a MacOS extension on Linuxlibarchive 3.1.2
-FIND_PACKAGE(LibArchive 3.5.0 REQUIRED)
+FIND_PACKAGE(LibArchive 3.5.1 REQUIRED)
```
daltonbohning, line 87: Something we could do here is just keep this at 3.5.0 to be compatible, but just use 3.5.1 in the travis build and documentation.

adammoody:
Thanks, @daltonbohning .  Let's go ahead and bump everything to 3.5.1.

Would you mind updating travis and the docs to use 3.5.1 as part of this PR?

daltonbohning:
> Thanks, @daltonbohning . Let's go ahead and bump everything to 3.5.1.
> 
> Would you mind updating travis and the docs to use 3.5.1 as part of this PR?

@adammoody I believe this PR does contain all of the necessary changes for 3.5.1
- travis
- cmake "find"
- build docs

Also, it seems that:
https://github.com/hpc/mpifileutils/blob/a7afd70adc8807b5ce180823d5847497e41db9a6/CMakeLists.txt#L87
Does not enforce the version.
From https://cmake.org/cmake/help/latest/command/find_package.html
"CMake searches for a file called `Find<PackageName>.cmake` ... It is responsible for finding the package, checking the version, and producing any needed messages."

For example, I did this locally:
`FIND_PACKAGE(LibArchive 3.5.0 REQUIRED)`
And cmake accepted it:
`-- Found LibArchive: <path>/libarchive.so (Required is at least version "4.0")`

I'm not familiar enough with cmake to know how to enforce the version in FindLibarchive.cmake, but maybe this is okay, since the documentation directs users to use 3.5.1

daltonbohning:
@adammoody Also, for what it's worth, `yum install libarchive` doesn't have an rpm for 3.5.0+ on all distributions, so _**not**_ enforcing the version could actually be beneficial to users who don't _**need**_ the improvements, because it means that `yum install libarchive`, etc. (which gives version < 3.5 on some systems) would be sufficient for a compile.

This would actually be beneficial to us for DAOS, since we aren't currently using any utilities that absolutely require this updated version.

adammoody:
Oh, right.  I skimmed things too quickly the first time through.  And I also agree with your point about not making the newer libarchive be a strict requirement.  This all looks good to merge.  Thanks again.

daltonbohning:
No worries. Thanks!

PR #439, opened 2021-01-20
daos: dsync and dcmp support

daltonbohning:
Added daos support to dsync.
Added daos support to dcmp.

`dsync.c`
- Added daos options
- Modified IO calls to use `mfu_file_*`

`dcmp.c`
-- Added daos options

`mfu_io`
- Added `mfu_file_rmdir` and `daos_rmdir`

`mfu_flist_unlink`
- Added `mfu_file_t` param

`mfu_daos.c`
- Fixed error checking for DAOS args

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -47,9 +47,14 @@ static int daos_check_args(
     if (have_src_pool || have_src_cont || have_dst_pool || have_dst_cont
             || have_prefix) {
         *flag_daos_args = 1;
-    } else {
+    }
+    if ((have_src_path && (strncmp(src_path, "daos:", 5) == 0)) ||
+            (have_dst_path && (strncmp(dst_path, "daos:", 5) == 0))) {
+        *flag_daos_args = 1;
+    }
+    if (*flag_daos_args == 0) {
         return 0;
-    } 
+    }
     
     /* Determine whether the source and destination
      * use the same pool and container */
@@ -143,7 +148,7 @@ static int daos_parse_path(
         } else {
             strcpy(path, dattr.da_rel_path);
         }
-    } else if (strncmp(path, "daos://", 7) == 0) {
+    } else if (strncmp(path, "daos:", 5) == 0) {
         /* Actual error, since we expect a daos path */
         rc = -1;
     } else {
@@ -246,7 +251,7 @@ static int daos_set_paths(
         if (src_rc == 0) {
             argpaths[0] = da->src_path = strdup(src_path);
         } else if (src_rc == -1) {
-            MFU_LOG(MFU_LOG_ERR, "Failed to parse DAOS source path.");
+            MFU_LOG(MFU_LOG_ERR, "Failed to parse DAOS source path: daos://<pool>/<cont>[/<path>]");
             return 1;
         }
     }
@@ -257,7 +262,7 @@ static int daos_set_paths(
         if (dst_rc == 0) {
             argpaths[1] = da->dst_path = strdup(dst_path);
         } else if (dst_rc == -1) {
-            MFU_LOG(MFU_LOG_ERR, "Failed to parse DAOS destination path.");
+            MFU_LOG(MFU_LOG_ERR, "Failed to parse DAOS destination path: daos://<pool>/<cont>[/<path>]");
             return 1;
         }
     }
@@ -701,6 +706,16 @@ int daos_setup(
     if (tmp_rc != 0) {
         MFU_LOG(MFU_LOG_ERR, "Failed to initialize daos");
         local_daos_error = true;
+    }
+
+    /* Do a preliminary check on the DAOS args */
+    if (!local_daos_error) {
+        tmp_rc = daos_check_args(rank, argpaths, da, &flag_daos_args);
+        if (tmp_rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "Invalid DAOS args: "
+                    MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS_INVAL_ARG));
+            local_daos_error = true;
+        }
     }
 
     /* Figure out if daos path is the src or dst,
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -602,7 +602,12 @@ int daos_utimensat(int dirfd, const char* pathname, const struct timespec times[
     dfs_obj_t* parent = daos_hash_lookup(dir_name, mfu_file);
     if (parent == NULL) {
         rc = -1;
-    } else {
+    } else if (name) {
+        /* TODO DAOS properly handle when name is NULL. I.e. root /
+         * For now, just skip if root, since a proper fix needs to be
+         * done in several functions, and these functions will soon be moved
+         * internal to DAOS anyway, in which case they will be refactored and fixed. */
+
         /* Lookup the object */
         dfs_obj_t* obj;
         int lookup_flags = O_RDWR;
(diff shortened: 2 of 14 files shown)


daltonbohning:
This is fairly stable.
It seems to work for basic test cases with DAOS, and it doesn't seem to have broken the existing "posix" cases.

src/dsync/dsync.c
```diff
@@ -1305,9 +1325,17 @@ static int dsync_sync_files(
...
             MFU_LOG(MFU_LOG_ERR, "Exiting run");
         }
-        mfu_finalize();
-        MPI_Finalize();
-        return 0;
+         /* TODO this is problematic for a few reasons:
+         * 1. The calling function doesn't see this as an error.
+         * 2. The calling function still tries to use MPI.
+         * 3. Resources are not cleaned up. */
+        //mfu_finalize();
```
dsikich, line 1332: @daltonbohning did you mean to leave in the commented out functions? So, it looks like before that if MPI_Finalize is called are you saying that it still tries to use MPI afterward?

src/dsync/dsync.c
```diff
@@ -3030,16 +3085,52 @@ int main(int argc, char **argv)
...
+    if (options.link_dest != NULL &&
+            (mfu_src_file->type == DAOS || mfu_src_file->type == DFS) &&
+            (mfu_dst_file->type == DAOS || mfu_dst_file->type == DFS)) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS does not support hard links.");
+        daos_do_cleanup = true;
+        daos_do_exit = true;
+    }
+
+    /* Not yet supported */
+    if (mfu_src_file->type == DAOS || mfu_dst_file->type == DAOS) {
```
dsikich, line 3115: @daltonbohning if someone is using --daos-api=DAOS even though they pass in two POSIX containers it looks like this will fail. I think that is ok, but the error message might be kind of confusing in that case.

src/common/mfu_daos.c
```diff
@@ -143,7 +148,7 @@ static int daos_parse_path(
         } else {
             strcpy(path, dattr.da_rel_path);
         }
-    } else if (strncmp(path, "daos://", 7) == 0) {
+    } else if (strncmp(path, "daos:", 5) == 0) {
```
dsikich, line 151: @daltonbohning why don't we check the whole daos prefix here? Just wondering really since it looks like you had it before.

src/dsync/dsync.c
daltonbohning: @dsikich Yeah, really I was hoping Adam might have some input here. The problem is that this is not in main, so finalizing shouldn't be done here. "return 0" just returns from the function, which still expects the MPI context to be initialized.

src/dsync/dsync.c
daltonbohning: @dsikich Hmm... I see what you mean. I updated this.

src/common/mfu_daos.c
daltonbohning: @dsikich I was trying to handle the case where maybe the user just type the format incorrectly.
For example:
`daos:/pool/cont`
In this case, it's safe to say that the "daos:" prefix meant that the user wants this to be DAOS, so we can error and alert them.

It's probably incredibly unlikely that an _**actual**_ file/dir would start with "daos:".

What do you think?

dsikich: @daltonbohning yeah probably unlikely especially with the check for the colon added. In that case it might help to print out a message to the user reminding them of the correct format.

dsikich: @daltonbohning I can see people making this mistake or missing the second forward slash.

src/dsync/dsync.c
dsikich: @daltonbohning @adammoody ok, yeah this is kind of weird.

dsikich:
@daltonbohning I think this looks pretty good. I will hold off approving as I would like to hear back about where the resources are not cleaned up in dsync, and if we can do something better for that case.

src/common/mfu_daos.c
daltonbohning: @dsikich I added a more descriptive message.

src/dsync/dsync.c
daltonbohning: I'm hoping that maybe the "return -1" I added is sufficient, though really it seems that this is a "fatal" error, so the caller should also immediately return, and the caller's caller also... which is what makes this complex.
So really, if we can turn this into a "log error and continue" instead of "exit immediately", the "return -1" should be okay.

dsikich: @daltonbohning yes I think it should be ok, but we will probably want to delete the commented out functions if we decide to do it that way.

dsikich:
> @daltonbohning I think this looks pretty good. I will hold off approving as I would like to hear back about where the resources are not cleaned up in dsync, and if we can do something better for that case.

I am going to approve this, but I will let @adammoody comment if he wants something updated for the dsync failure case.

dsikich (approved):

src/common/mfu_io.c
```diff
@@ -2080,6 +2171,78 @@ int mfu_rmdir(const char* dir)
...
+        rc = -1;
+        goto out;
+    }
+
+    /* Lookup the parent directory */
+    dfs_obj_t* parent = daos_hash_lookup(dir_name, mfu_file);
+    if (parent == NULL) {
+        rc = -1;
+        goto out;
+    }
```
adammoody, line 2196:
```suggestion
    /* Lookup the parent directory */
    dfs_obj_t* parent = daos_hash_lookup(dir_name, mfu_file);
    if (parent == NULL) {
        // errno is set by daos_hash_lookup, right?  If so, let's add a comment here, since I'll keep forgetting
        rc = -1;
        goto out;
    }
```

src/dsync/dsync.c
adammoody: That "exit when invalid" code probably came in from a cut-and-paste when moving this logic from the main function of the dsync tool into a function.  Yes, we should just have the function return an error in this case.  Thanks @daltonbohning and @dsikich

adammoody:
This looks good.  Picking up dsync and dcmp support in DAOS is a nice addition.

adammoody (approved):

src/dsync/dsync.c
adammoody: The goal was to eventually move this dsync_sync_files function into libmfu as mfu_flist_sync that other tools could call.  I think it needed more work, and that I stopped half way after pulling code from the dsync main() function into separate functions.

Looking at dsync_sync_files a bit closer, let's ensure that it returns the same value on all ranks (global return value).  I think we can just have this part return -1, assuming all ranks get ``!valid`` at the same time.  I don't remember off hand about other points where we set ``rc = -1`` further down in that function.  To be safe, we could add a ``mfu_alltrue`` call before returning.

Then in the caller, my first reaction was that we should avoid setting metadata on failure.  However, now I'm not sure.  I could see that we might get some errors when sync'ing files, because perhaps someone deleted a file while the tool was running.  In that case, they probably would want to go ahead and set metadata on things.

daltonbohning: @adammoody `dsync_sync_files` sets `rc = -1` for these cases:
- `mfu_flist_hardlink` - safe - Performs MPI_Allreduce before returning
- `mfu_flist_copy` - safe - Performs MPI_Allreduce before returning

And for `mfu_param_path_check_copy` (the one in question here), rank 0 performs a bcast into `valid`.
So I think the return from `dsync_sync_files` should be the same for all processes.

For now, I kept the `return -1` for `mfu_param_path_check_copy`, and kept a TODO comment in place so we can keep track of the potential improvement here.

As for setting metadata, I agree. There are some cases where maybe we should and some cases where maybe we should not. For this particular case - `mfu_param_path_check_copy` error, it's probably okay to NOT set the metadata, since it's possible that the source or destination path itself is invalid. But I think this would require more careful analysis, and possibly some refactoring as you suggested.

src/common/mfu_io.c
daltonbohning: @adammoody Yes, daos_hash_lookup does set errno. But... most of the daos_ functions call daos_hash_lookup, so I would need to add a similar comment to each function (~20 of them). We are actually working on moving these daos_ IO functions internal to DAOS, in which case they will RETURN an errno code instead of setting errno, so it will be more more explicit.
If it's okay with you, I'd like to just let this be for now.

daltonbohning:
> This looks good. Picking up dsync and dcmp support in DAOS is a nice addition.

Definitely, and thanks @adammoody ! I didn't make any substantial changes in my last commit - just cleaned up the comment in dsync_sync_files. If you're okay with this, I'll go ahead and merge.

src/common/mfu_io.c
adammoody: Sure.  Sounds good.

adammoody:
Yep, good to merge.  Thanks!

PR #442, opened 2021-01-21
libarchive: conditional compile

daltonbohning:
Not all systems have a libarchive-3.5* package available
through yum, apt-get, dnf, etc.

To avoid the need to build libarchive from source when it
is not needed, add the optional cmake flag ENABLE_LIBARCHIVE.
By default, this is turned on.
When turned off, mfu will not link against libarchive, nor will
it build dtar, which requires libarchive.


daltonbohning:
@adammoody Internally, for DAOS, we build mpifileutils with our own .spec to build against multiple distributions. But, for example, the latest _**public**_ rpm available for LibArchive on CentOS 7 is only v3.1.2. To avoid needing to package libarchive ourselves for v3.5+, for multiple distributions, it would be helpful if we could disable libarchive altogether, since we don't currently have DAOS support in dtar anyway.
In the future, if we add DAOS support to dtar, then I guess we would have to package libarchive ourselves, but for now this is the simplest route for us.

Are you okay with these changes? It doesn't modify the **_default_** build behavior at all.

doc/rst/build.rst
```diff
@@ -134,6 +134,12 @@ To enable Lustre, GPFS, and experimental tools, add the following flags during C
     -DENABLE_GPFS=ON
     -DENABLE_EXPERIMENTAL=ON
 
+To disable linking against LibArchive, and tools requiring LibArchive, add the following flag during CMake:
```
adammoody, line 137:
```suggestion
To disable linking against libarchive, and tools requiring libarchive, add the following flag during CMake:
```

adammoody (approved):
Looks good.  Thanks, @daltonbohning

daltonbohning:
> Looks good. Thanks, @daltonbohning

Thanks @adammoody ! I'll merge once the Travis build completes successfully.

PR #447, opened 2021-01-26
bump shared library version to 3.0.0

daltonbohning:
Bump shared library version to 3.0.0 due to previous commits.


daltonbohning:
@adammoody I merged a previous commit that probably should have bumped the shared library version.
It's probably not a huge deal for individual commits, but we'll probably want to merge this change right before the next release so an installation from a release has a "proper" shared library version.

adammoody:
Thanks @daltonbohning .  Yeah, I also made some changes that would require a so version bump.  So that we don't have to bump the number of every commit that changes the interface, we could aim to only bump things on actual releases.  That would be more manageable for us.  It would mean people might be disrupted if they are developing tools against the development branch, but that's probably workable so long as we don't go too long between releases.

PR #454, opened 2021-02-02
Abstract logic for printing mfu_copy_stats

daltonbohning:
Moved the prints for mfu_copy_stats into a new function,
that allows printing in 3 different versions, for each
of the versions that previously existed.
Most of the logic is the same, where just the output varies slightly.

Also adjusted the allreduce to be in place.


daltonbohning:
@adammoody I'm working on adding metrics for "DAOS objects", which will require some conditional compilation and checks to know whether to print "objects" or "files", etc.
This centralizes the logic for printing the stats, which will reduce the code footprint when I add stats for DAOS objects.

daltonbohning:
Actually, on second thought, I realize now that this won't work the way I need it to...
Closing this PR until it has been reworked.
Sorry about that.

PR #455, opened 2021-02-03
daos: dsync support for daos objects

daltonbohning:
Adds support for DAOS objects in dsync.
Functionally, this should work, although these points will be tackled at a later time:
- General performance improvements
- Updated/improved metrics for DAOS objects

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -18,6 +18,107 @@
  * Private definitions.
  */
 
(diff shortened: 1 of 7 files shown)


daltonbohning:
@dsikich This is still a little rough. I need to clean some things up and I have a few TODO items. If you get the chance, I would appreciate a quick review in case you spot something.

PR #456, opened 2021-02-03
fix compile warnings

daltonbohning:
Fixed some compile warnings for shadowed variable declarations.


adammoody (approved):
Thanks, @daltonbohning

Issue #10, opened 2013-12-06
Move dtar to non-experimental


bringhurst:
I haven't really looked into it, but is there any reason why dtar is in experimental?

gonsie:
This was discussed further in #84. There are still a few issues preventing dtar from being released.

PR #457, opened 2021-02-04
daos: add daos-serialize and daos-deserialize tools

dsikich:
* add optional HDF5_SUPPORT to build
    * add helper functions to mfu_daos
    * daos-serialize serializes daos containers into hdf5
      files, for long term storage on a POSIX filesystem.
    * daos-deserialize deserializes an hdf5 file by
      restoring it back into a new daos container

daos-deserialize can only be used when daos-serialize
was the tool used to create the hdf5 file. The serialize
tool evenly distributes the containers object ids among 
ranks, and each rank then writes to its own HDF5 file.
The deserialize tool evenly distributes the HDF5 files
among ranks. 

examples:
mpirun -np 2 daos-serialize -v /pool/cont 
mpirun -np 2 daos-deserialize -v --pool <pool> h5file h5file ...

TODO: add obj level statistics after dsync obj level support PR is merged
TODO: investigate using libcircle for distribution of files among ranks in
daos-deserialize


dsikich:
@daltonbohning @adammoody still a WIP, I need to do more testing to make sure this is properly integrated. I also still have some debugging statements in there until after I test some more. We can review this Friday, as it should be mostly there in terms of organization and code.

src/daos-serialize/daos-serialize.c
```diff
@@ -0,0 +1,226 @@
...
+    mfu_flist_summarize(tmplist);
+    mfu_flist newflist = mfu_flist_spread(tmplist);
+
+    /* get size of local list for each rank */
+    uint64_t size = mfu_flist_size(newflist);
+
+    uuid_t pool_uuids[size];
+    uuid_t cont_uuids[size];
+    daos_handle_t pohs[size];
+    daos_handle_t cohs[size];
```
adammoody, line 159: Since ``size`` is not known until runtime, I think these arrays will need to be MFU_MALLOC'd and mfu_free'd

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+    }
+out:
+    if (buf != NULL)
+        mfu_free(&buf);
+    for (j = 0; j < num_attrs; j++) {
+        mfu_free(&val_buf[j]);
+    }
+    if (val_buf != NULL) {
+        mfu_free(&val_buf);
+    }
```
adammoody, line 2377: It's valid to call ``mfu_free`` even when ``val_buf == NULL``, so you can drop the if check on these to simplify like so:
```suggestion
    mfu_free(&val_buf);
```

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+                rc);
+        goto out;
+    }
+
+    /* skip serializing attrs if total_size is zero */
+    if (total_size == 0) {
+        goto out;
+    }
+
+    buf = malloc(total_size);
```
adammoody, line 2244: This goes back to how much we want to make this code look like the rest of the mfu code, but we could switch to use the MFU_MALLOC wrappers like so:
```suggestion
    buf = MFU_MALLOC(total_size);
```
The MFU_MALLOC wrappers always return NULL when given a ``size==0``, so that we know by looking at the pointer value as to whether it is pointing to memory or not.

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+    uint32_t                i;
+    char                    cont_str[64];
+    int                     len = 0;
+
+    /*
+     * Get all props except the ACL first.
+     */
+    prop_query = daos_prop_alloc(1);
+    if (prop_query == NULL)
+        return ENOMEM;
```
adammoody, line 2569: Another mfu style thing is to always use braces, even for one-liners.
```suggestion
    if (prop_query == NULL) {
        return ENOMEM;
    }
```

adammoody:
I have some styling tools that can help make quick work of things like adding braces and converting tabs to spaces.  Let me dig those up and get back to you before you go do all of that by hand.

adammoody:
Thinking ahead a bit, does DAOS provide ways to query for object ids in parallel rather than have a single process have to get the whole list?

How about for dkeys/akeys?

If so, it would be easier and more efficient to distribute lists across ranks.

daltonbohning:
> Thinking ahead a bit, does DAOS provide ways to query for object ids in parallel rather than have a single process have to get the whole list?
> 
> How about for dkeys/akeys?
> 
> If so, it would be easier and more efficient to distribute lists across ranks.

daos has functionality to "split" an anchor, which allows iterating over the oids, dkeys, and akeys in parallel.
We plan to add this at some point, including for `dcp` and `dsync`.

adammoody:
For styling, we can get pretty far with astyle.  Here are the steps to build the linux option here:

https://sourceforge.net/projects/astyle/files/astyle/astyle%203.1/

```
tar -zxf astyle_3.1_linux.tar.gz
cd astyle
cd build/gcc
make
```

If you're using this on Windows or MacOS, you may need to use different build instructions.

Once you have an astyle binary, you do things like the following:
```
# convert any tabs to spaces, using 4 spaces for indentation
astyle --indent=spaces=4 src/common/mfu_daos.c

# add braces around one-liners
astyle --add-braces src/common/mfu_daos.c
```

You can use redirection in case you want to preserve the original file:
http://astyle.sourceforge.net/astyle.html#_Usage

Documentation for other options:
http://astyle.sourceforge.net/astyle.html

We have an astyle file listed in the project, but it's a been a while since I used it.  The option names may have changed.  Also it may end up changing more than you want, so it's best to use once up front when bringing in a new source file and then tweaking by hand.
https://github.com/hpc/mpifileutils/blob/master/astyle.options

You could pass this to astyle with ``astyle --options=astyle.options file.c``
http://astyle.sourceforge.net/astyle.html#_options

src/common/mfu_daos.c
daltonbohning: I agree that we should use the MFU_MALLOC wrappers when possible, because they will log the file and line number for a failed malloc.

daltonbohning:
> For styling, we can get pretty far with astyle. Here are the steps to build the linux option here:
> 
> https://sourceforge.net/projects/astyle/files/astyle/astyle%203.1/
> ...

Thanks @adammoody ! It might be possible for us to create a githook to run this on changed files.
Or maybe even as a single "cleanup" PR we could standardize all files at once (assuming the diffs are reasonably small) and then a githook on changed files would not be too intrusive.
I'll take a look at this when I get some free time.

src/daos-deserialize/daos-deserialize.c
```diff
@@ -0,0 +1,225 @@
...
+#include <ctype.h>
+#include <inttypes.h>
+#include <unistd.h>
+#include <errno.h>
+#include <stdbool.h>
+
+/* for daos */
+#ifdef DAOS_SUPPORT
+#include "mfu_daos.h"
+#endif
```
daltonbohning, line 14: @dsikich Since the tool has `daos` in the name, I wonder if we should just remove the `#ifdef DAOS_SUPPORT`?

src/CMakeLists.txt
```diff
@@ -15,6 +15,10 @@ ADD_SUBDIRECTORY(dreln)
 ADD_SUBDIRECTORY(drm)
 ADD_SUBDIRECTORY(dstripe)
 ADD_SUBDIRECTORY(dsync)
+IF(ENABLE_HDF5)
+  ADD_SUBDIRECTORY(daos-serialize)
+  ADD_SUBDIRECTORY(daos-deserialize)
+ENDIF(ENABLE_HDF5)
```
daltonbohning, line 21: @dsikich Maybe we should also make sure ENABLE_DAOS is turned on before adding these tools, just in case a user tries to compile with HDF5 support but not DAOS support.

src/common/mfu_daos.h
```diff
@@ -37,6 +41,84 @@ typedef struct {
     daos_epoch_t epc;      /* src container epoch */
 } daos_args_t;
 
+#ifdef HDF5_SUPPORT
+/* for oid dataset */
+typedef struct {
+	uint64_t oid_hi;
+	uint64_t oid_low;
+	uint64_t dkey_offset;
+} oid_t;
```
daltonbohning, line 50: @dsikich In the future, we may consider a namespace for these new types/functions to avoid any potential conflicts down the road.

src/daos-serialize/daos-serialize.c
```diff
@@ -0,0 +1,226 @@
...
+#include <ctype.h>
+#include <inttypes.h>
+#include <unistd.h>
+#include <errno.h>
+#include <stdbool.h>
+
+/* for daos */
+#ifdef DAOS_SUPPORT
+#include "mfu_daos.h"
+#endif
```
daltonbohning, line 14: @dsikich Similar to `daos-serialize`, maybe we don't need the `#ifdef DAOS_SUPPORT`

src/daos-serialize/daos-serialize.c
```diff
@@ -0,0 +1,226 @@
...
+    printf("\n");
+    printf("Usage: daos-serialize [options] /<pool>/<cont> /<pool>/<cont> ... \n");
+    printf("\n");
+    printf("DAOS paths can be specified as:\n");
+    printf("       /<pool>/<cont> | <UNS path>\n");
+    printf("\n");
+    printf("Options:\n");
+    printf("  -v, --verbose            - verbose output\n");
+    printf("  -h, --help               - print usage\n");
+    printf("For more information see https://mpifileutils.readthedocs.io.\n");
```
daltonbohning, line 32: @dsikich This reminds me that we'll probably want to update the DAOS readme and the man pages

src/daos-serialize/daos-serialize.c
```diff
@@ -0,0 +1,226 @@
...
+    /* initialize MPI */
+    MPI_Init(&argc, &argv);
+    mfu_init();
+
+    /* get our rank */
+    int rank;
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+    /* verbose by default */
+    mfu_debug_level = MFU_LOG_VERBOSE;
```
daltonbohning, line 51: @dsikich Since verbose is turned on by default, maybe we should also add a quite option:
```
case 'q':
                mfu_debug_level = MFU_LOG_NONE;
```

src/daos-serialize/daos-serialize.c
```diff
@@ -0,0 +1,226 @@
...
+     */
+
+    /* create an empty file list */
+    mfu_flist tmplist = mfu_flist_new();
+
+    rc = daos_init();
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to initialize daos");
+        return 1;
+    }
```
daltonbohning, line 137: @dsikich We need to make sure we finalize mfu and mpi here

src/daos-serialize/daos-serialize.c
daltonbohning, line 159: @dsikich In the future, we can consider adjust the existing `daos_args` to hold multiple pools/conts. It would take some refactoring of the existing functions, but it would also allow us to support multiple inputs for `dcp`, and would generally centralize the code.

src/daos-serialize/daos-serialize.c
```diff
@@ -0,0 +1,226 @@
...
+            MFU_LOG(MFU_LOG_ERR, "Failed to connect to DAOS");
+            rc = 1;
+        }
+
+        /* serialize this pool/cont to an hdf5 file */
+        rc = cont_serialize_hdlr(pool_uuids[i], cont_uuids[i],
+                                 pohs[i], cohs[i]);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "Failed to serialize container (%d)", rc);
+        }
```
daltonbohning, line 189: @dsikich The `rc` from here is clobbered by the next functions. (See below comments)

src/daos-serialize/daos-serialize.c
```diff
@@ -0,0 +1,226 @@
...
+            MFU_LOG(MFU_LOG_ERR, "Failed to close container (%d)", rc);
+            rc = 1;
+        }
+
+        rc = daos_pool_disconnect(pohs[i], NULL);
+        MPI_Barrier(MPI_COMM_WORLD);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "Failed to disconnect pool (%d)", rc);
+            rc = 1;
+        }
```
daltonbohning, line 203: @dsikich We should maybe do something like:
```
tmp_rc = ...
if (tmp_rc != 0) {
    rc = 1;
}
```

src/daos-serialize/daos-serialize.c
```diff
@@ -0,0 +1,226 @@
...
+    }
+    
+    /* Alert the user if there were copy errors */
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "One or more errors were detected while "
+                "serializing: " MFU_ERRF, MFU_ERRP(MFU_ERR_DAOS));
+    }
+
+    rc = daos_fini();
```
daltonbohning, line 212: @dsikich Maybe use `tmp_rc` here as well. Also, if we're going to set `rc = 1` when this fails, we should maybe also log a message. Otherwise, it might be possible that we `return 1` from main, but the user has no indication about what went wrong.

src/daos-deserialize/daos-deserialize.c
```diff
@@ -0,0 +1,225 @@
...
+    /* initialize MPI */
+    MPI_Init(&argc, &argv);
+    mfu_init();
+
+    /* get our rank */
+    int rank;
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+    /* verbose by default */
+    mfu_debug_level = MFU_LOG_VERBOSE;
```
daltonbohning, line 49: @dsikich Similar to my comment for `daos-serialize`, maybe we want to add a quiet option

src/daos-deserialize/daos-deserialize.c
```diff
@@ -0,0 +1,225 @@
...
+     * the same, then use the same handle */
+
+    /* create an empty file list */
+    mfu_flist tmplist = mfu_flist_new();
+
+    rc = daos_init();
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to initialize daos");
+        return 1;
+    }
```
daltonbohning, line 139: @dsikich Need to finalize mfu and mpi here

src/daos-deserialize/daos-deserialize.c
```diff
@@ -0,0 +1,225 @@
...
+    /* get size of local list for each rank */
+    uint64_t size = mfu_flist_size(newflist);
+
+    daos_handle_t poh;
+    uuid_t cont_uuids[size];
+    daos_handle_t cohs[size];
+    bool daos_no_prefix = true;
+    daos_pool_info_t pool_info = {0};
+
+#if DAOS_API_VERSION_MAJOR < 1
```
daltonbohning, line 164: @dsikich I'm not 100% sure, but we may be able to remove the DAOS_API_VERSION_MAJOR  check now

src/daos-deserialize/daos-deserialize.c
```diff
@@ -0,0 +1,225 @@
...
+    rc = daos_pool_connect(pool_uuid, NULL, NULL, DAOS_PC_RW,
+                           &poh, &pool_info, NULL);
+#else
+    rc = daos_pool_connect(pool_uuid, NULL, DAOS_PC_RW,
+                           &poh, &pool_info, NULL);
+#endif
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to connect to pool");
+        rc = 1;
+    }
```
daltonbohning, line 174: @dsikich Should this be a fatal error? I.e. should we exit here?

src/daos-deserialize/daos-deserialize.c
```diff
@@ -0,0 +1,225 @@
...
+        uuid_generate(cont_uuids[i]);
+        printf("rank: %d, size: %d path: %s\n", rank, (int)size, path);
+
+
+        /* deserialize this hdf5 file to a DAOS container */
+        rc = cont_deserialize_hdlr(pool_uuid, cont_uuids[i],
+                                   poh, &cohs[i], path);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "Failed to deserialize container (%d)", rc);
+        }
```
daltonbohning, line 188: @dsikich Similar to my comments for `daos-serialize`, this `rc` is overwritten by some later function calls.

src/daos-deserialize/daos-deserialize.c
```diff
@@ -0,0 +1,225 @@
...
+
+        rc = daos_cont_close(cohs[i], NULL);
```
daltonbohning, line 190: @dsikich I.e. here and below, we should maybe use a `tmp_rc`

src/daos-deserialize/daos-deserialize.c
```diff
@@ -0,0 +1,225 @@
...
+    /* Alert the user if there were copy errors */
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "One or more errors were detected while "
+                "serializing: " MFU_ERRF, MFU_ERRP(MFU_ERR_DAOS));
+    }
+
+    rc = daos_fini();
+    if (rc != 0) {
+        rc = 1;
+    }
```
daltonbohning, line 214: @dsikich maybe log a message here so we don't return 1 from main without an indication of what went wrong.

src/common/mfu_daos.c
```diff
@@ -148,7 +154,9 @@ static int daos_parse_path(
         } else {
             strncpy(path, dattr.da_rel_path, path_len);
         }
-    } else if (strncmp(path, "daos:", 5) == 0) {
+    /* da_no_prefix is only used when only daos paths are expected,
+     * so if above parsing fails, then it is an error */
+    } else if (strncmp(path, "daos:", 5) == 0 || daos_no_prefix) {
```
daltonbohning, line 159: @dsikich We might consider allowing the serialize tools to use a daos:// prefix. Maybe not _**require**_ the prefix, but just _**allow**_ it in case a user does pass it.

src/common/mfu_daos.c
```diff
@@ -562,6 +578,8 @@ static int daos_connect(
         /* everything looks good so far */
         valid = 1;
     }
+    if (connect_all_ranks)
+        goto skip_bcast;
 
 bcast:
     /* broadcast valid from rank 0 */
```
daltonbohning, line 585: @dsikich Since some error conditions above have `goto bcast`, we might want to instead do this:
```suggestion

bcast:
    if (connect_all_ranks)
        goto skip_bcast;
    /* broadcast valid from rank 0 */
```

daltonbohning (commented):
@dsikich I'm still going over the core functionality in `mfu_daos.c`, for which I'll post comments once I've had another look

src/common/mfu_daos.c
```diff
@@ -14,6 +14,10 @@
 #include <gurt/common.h>
 #include <gurt/hash.h>
 
+#ifdef HDF5_SUPPORT
+#include <hdf5.h>
+#endif
+
```
daltonbohning, line 20: @dsikich I think this should be compatible with different versions, along with calling H5Sencode1  instead of H5Sencode in the code.
```suggestion
#ifdef HDF5_SUPPORT
#include <hdf5.h>
#if H5_VERS_MAJOR == 1 && H5_VERS_MINOR < 12
#define H5Sencode1 H5Sencode
#endif
#endif

```

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+                goto out;
+            }
+            printf("\tRECX DSET ID WRITTEN: %d\n",
+                (int)hdf5->rx_dset);
+            printf("\tRECX DSPACE ID WRITTEN: %d\n",
+                (int)hdf5->rx_dspace);
+            /* get size of buffer needed
+             * from nalloc
+             */
+            status = H5Sencode(hdf5->rx_dspace, NULL, &nalloc);
```
daltonbohning, line 1551: @dsikich 
```suggestion
            status = H5Sencode1(hdf5->rx_dspace, NULL, &nalloc);
```

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+                MFU_LOG(MFU_LOG_ERR, "Failed to get size of buffer needed");
+                rc = 1;
+                goto out;
+            }
+            /* encode dataspace description
+             * in buffer then store in
+             * attribute on dataset
+             */
+            encode_buf = malloc(nalloc * sizeof(unsigned char));
+            status = H5Sencode(hdf5->rx_dspace, encode_buf,
```
daltonbohning, line 1562: @dsikich 
```suggestion
            status = H5Sencode1(hdf5->rx_dspace, encode_buf,
```

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+    hdf5->akey_dims[0] = 0;     
+    /* data for keys */
+    hdf5->oid_data = NULL;
+    hdf5->dkey_data = NULL;
+    hdf5->akey_data = NULL;
+    hdf5->dk = NULL;
+    hdf5->ak = NULL;
+}
+
+static int serialize_recx_single(struct hdf5_args *hdf5,
```
daltonbohning, line 1365: @dsikich There is some overlap between these iteration functions and the ones for dcp/dsync.
When I refactor the functions for dcp/dsync like we talked about before, I'll see if this can fit in as well.

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+    hid_t               status = 0;
+
+    hdf5->rx_dims[0] = 0;   
+    while (!daos_anchor_is_eof(&recx_anchor)) {
+        memset(recxs, 0, sizeof(recxs));
+        memset(eprs, 0, sizeof(eprs));
+
+        /* list all recx for this dkey/akey */
+            number = 5;
+        rc = daos_obj_list_recx(*oh, DAOS_TX_NONE, dkey,
```
daltonbohning, line 1431: @dsikich Check `rc` and return here?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+            d_iov_set(&iov, buf, buf_len);  
+            printf("\ti: %d iod_size: %d rx_nr:%d, rx_idx:%d\n",
+                   i, (int)size, (int)recxs[i].rx_nr,
+                   (int)recxs[i].rx_idx);
+            /* fetch recx values from source */
+            rc = daos_obj_fetch(*oh, DAOS_TX_NONE, 0, dkey, 1, iod,
+                                &sgl, NULL, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to fetch object: %d", rc);
+            }
```
daltonbohning, line 1468: @dsikich return here?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+                goto out;
+            }
+            /* created attribute in HDF5 file with encoded
+             * dataspace for this record extent */
+            memset(attr_name, 64, sizeof(attr_name));
+            memset(number_str, 16, sizeof(number_str));
+            memset(attr_num_str, 16, sizeof(attr_num_str));
+            path_len = snprintf(number_str, 10, "%d",
+                        (int)(*ak_index));
+            if (path_len >= 16) {
```
daltonbohning, line 1576: @dsikich Why 16? Also, for these 16 and 64 uses, maybe we should have a #define in case we need to change the value?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+                goto out;
+            }
+            path_len = snprintf(attr_name, 64, "%s", "A-");
+            if (path_len >= 64) {
+                MFU_LOG(MFU_LOG_ERR, "attr name is too long");
+                rc = 1;
+                goto out;
+            }
+            strcat(attr_name, number_str);
+            strcat(attr_name, attr_num_str);
```
daltonbohning, line 1594: @dsikich I do think the asprintf, snprintf, etc. functions are more efficient that strcat here

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+                goto out;
+            }
+            if (encode_buf != NULL) 
+                free(encode_buf);
+            attr_num++;
+        }
+    }
+out:
+    return rc;
+}
```
daltonbohning, line 1642: @dsikich In general, I don't think everything is free'ed here on error conditions.
We might consider declaring as many variables as we can at the top of this function and initializing them to NULL. Then, we can easily call mfu_free here on each one of them if rc is not 0.

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+    uint32_t        dkey_number = ENUM_DESC_NR;
+    char            dkey_enum_buf[ENUM_DESC_BUF] = {0};
+    char            dkey[ENUM_KEY_BUF] = {0};
+    char            *dkey_ptr;
+    daos_key_t      diov;
+    int             path_len = 0;
+    hvl_t           *dkey_val;
+    int             size = 0;
+
+    rc = init_recx_data(hdf5);
```
daltonbohning, line 1918: @dsikich Check rc and return here?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+            MFU_LOG(MFU_LOG_ERR, "failed to list akeys: %d", rc);
+            goto out;
+        }
+
+        /* if no akeys returned move on */
+        if (akey_number == 0)
+            continue;
+
+        size = (akey_number + *total_akeys) * sizeof(akey_t);
+        *hdf5->ak = realloc(*hdf5->ak, size);
```
daltonbohning, line 1751: @dsikich NULL check for this realloc?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+                if (err < 0) {
+                    MFU_LOG(MFU_LOG_ERR, "failed to serialize recx array");
+                    rc = 1;
+                    goto out;
+                }
+            } else {
+                hdf5->single_dtype = H5Tcreate(H5T_OPAQUE,
+                                               iod.iod_size);
+                H5Tset_tag(hdf5->single_dtype, "Opaque dtype");
+                hdf5->single_dset = H5Dcreate(hdf5->file,
```
daltonbohning, line 1849: @dsikich Error handing for these create functions?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+                rc = serialize_recx_single(hdf5, &diov, oh,
+                                           &iod);
+                if (rc != 0) {
+                    MFU_LOG(MFU_LOG_ERR, "failed to serialize recx single: %d",
+                            rc);
+                    goto out;
+                }
+                err = H5Dclose(hdf5->single_dset);
+                if (err < 0) {
+                    MFU_LOG(MFU_LOG_ERR, "failed to serialize recx single");
```
daltonbohning, line 1873: @dsikich Maybe a different error message?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+                                NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "failed to list dkeys: %d", rc);
+            goto out;
+        }
+        /* if no dkeys were returned move on */
+        if (dkey_number == 0)
+            continue;
+        *hdf5->dk = realloc(*hdf5->dk,
+                            (dkey_number + *total_dkeys) * sizeof(dkey_t));
```
daltonbohning, line 1946: @dsikich NULL check for this realloc?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+                goto out;
+            }
+            memset(&diov, 0, sizeof(diov));
+            d_iov_set(&diov, (void*)dkey, dkey_kds[i].kd_key_len);
+            dkey_val = &(*hdf5->dk)[*dk_index].dkey_val;
+            size = (uint64_t)dkey_kds[i].kd_key_len * sizeof(char);
+            printf("i:%d dkey iov buf:%s len:%lu\n", i,
+                (char*)diov.iov_buf,
+                (uint64_t)dkey_kds[i].kd_key_len);
+            dkey_val->p = malloc(size);
```
daltonbohning, line 1966: @dsikich NULL check for this malloc?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1330,2195 @@ int mfu_flist_copy_daos(
...
+    }
+    hdf5->status = H5Tinsert(hdf5->akey_memtype, "Akey Value",
+                             HOFFSET(akey_t, akey_val), hdf5->akey_vtype);
+    if (hdf5->status < 0) {
+        MFU_LOG(MFU_LOG_ERR, "failed to insert akey value");
+        rc = 1;
+        goto out;
+    }
+out:
+    return rc;
```
daltonbohning, line 2094: @dsikich In general, if rc is not 0, do we need to close or cleanup anything here?

daltonbohning (commented):
@dsikich In general, it seems that we might not be cleaning up all allocations or handles for error conditions like:
```
rc = 1;
goto out;
```
For each of these new functions, I'd recommend something like:
```
goto err;
...
err:
    // Any potential cleanup
```
Or maybe this pattern:
```
goto out:
...
out:
    if rc != 0
        // cleanup
```

daltonbohning (commented):
@dsikich I wonder if we should name these tools something like `dserialize` to be similar to the rest of the tools?

daltonbohning:
@dsikich I used this to create a directory structure:
`mdtest -C -b 2 -d dir_small -n 100 -w 4k -z 2`
And then `daos-serialize` seemed to complete correctly.
But then when I ran `daos-deserialize`, I get:
`ERROR: failed to create posix container`

daltonbohning:
@dsikich I created a large directory structure like this:
`mdtest -C -b 4 -d dir_large -n 5000 -w 10k -z 4`
And then ran `daos-serialize` (without mpirun), and I get this output:
```
...
cont_str: DAOS_PROP_CO_LAYOUT_TYPE
total dkeys: 8
total akeys: 8
returned 8 oids
HDF5-DIAG: Error detected in HDF5 (1.13.0) MPI-process 0:
  #000: H5D.c line 189 in H5Dcreate2(): unable to synchronously create dataset
    major: Dataset
    minor: Unable to create file
  #001: H5D.c line 137 in H5D__create_api_common(): unable to create dataset
    major: Dataset
    minor: Unable to create file
  #002: H5VLcallback.c line 1809 in H5VL_dataset_create(): dataset create failed
    major: Virtual Object Layer
    minor: Unable to create file
  #003: H5VLcallback.c line 1774 in H5VL__dataset_create(): dataset create failed
    major: Virtual Object Layer
    minor: Unable to create file
  #004: H5VLnative_dataset.c line 73 in H5VL__native_dataset_create(): unable to create dataset
    major: Dataset
    minor: Unable to initialize object
  #005: H5Dint.c line 396 in H5D__create_named(): unable to create and link to dataset
    major: Dataset
    minor: Unable to initialize object
  #006: H5L.c line 2359 in H5L_link_object(): unable to create new link to object
    major: Links
    minor: Unable to initialize object
  #007: H5L.c line 2601 in H5L__create_real(): can't insert link
    major: Links
    minor: Unable to insert object
  #008: H5Gtraverse.c line 838 in H5G_traverse(): internal path traversal failed
    major: Symbol table
    minor: Object not found
  #009: H5Gtraverse.c line 614 in H5G__traverse_real(): traversal operator failed
    major: Symbol table
    minor: Callback failed
  #010: H5L.c line 2394 in H5L__link_cb(): name already exists
    major: Links
    minor: Object already exists
[2021-02-10T22:47:25] [0] [/home/dbohninx/mpifileutils_dev/mpifileutils/src/common/mfu_daos.c:2672] ERROR: failed to create oid dset
```

dsikich:
> @dsikich I created a large directory structure like this:
> `mdtest -C -b 4 -d dir_large -n 5000 -w 10k -z 4`
> And then ran `daos-serialize` (without mpirun), and I get this output:
> 
> ```
> ...
> cont_str: DAOS_PROP_CO_LAYOUT_TYPE
> total dkeys: 8
> total akeys: 8
> returned 8 oids
> HDF5-DIAG: Error detected in HDF5 (1.13.0) MPI-process 0:
>   #000: H5D.c line 189 in H5Dcreate2(): unable to synchronously create dataset
>     major: Dataset
>     minor: Unable to create file
>   #001: H5D.c line 137 in H5D__create_api_common(): unable to create dataset
>     major: Dataset
>     minor: Unable to create file
>   #002: H5VLcallback.c line 1809 in H5VL_dataset_create(): dataset create failed
>     major: Virtual Object Layer
>     minor: Unable to create file
>   #003: H5VLcallback.c line 1774 in H5VL__dataset_create(): dataset create failed
>     major: Virtual Object Layer
>     minor: Unable to create file
>   #004: H5VLnative_dataset.c line 73 in H5VL__native_dataset_create(): unable to create dataset
>     major: Dataset
>     minor: Unable to initialize object
>   #005: H5Dint.c line 396 in H5D__create_named(): unable to create and link to dataset
>     major: Dataset
>     minor: Unable to initialize object
>   #006: H5L.c line 2359 in H5L_link_object(): unable to create new link to object
>     major: Links
>     minor: Unable to initialize object
>   #007: H5L.c line 2601 in H5L__create_real(): can't insert link
>     major: Links
>     minor: Unable to insert object
>   #008: H5Gtraverse.c line 838 in H5G_traverse(): internal path traversal failed
>     major: Symbol table
>     minor: Object not found
>   #009: H5Gtraverse.c line 614 in H5G__traverse_real(): traversal operator failed
>     major: Symbol table
>     minor: Callback failed
>   #010: H5L.c line 2394 in H5L__link_cb(): name already exists
>     major: Links
>     minor: Object already exists
> [2021-02-10T22:47:25] [0] [/home/dbohninx/mpifileutils_dev/mpifileutils/src/common/mfu_daos.c:2672] ERROR: failed to create oid dset
> ```

@daltonbohning did it even create the hdf5 file? it looks like the file already exists?

dsikich:
> @dsikich I used this to create a directory structure:
> `mdtest -C -b 2 -d dir_small -n 100 -w 4k -z 2`
> And then `daos-serialize` seemed to complete correctly.
> But then when I ran `daos-deserialize`, I get:
> `ERROR: failed to create posix container`

@daltonbohning are you creating the dir structure inside a daos container with dfuse? I find it odd that it wouldn't be able to create a POSIX container because that is the first thing it is supposed to do.. anyway this is helpful and I will be able to take a look Thursday and Friday, thanks

daltonbohning:
> > @dsikich I created a large directory structure like this:
> > `mdtest -C -b 4 -d dir_large -n 5000 -w 10k -z 4`
> > And then ran `daos-serialize` (without mpirun), and I get this output:
> > ```
> > ...

> @daltonbohning did it even create the hdf5 file? it looks like the file already exists?

I ran it again. The file did NOT exist BEFORE trying to deserialize, but the file was in fact created, and it has some amount of data in it:
`3120 -rw-rw-r--   1 dbohninx dbohninx 3194696 Feb 10 22:56 ddff6799-cb1e-49e8-a6ab-6813ed4e4ee5`

dsikich:
> > > @dsikich I created a large directory structure like this:
> > > `mdtest -C -b 4 -d dir_large -n 5000 -w 10k -z 4`
> > > And then ran `daos-serialize` (without mpirun), and I get this output:
> > > ```
> > > ...
> > > ```
> 
> > @daltonbohning did it even create the hdf5 file? it looks like the file already exists?
> 
> I ran it again. The file did NOT exist BEFORE trying to deserialize, but the file was in fact created, and it has some amount of data in it:
> `3120 -rw-rw-r-- 1 dbohninx dbohninx 3194696 Feb 10 22:56 ddff6799-cb1e-49e8-a6ab-6813ed4e4ee5`

@daltonbohning thanks, I will re-create this

daltonbohning:
> > @dsikich I used this to create a directory structure:
> > `mdtest -C -b 2 -d dir_small -n 100 -w 4k -z 2`
> > And then `daos-serialize` seemed to complete correctly.
> > But then when I ran `daos-deserialize`, I get:
> > `ERROR: failed to create posix container`
> 
> @daltonbohning are you creating the dir structure inside a daos container with dfuse? I find it odd that it wouldn't be able to create a POSIX container because that is the first thing it is supposed to do.. anyway this is helpful and I will be able to take a look Thursday and Friday, thanks

Yes, I ran that `mdtest` command inside the daos container.

dsikich:
> > > > @dsikich I created a large directory structure like this:
> > > > `mdtest -C -b 4 -d dir_large -n 5000 -w 10k -z 4`
> > > > And then ran `daos-serialize` (without mpirun), and I get this output:
> > > > ```
> > > > ...
> > > > ```
> > 
> > 
> > > @daltonbohning did it even create the hdf5 file? it looks like the file already exists?
> > 
> > 
> > I ran it again. The file did NOT exist BEFORE trying to deserialize, but the file was in fact created, and it has some amount of data in it:
> > `3120 -rw-rw-r-- 1 dbohninx dbohninx 3194696 Feb 10 22:56 ddff6799-cb1e-49e8-a6ab-6813ed4e4ee5`
> 
> @daltonbohning thanks, I will re-create this

@daltonbohning can you show a h5dump <file>

dsikich:
@daltonbohning this push should fix the issues you were seeing running on the mdtest datasets

src/daos-deserialize/daos-deserialize.c
```diff
@@ -26,12 +24,107 @@ void print_usage(void)
...
+            rc = ENOMEM;
+            goto out;
+        }
+
+        for (i = 0; i < *numpaths; i++) {
+            (*paths)[i] = MFU_STRDUP(argpaths[i]);
+            if ((*paths)[i] == NULL) {
+                rc = ENOMEM;
+                goto out;
+            }
```
daltonbohning, line 121: @dsikich similar to above
```suggestion
            if ((*paths)[i] == NULL) {
                MFU_LOG(MFU_LOG_ERR, "Failed to allocate paths.");
                rc = 1;
                goto out;
            }
```

src/daos-deserialize/daos-deserialize.c
```diff
@@ -26,12 +24,107 @@ void print_usage(void)
...
+        }
+    }
+out:
+    return rc;
```
daltonbohning, line 125: @dsikich On error, the paths would not be free'ed. Maybe we can set `num_files = *numpaths` for the `else` condition in this function, and then just do something like this:
```suggestion
out:
    if (rc != 0) {
        if (*paths != NULL) {
            for (i = 0; i < num_files; i++) {
                mfu_free(&(*paths)[i]); // TODO check this pointer handling
            }
            mfu_free(paths);
        }
    }
    return rc;
```

src/daos-deserialize/daos-deserialize.c
```diff
@@ -244,7 +371,7 @@ int main(int argc, char** argv)
 
     rc = daos_fini();
     if (rc != 0) {
-        rc = 1;
+        MFU_LOG(MFU_LOG_ERR, "Failed to finalize DAOS (%d)", rc);
```
daltonbohning, line 374: @dsikich Formatted -DER
```suggestion
        MFU_LOG(MFU_LOG_ERR, "Failed to finalize DAOS "DF_RC, DP_RC(rc));
```

src/daos-serialize/daos-serialize.c
```diff
@@ -208,7 +246,7 @@ int main(int argc, char** argv)
 
     rc = daos_fini();
     if (rc != 0) {
-        rc = 1;
+        MFU_LOG(MFU_LOG_ERR, "Failed to finalize DAOS (%d)", rc);
```
daltonbohning, line 249: @dsikich Formatted -DER
```suggestion
        MFU_LOG(MFU_LOG_ERR, "Failed to finalize DAOS "DF_RC, DP_RC(rc));
```

src/common/mfu_daos.c
```diff
@@ -1430,17 +1427,18 @@ static int serialize_recx_array(struct hdf5_args *hdf5,
...
         /* list all recx for this dkey/akey */
-            number = 5;
+        number = 5;
         rc = daos_obj_list_recx(*oh, DAOS_TX_NONE, dkey,
                                 akey, &size, &number, recxs, eprs, &recx_anchor,
                                 true, NULL);
-        printf("RECX RETURNED: %d\n", (int)number);
-        printf("RECX SIZE: %d\n", (int)size);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "Failed to list recx: %d", rc);
```
daltonbohning, line 1435: @dsikich Formatted -DER
```suggestion
            MFU_LOG(MFU_LOG_ERR, "Failed to list recx "DF_RC, DP_RC(rc));
```

src/common/mfu_daos.c
```diff
@@ -1571,35 +1549,32 @@ static int serialize_recx_array(struct hdf5_args *hdf5,
...
              * dataspace for this record extent */
-            memset(attr_name, 64, sizeof(attr_name));
-            memset(number_str, 16, sizeof(number_str));
-            memset(attr_num_str, 16, sizeof(attr_num_str));
-            path_len = snprintf(number_str, 10, "%d",
-                        (int)(*ak_index));
-            if (path_len >= 16) {
+            memset(attr_name, ATTR_NAME_LEN, sizeof(attr_name));
+            memset(number_str, ATTR_NAME_LEN, sizeof(number_str));
+            memset(attr_num_str, ATTR_NAME_LEN, sizeof(attr_num_str));
```
daltonbohning, line 1554: @dsikich Is this what you meant?
```suggestion
            memset(attr_name, 0, ATTR_NAME_LEN);
            memset(number_str, 0, ATTR_NAME_LEN);
            memset(attr_num_str, 0, ATTR_NAME_LEN);
```

src/common/mfu_daos.c
```diff
@@ -1841,22 +1828,29 @@ static int serialize_akeys(struct hdf5_args *hdf5,
             } else {
                 hdf5->single_dtype = H5Tcreate(H5T_OPAQUE,
                                                iod.iod_size);
-                H5Tset_tag(hdf5->single_dtype, "Opaque dtype");
+                if (hdf5->single_dtype < 0) {
+                    MFU_LOG(MFU_LOG_ERR, "failed to create single record dtype");
+                    rc = 1;
+                    goto out;
+                }
+                err = H5Tset_tag(hdf5->single_dtype, "Opaque dtype");
```
daltonbohning, line 1836: @dsikich I think if we're going to set the tag, we should make it descriptive in case someone looks at it in the HDF5 file. For example:
"dkey value"
"akey value"
etc.

src/common/mfu_daos.c
```diff
@@ -2862,21 +2871,32 @@ int cont_serialize_hdlr(int rank, uint64_t *files_written,  daos_args_t *da,
...
+    char            filename[FILENAME_LEN];
+    char            cont_str[FILENAME_LEN];
+    char            rank_str[FILENAME_LEN];
 
     /* init HDF5 args */
     init_hdf5_args(&hdf5);
 
     memset(rank_str, 0, sizeof(rank_str));
     memset(cont_str, 0, sizeof(cont_str));
     memset(filename, 0, sizeof(filename));
```
daltonbohning: @dsikich Maybe it's okay to use sizeof here since they are static, but I think this is less error prone in the general case:
```suggestion
    memset(rank_str, 0, FILENAME_LEN);
    memset(cont_str, 0, FILENAME_LEN);
    memset(filename, 0, FILENAME_LEN);
```

src/daos-serialize/daos-serialize.c
dsikich: @adammoody this has been updated since we are now only using one pool/cont and writing a different hdf5 file for each rank

src/common/mfu_daos.c
dsikich: @daltonbohning oof, thanks for catching that. Yes, that is what I meant to do.

src/common/mfu_daos.c
dsikich: @daltonbohning I don't think it is really necessary to set the tags, and this is just the datatype for the byte data, so there is no real clear label to give them like dkey/akey value. I can remove it.

daltonbohning: @dsikich Yeah, I don't think it's necessary either. I think the intent for "tags" is that since it's an opaque datatype, you can specify a tag as a description for the data. But I don't think that's necessary in this case, because we give the dataset itself a meaningful name.

dsikich:
> @dsikich I wonder if we should name these tools something like `dserialize` to be similar to the rest of the tools?

@daltonbohning I think dserialize would be a little confusing and get mixed up with deserialize. Maybe we could have had one dserialize tool with --serialize and --deserialize options, but I'm not sure if that is more clear or not. Most of the functionality is in mfu_daos vs the tools anyway.

daltonbohning (commented):
@dsikich I still want to run some manual tests again as a sanity check before approving this.
I've also been working on some test infrastructure to generate and verify some dataset at the obj level, so I'd like to run that as a sanity check as well.

src/common/mfu_daos.c
```diff
@@ -1313,3 +1326,2840 @@ int mfu_flist_copy_daos(
...
+    d_iov_set(&iov, buf, buf_len);
+    rc = daos_obj_fetch(*oh, DAOS_TX_NONE, 0, dkey, 1, iod, &sgl,
+                        NULL, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to fetch object");
+        goto out;
+    }
+    /* write single val record to dataset */
+    H5Dwrite(hdf5->single_dset, hdf5->single_dtype, H5S_ALL,
+             hdf5->single_dspace, H5P_DEFAULT, sgl.sg_iovs[0].iov_buf);
```
daltonbohning, line 1392: @dsikich Error check for write?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1326,2840 @@ int mfu_flist_copy_daos(
...
+            MFU_LOG(MFU_LOG_ERR, "failed to close attr");
+            rc = 1;
+            goto out;
+        }
+        status = H5Sclose(hdf5->rx_memspace);
+        if (status < 0) {
+            MFU_LOG(MFU_LOG_ERR, "failed to close rx_memspace");
+            rc = 1;
+            goto out;
+        }
```
daltonbohning, line 1623: @dsikich These goto statements cause an infinite loop. Maybe we don't need to log or check the status here at all, since this is for an error condition anyway.

```suggestion
        H5Aclose(hdf5->selection_attr);
        H5Sclose(hdf5->rx_memspace);
```

daltonbohning: @dsikich Actually, it also seems that HDF5 itself throws errors if trying to close something that isn't valid, so I think we might also need to only try to close if it was opened/valid/etc., but I'm not sure the best way.

src/common/mfu_daos.c
```diff
@@ -1313,3 +1326,2840 @@ int mfu_flist_copy_daos(
...
+                goto out;
+            }
+            status = H5Dwrite(hdf5->rx_dset, hdf5->rx_dtype,
+                              hdf5->rx_memspace, hdf5->rx_dspace,
+                              H5P_DEFAULT, sgl.sg_iovs[0].iov_buf);
+            if (status < 0) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to write rx_dset");
+                rc = 1;
+                goto out;
+            }
```
daltonbohning, line 1517: @dsikich For a simple dataset I get:
```
HDF5-DIAG: Error detected in HDF5 (1.13.0) MPI-process 0:
#000: H5D.c line 1156 in H5Dwrite(): can't synchronously write data
...
  #005: H5Dio.c line 395 in H5D__write(): unable to set up type info
    major: Dataset
    minor: Unable to initialize object
  #006: H5Dio.c line 693 in H5D__typeinfo_init(): unable to convert between src and dest datatype
    major: Dataset
    minor: Feature is unsupported
    minor: Write failed
```

dsikich: @daltonbohning hmm I must have updated something that I thought wouldn't matter before I pushed this since this is new. Thanks, I will take a look.

src/common/mfu_daos.c
```diff
@@ -1313,3 +1326,2840 @@ int mfu_flist_copy_daos(
...
+    rc = cont_serialize_props(&hdf5, da->src_coh);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "failed to serialize cont layout");
+        goto out;
+    }
+
+    /* sum files_written across all ranks to get total */
+    uint64_t total_files_written = 0;
+    MPI_Allreduce(files_written, &total_files_written, 1, MPI_UNSIGNED,
+               MPI_SUM, MPI_COMM_WORLD);
```
daltonbohning, line 2973: @dsikich All/most of the goto statements above this reduction might potentially cause this to hang if only some of the ranks error.

src/common/mfu_daos.c
dsikich: @daltonbohning can you try again with my latest push. I deleted the set_tag functions and forgot one, so I think that what causing the write error.

daltonbohning: @dsikich That seems to have worked, thanks!

daltonbohning:
@dsikich If I try to serialize a container with more ranks than there are objects in the container, I get:
```
[/home/dbohninx/mpifileutils_dev/mpifileutils/src/daos-serialize/daos-serialize.c:202] ERROR: Failed to serialize container (0)
```
For EACH "extra" rank. And then the application hangs (maybe because of the previous reduction issue noted before).

src/daos-deserialize/daos-deserialize.c
```diff
@@ -0,0 +1,420 @@
...
+    /* advance to next set of options */
+    optind += numpaths;
+
+    if (numpaths < 1) {
+        MFU_LOG(MFU_LOG_ERR, "At least one file or directory is required:"
+                MFU_ERRF, MFU_ERRP(-MFU_ERR_INVAL_ARG));
+        mfu_finalize();
+        MPI_Finalize();
+        return 1;
+    }
```
daltonbohning, line 254: @dsikich We also need to make sure the --pool is passed in. Not supplying one currently gives an error.

dsikich: > @dsikich If I try to serialize a container with more ranks than there are objects in the container, I get:
> 
> ```
> [/home/dbohninx/mpifileutils_dev/mpifileutils/src/daos-serialize/daos-serialize.c:202] ERROR: Failed to serialize container (0)
> ```
> 
> For EACH "extra" rank. And then the application hangs (maybe because of the previous reduction issue noted before).

@daltonbohning yeah, it is because not all ranks call Allreduce. I think I can find a way to move this out of the handler and just pass in the total to it.

dsikich: @daltonbohning I might also just be able to split MPI_COMM_WORLD and call Allreduce on only the ranks doing work, thanks

DAOS-Support.md
```diff
@@ -163,3 +155,57 @@ Number of items that exist in both directories and have different types: 0 (Src:
...
+
+#### Example One
+
+Show the serialization of a DAOS container
+
+```shell
+$ mpirun -np 3 daos-serialize -v /$pool1/$cont1
+Serializing Container to 0-7bf8037d-823f-4fa5-ac2a-c2cae8f81f57
+Serializing Container to 1-7bf8037d-823f-4fa5-ac2a-c2cae8f81f57
+Serializing Container to 2-7bf8037d-823f-4fa5-ac2a-c2cae8f81f57
```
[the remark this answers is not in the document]
dsikich: @daltonbohning hmm I think maybe I could use <rank>_ I prefer to use it as a prefix though, because it is easier to specify each one if you are adding the files one by one to the command line with tab completion.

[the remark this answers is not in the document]
dsikich: @daltonbohning I can also add "rank"

DAOS-Support.md
```diff
@@ -163,3 +155,57 @@ Number of items that exist in both directories and have different types: 0 (Src:
...
+Serializing Container to serialize/1-7bf8037d-823f-4fa5-ac2a-c2cae8f81f57
+Serializing Container to serialize/2-7bf8037d-823f-4fa5-ac2a-c2cae8f81f57
+```
+#### Example Three 
+
+Show the deserialization of an HDF5 file
+
+```shell
+$ mpirun -np 3 daos-deserialize -v 0-7bf8037d-823f-4fa5-ac2a-c2cae8f81f57
+1-7bf8037d-823f-4fa5-ac2a-c2cae8f81f57 2-7bf8037d-823f-4fa5-ac2a-c2cae8f81f57
```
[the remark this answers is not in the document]
dsikich: @daltonbohning hmm I see, yeah that's a good idea we can do that

src/daos-serialize/daos-serialize.c
```diff
@@ -0,0 +1,263 @@
...
+
+    tmp_rc = daos_cont_serialize_hdlr(rank, output_path, &files_written,
+                                      daos_args, newflist, size);
+    if (tmp_rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to serialize container (%d)", rc);
+        rc = 1;
+    }
+
+    /* destroy snapshot after copy */
+    /* TODO consider moving this into mfu_flist_copy_daos */
```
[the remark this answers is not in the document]
dsikich: @daltonbohning ok, I can update the daos_connect and remove the destroy snapshot after the dsync PR is merged. Since I will probably get a few merge conflicts anyway.

src/daos-deserialize/daos-deserialize.c
```diff
@@ -0,0 +1,420 @@
...
+                    rc = ENOMEM;
+                    goto out;
+                }
+                len = snprintf((*paths)[num_files], path_len, "%s", name);
+                if (len >= path_len) {
+                    MFU_LOG(MFU_LOG_ERR, "filename is too long");
+                    rc = 1;
+                    mfu_free(&path);
+                    goto out;
+                }
```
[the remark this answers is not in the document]
dsikich: @daltonbohning yep :-)

src/common/mfu_daos.c
dsikich: @daltonbohning I think we can maybe check if these are > 0, and if they are then they should be closed

dsikich:
> @dsikich If I try to serialize a container with more ranks than there are objects in the container, I get:
> 
> ```
> [/home/dbohninx/mpifileutils_dev/mpifileutils/src/daos-serialize/daos-serialize.c:202] ERROR: Failed to serialize container (0)
> ```
> 
> For EACH "extra" rank. And then the application hangs (maybe because of the previous reduction issue noted before).

@daltonbohning This should be fixed now in my latest push where I addressed your comments. The chunk size I had set was also way too small, which increased the serialization of bigger files (1GB, etc) dramatically.

DAOS-Support.md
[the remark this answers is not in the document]
daltonbohning: @dsikich I'm not set on any way in particular, so we can always revisit this later if needed

DAOS-Support.md
```diff
@@ -1,7 +1,13 @@
 # DAOS Support
 
-[DAOS](https://github.com/daos-stack/daos) is supported as a backend storage system in dcp, dsync, and dcmp. The build instructions for
-enabling DAOS support can be found here:
+[DAOS](https://github.com/daos-stack/daos) is supported as a backend storage system in
+dcp, dsync, and dcmp. Custom serialization and deserialization for DAOS
+containers to and from a POSIX filesystem is provided with daos-serialize and
+daos-deserialize. These tools serialize a DAOS container and its associated
+metadata to an HDF5 file for storage on a POSIX filesystem. The container data
```
adammoody, line 7: Rewording a bit to be specific about naming which tool to use for generating the HDF file.
```suggestion
daos-deserialize. A DAOS container and its associated
metadata can be serialized to an HDF5 file for storage on a POSIX filesystem using daos-serialize. The container data
```

adammoody:
Thanks for all of the work on this @dsikich and @daltonbohning !  That's a ton of code, reviews, and testing in there!

DAOS-Support.md
adammoody, line 178: I haven't used HDF a lot, but is it convention to use an .hdf extension on those?  That might help people remember the file format, especially if other formats might be supported in the future.  I'm not stuck on that, though, and we can add it later if we think it'd be helpful.

doc/rst/daos-serialize.1.rst
```diff
@@ -0,0 +1,54 @@
...
+
+DESCRIPTION
+-----------
+
+Parallel MPI application to serialize a DAOS container to an HDF5 file.
+
+daos-serialize is a serialization tool that will allow storing any type
+of DAOS container on a POSIX filesystem. It allows DAOS container data to
+be stored outside of the DAOS system. The DAOS container is converted
+to an HDF5 format. The serialization format of the container is meant to
```
adammoody, line 17: Still accurate if I make this change?
```suggestion
to a set of files in HDF5 format. The serialization format of the container is meant to
```

adammoody (approved):

DAOS-Support.md
daltonbohning: @dsikich I think we probably should use a `.h5`  or `.hdf5` extension, per Mohamad's comment. Also, we should update the examples in this file with the `_rank` suffix you added.

doc/rst/daos-serialize.1.rst
daltonbohning: I think this is more clear.

dsikich:
@adammoody thanks for the review!

DAOS-Support.md
dsikich: @daltonbohning this has been updated.

daltonbohning (approved):

src/common/mfu_daos.c
```diff
@@ -1881,3 +1894,2824 @@ int mfu_daos_flist_sync(
...
+            status = H5Sselect_hyperslab(hdf5->rx_dspace,
+                                         H5S_SELECT_AND, &start,
+                                         NULL, &count, NULL);
+            if (status < 0) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to select hyperslab");
+                rc = 1;
+                goto out;
+            }
+
+            hdf5->rx_dtype = H5Tcreate(H5T_OPAQUE, (*iod).iod_size);
```
daltonbohning, line 2077: @dsikich if `iod_size` is 0:
```
HDF5-DIAG: Error detected in HDF5 (1.13.0) MPI-process 0:
  #000: H5T.c line 1843 in H5Tcreate(): size must be positive
    major: Invalid arguments to routine
    minor: Bad value
[2021-03-01T20:54:05] [0] [/home/dbohninx/mpifileutils_dev/mpifileutils/src/common/mfu_daos.c:2079] ERROR: Failed to create rx_dtype
```

src/common/mfu_daos.c
```diff
@@ -1881,3 +1894,2824 @@ int mfu_daos_flist_sync(
...
+
+            /* set sgl values */
+            sgl.sg_nr     = 1;
+            sgl.sg_nr_out = 0;
+            sgl.sg_iovs   = &iov;
+
+            d_iov_set(&iov, buf, buf_len);  
+            /* fetch recx values from source */
+            rc = daos_obj_fetch(*oh, DAOS_TX_NONE, 0, dkey, 1, iod,
+                                &sgl, NULL, NULL);
```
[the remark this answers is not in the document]
dsikich: @daltonbohning fix added in my latest push

src/common/mfu_daos.c
```diff
@@ -1881,3 +1901,2836 @@ int mfu_daos_flist_sync(
...
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to fetch object "DF_RC, DP_RC(rc));
+                goto out;
+            }
+
+            /* reset size to 0 after a successful fetch */
```
daltonbohning, line 2084: @dsikich 
```suggestion
            /* reset size after a successful fetch */
```

src/common/mfu_daos.c
```diff
@@ -1881,3 +1901,2836 @@ int mfu_daos_flist_sync(
...
+    daos_recx_t         recxs[5];
+    daos_size_t         size = 0;
+    char                attr_name[ATTR_NAME_LEN];
+    char                number_str[ATTR_NAME_LEN];
+    char                attr_num_str[ATTR_NAME_LEN];
+    unsigned char       *encode_buf = NULL;
+    d_sg_list_t         sgl;
+    d_iov_t             iov;
+    hid_t               status = 0;
+
```
daltonbohning, line 2005: @dsikich We need to make sure these are initialized to 0, so if we `goto out` before opening/creating them, they are valid integers.
```suggestion

    hdf5->rx_dset = 0;
    hdf5->selection_attr = 0;
    hdf5->rx_memspace = 0;

```

src/common/mfu_daos.c
```diff
@@ -1881,3 +1901,2836 @@ int mfu_daos_flist_sync(
...
+    number = 1;
+    rc = daos_obj_list_recx(*oh, DAOS_TX_NONE, dkey,
+                            akey, &size, &number, NULL, eprs, &fetch_anchor,
+                            true, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to list recx "DF_RC, DP_RC(rc));
+        goto out;
+    }
+
+    /* create the dataset with the correct type size */
```
daltonbohning, line 2017: @dsikich See my below comment for why this is necessary.
```suggestion
    if (size > 2000) {
        MFU_LOG(MFU_LOG_ERR, "recx size is too large: %llu", size);
        rc = 1;
        goto out;
    }

    /* create the dataset with the correct type size */
```

src/common/mfu_daos.c
```diff
@@ -1881,3 +1901,2836 @@ int mfu_daos_flist_sync(
...
+                MFU_LOG(MFU_LOG_ERR, "failed to create selection attr");
+                rc = 1;
+                goto out;
+            }   
+            status = H5Awrite(hdf5->selection_attr, hdf5->rx_dtype,
+                              encode_buf);
+            if (status < 0) {
+                MFU_LOG(MFU_LOG_ERR, "failed to write attr");
+                rc = 1;
+                goto out;
```
daltonbohning, line 2211: @dsikich Apparently, HDF5 has a limit of 2000 characters for metadata columns. I don't fully understand it, but this fails if roughly `size > 2000`. For now, we should probably enforce some limit on `size` above when we fetch. Maybe if `size > 2000` we can print and return an error.

src/common/mfu_daos.c
```diff
@@ -1881,3 +1901,2836 @@ int mfu_daos_flist_sync(
...
+     * create the dataset with the correct datatype size */
+    number = 1;
+    rc = daos_obj_list_recx(*oh, DAOS_TX_NONE, dkey,
+                            akey, &size, &number, NULL, eprs, &fetch_anchor,
+                            true, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to list recx "DF_RC, DP_RC(rc));
+        goto out;
+    }
+
```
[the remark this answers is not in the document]
dsikich: @daltonbohning I think that is ok

src/common/mfu_daos.c
dsikich: @daltonbohning hmm ok, I wonder if this would happen very often, were you testing with larger and larger sizes of recx?

daltonbohning: @dsikich Yeah, I was using recx sizes of 2049 and 4097

src/common/mfu_daos.c
```diff
@@ -1392,6 +1392,7 @@ static int mfu_daos_obj_sync_recx_array(
         (*iod).iod_nr    = number;
         (*iod).iod_recxs = recxs;
         (*iod).iod_size  = size;
+        (*iod).iod_name = *akey;
```
dsikich, line 1395: @daltonbohning is there a reason this was added to the sync_ function instead of the serialization one?

src/common/mfu_daos.c
```diff
@@ -2066,6 +2081,7 @@ static int serialize_recx_array(struct hdf5_args *hdf5,
             (*iod).iod_size  = size;
             (*iod).iod_nr    = 1;
             (*iod).iod_recxs = &recxs[i];
+            (*iod).iod_name = *akey;
```
dsikich, line 2084: @daltonbohning in serialize_akeys this line:
   d_iov_set(&iod.iod_name, (void*)akey, strlen(akey));  
I think it needs to use akey_kds[j].kd_key_len instead of strlen. I am not sure that is the correct way to set the iod_name.

dsikich: @daltonbohning either way, I think the d_iov_set call needs to use kd_key_len instead of strlen

daltonbohning: @dsikich I think the reason dfs.c only uses d_iov_set is because in dfs.c, there isn't a daos_key_t instance for an akey.
I do see some examples of this way, such as here:
https://github.com/daos-stack/daos/blob/c9ab50cd1047f483c4b8720cda086ce0611ecf7a/src/object/srv_ec_aggregate.c#L528
This works because a daos_key_t is actually just a d_iov_t:
https://github.com/daos-stack/daos/blob/c9ab50cd1047f483c4b8720cda086ce0611ecf7a/src/include/daos_types.h#L122

But yes, we should probably use kd_key_len in that case.

dsikich: @daltonbohning thanks, yeah I saw that the key was really just a d_iov_t last night after writing this comment and looking up the types, but yeah I think we will just need to remove the strlen in the d_iov_set call

src/common/mfu_daos.c
```diff
@@ -1881,3 +1912,2856 @@ int mfu_daos_flist_sync(
...
+        *hdf5->ak = realloc(*hdf5->ak, size);
+        if (*hdf5->ak == NULL) {
+            rc = ENOMEM;
+            goto out;
+        }
+
+        /* parse out individual akeys based on key length and
+         * numver of dkeys returned
+         */
+        (*hdf5->dk)[*dk_index].akey_offset = *ak_index;
```
daltonbohning, line 2387: @dsikich We need to set this BEFORE the while loop above. Otherwise, it gets overwritten

PR #458, opened 2021-02-05
daos dsync: simple DAOS object support

daltonbohning:
Abstracted some code from `dcp`.
Added very simple support to `dsync` for DAOS objects.
Improved some error handling in `dsync`.
Added supporting functions to `mfu_daos`.
Renamed some DAOS functions to use the prefix mfu_daos.

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -389,9 +389,48 @@ static int daos_set_paths(
             argpaths[1] = da->dst_path = strdup(dst_path);
             mfu_free(&dst_path);
         } else if (dst_rc == -1) {
-            MFU_LOG(MFU_LOG_ERR, "Failed to parse DAOS destination path: daos://<pool>/<cont>[/<path>]");
+            /* The destination has a daos: prefix, so try to parse just the pool uuid. */
+            char* saveptr;
+            char* t = strtok_r(dst_path, "/", &saveptr);
+            if (t == NULL) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to parse DAOS destination path: daos://<pool>/<cont>[/<path>]");
+                rc = 1;
+                goto dst_out;
+            }
+            /* Skip over the daos: prefix */
+            t = strtok_r(NULL, "/", &saveptr);
+            if (t == NULL) {
+                /* No pool provided, so use the source pool uuid */
+                if (!daos_uuid_valid(da->src_pool_uuid)) {
+                    MFU_LOG(MFU_LOG_ERR, "Destination pool and/or container expected.");
+                    rc = 1;
+                    goto dst_out;
+                }
+                uuid_copy(da->dst_pool_uuid, da->src_pool_uuid);
+            } else {
+                /* Make sure the provided pool uuid is valid */
+                dst_rc = uuid_parse(t, da->dst_pool_uuid);
+                if (dst_rc != 0) {
+                    MFU_LOG(MFU_LOG_ERR, "Failed to parse destination pool UUID");
+                    rc = 1;
+                    goto dst_out;
+                }
+            }
+            /* The pool uuid should have been last */
+            t = strtok_r(NULL, "/", &saveptr);
+            if (t != NULL) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to parse DAOS destination path: daos://<pool>/<cont>[/<path>]");
+                rc = 1;
+                goto dst_out;
+            }
+            /* Generate a new container uuid */
+            uuid_generate(da->dst_cont_uuid);
+            argpaths[1] = da->dst_path = strndup("/", 1);
+
+            rc = 0;
+
+dst_out:
             mfu_free(&dst_path);
-            rc = 1;
             goto out;
         }
     }
@@ -433,22 +472,13 @@ static int daos_get_cont_type(
  */
 static int daos_set_api_cont_type(
     mfu_file_t* mfu_file,
-    daos_handle_t coh,
+    enum daos_cont_props cont_type,
     daos_api_t api)
 {
     /* If explicitly using DAOS, just set the type to DAOS */
     if (api == DAOS_API_DAOS) {
         mfu_file->type = DAOS;
         return 0;
-    }
-
-    /* Otherwise, query the container type, and use DFS for POSIX containers. */
-    enum daos_cont_props cont_type;
-
-    int rc = daos_get_cont_type(coh, &cont_type);
-    if (rc) {
-        MFU_LOG(MFU_LOG_ERR, "Failed to get DAOS container type.");
-        return rc;
     }
 
     if (cont_type == DAOS_PROP_CO_LAYOUT_POSIX) {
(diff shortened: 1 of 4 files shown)


src/common/mfu_daos.c
```diff
@@ -585,7 +680,7 @@ static int daos_connect(
 }
 
 /* Mount DAOS dfs */
-static int daos_mount(
+static int mfu_dfs_mount(
```
dsikich, line 683: @daltonbohning what determines if we use the mfu_dfs_ prefix vs. the mfu_daos_ prefix?

src/common/mfu_daos.c
```diff
@@ -919,112 +1056,281 @@ int daos_cleanup(
...
+    daos_recx_t*    recxs,  /* array of recxs */
+    d_iov_t*        iov,    /* array of iovs */
+    char**          buf,    /* pointer to static buffer */
+    uint64_t*       buf_len)/* pointer to static buffer lengths */
+{
+    for (uint32_t i = 0; i < number; i++) {
+        buf_len[i] = recxs[i].rx_nr * size;
+        buf[i] = MFU_MALLOC(buf_len[i]);
+        if (buf[i] == NULL) {
+            free_iov_buf(number, buf);
```
dsikich, line 1162: @daltonbohning I think this is similar to the comment you made on the cont clone PR, in that it doesn't look like we are freeing memory that was already allocated before but[I] in the case of an error.

src/dsync/dsync.c
```diff
@@ -2869,6 +2871,91 @@ static int dsync_validate_link_dest(const char *link_dest, const char *dest, mfu
...
+    if (options.delete && mfu_src_file->type == DAOS || mfu_dst_file->type == DAOS) {
+        MFU_LOG(MFU_LOG_ERR, "DAOS API does not support --delete.");
+        return 1;
+    }
+
+    return 0;
+}
+
+/* Perform a DAOS object sync */
+static int dsync_daos(
```
dsikich, line 2911: @daltonbohning confused why other functions are named with mfu_daos_ but these all start with dsync_? Is there a reason for this?

src/dcp/dcp.c
```diff
@@ -451,36 +448,25 @@ int main(int argc, char** argv)
...
         }
 
-        /* all objects are on rank 0 at this point,
-         * evenly spread them among the ranks */
-        mfu_flist newflist = mfu_flist_spread(flist);
-
-        /* perform copy after oids are spread evenly across all ranks */
-        tmp_rc = mfu_flist_copy_daos(daos_args, newflist);
+        /* Collectively copy all objects */
+        tmp_rc = mfu_daos_flist_sync(daos_args, flist, false, true);
```
dsikich, line 469: @daltonbohning maybe I missed it, but in this call I am not seeing where the destination gets updated for the copy after the recx are listed using this function?

src/common/mfu_daos.c
```diff
@@ -919,112 +1056,281 @@ int daos_cleanup(
...
+                        break;
+                    }
+                }
             }
+            free_iov_buf(number, dst_buf);
+        }
 
-            /* update fetched recx values and place in destination object */
+        /* Conditionally write to the destination */
+        if (write_dst && !recx_equal) {
```
dsikich, line 1299: @daltonbohning just trying to make sure I understand the code path for the copy correctly, which uses this function it looks like. So for the case of dcp the compare_dst is false, and the write_dst is on, so this is the update to the destination used in that case?

src/common/mfu_daos.c
dsikich: @daltonbohning never mind, I went back and re-read our messages about this, so I see this has to do with whether or not the function returns positive or negative error codes.

daltonbohning: @dsikich My thinking here was to use `mfu_dfs` since this is a wrapper function specifically for the DFS API.
Right, the idea is that we can know what sort of error code to expect.
Although... this function doesn't return the errno code from DFS. It generically returns -1 on error.
I'm on the fence about using `mfu_dfs` vs `mfu_daos` here.
What do you think?

src/dsync/dsync.c
daltonbohning: @dsikich Really just a question of namespaces. Since these functions are defined within `dsync.c` I thought it made sense to give them a `dsync_` prefix, particularly because they are only relevant in the context of `dsync`.
However, I'm open to criticism :)

src/common/mfu_daos.c
dsikich: @daltonbohning I guess the more I think about it I like using dfs_ in the name for functions that use that API, since we have a lot of functionality now that doesn't use DFS.

src/dcp/dcp.c
daltonbohning: @dsikich I'm not sure I understand the question. `mfu_daos_flist_sync` will iterate over the `recxs` and copy to the destination.

dsikich: @daltonbohning yeah, I guess I was looking for the call to obj_update on the destination handle but I maybe missed seeing it.

src/common/mfu_daos.c
daltonbohning: @dsikich Right. for `dcp`, we do _**not**_ compare the src and dst, but we _**do**_ update the dst. Which really just means "read src and write to dst".
The reason for the argument `write_dst` is that I was hoping we could perhaps reuse this function for `dcmp` by passing:
`compare_dst=true, write_dst=false`

daltonbohning: @dsikich Something we could consider is abstracting each action into a separate function. E.g a function for "compare" and a function for "write". Maybe then the code path would make more sense without all of the clutter in a single function.

src/common/mfu_daos.c
daltonbohning: @dsikich Yeah, and we could always change this later if needed. This function is `static`, so it's easy enough to rename it

src/dcp/dcp.c
daltonbohning: @dsikich Ah, yes. I think maybe I should abstract the `mfu_daos_flist_sync_recx_array` function into multiple smaller functions so it's not so cluttered.

src/common/mfu_daos.c
dsikich: @daltonbohning yeah, I think that is my main comment about the code. I think it would be a bit easier to understand if the cmp/copy functionality was separate. And I think calling a sync_ function in dcp threw me off a bit at first. I think putting the cmp and sync functionality together makes sense. I know some of this stuff runs together and it makes sense to be able to reuse some of it.

src/dsync/dsync.c
dsikich: @daltonbohning was just curious since I don't think we have any functions that begin with dcp_ but it is just a helper function in dsync, so I don't think it is that big a deal since it is clear it is a daos function.

dsikich (approved):
@daltonbohning Thanks, this is nice to have now. Adding in the stats for the objects is really nice too!

src/common/mfu_daos.c
daltonbohning: @dsikich I agree. There are some other performance improvements that need to be made in this function as well, so I'll include this refactor as part of that (once this initial PR is merged).

adammoody (approved):
This looks good, too.  Thanks @daltonbohning .  Having dsync available in DAOS will be great!

daltonbohning:
Thanks @adammoody @dsikich !

PR #459, opened 2021-02-09
daos: copy hdf5 containers to and from POSIX filesystems

dsikich:
If daos-api=hdf5, then h5repack is invoked. This is for copying to/from DAOS HDF5 containers. If the daos-api flag is not set, then the default is to move hdf5 files to DAOS POSIX containers. Also, h5repack is serial, so if more than one rank is specified when an HDF5 container is the source or destination the program will exit with an error explaining only one process is supported currently.

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -7,6 +7,7 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
 #include <unistd.h>
 
 #include <daos_fs.h>
@@ -572,7 +573,6 @@ static int daos_set_api_compat(
         }
         mfu_src_file->type = DAOS;
     }
-
     return 0;
 }
 
@@ -882,6 +882,8 @@ int daos_parse_api_str(
         *api = DAOS_API_DFS;
     } else if (strcasecmp(api_str, "DAOS") == 0) {
         *api = DAOS_API_DAOS;
+    } else if (strcasecmp(api_str, "HDF5") == 0) {
+        *api = DAOS_API_HDF5;
     } else {
         rc = 1;
     }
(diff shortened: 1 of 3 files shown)


src/common/mfu_daos.c
```diff
@@ -1313,3 +1315,120 @@ int mfu_flist_copy_daos(
...
+            if (dst_cont_uuid[0] != '\0') {
+                MFU_LOG(MFU_LOG_INFO, "Successfully copied to DAOS HDF5 container: %s",
+                        dst_cont_uuid);
+            } else {
+                MFU_LOG(MFU_LOG_INFO, "Successfully copied to HDF5 file: %s",
+                        argpaths[1]);
+            }
+        }
+        mfu_free(&args[3]);
+    }
```
daltonbohning, line 1432: @dsikich Else case for when NEITHER the src or dst is HDF5?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1315,120 @@ int mfu_flist_copy_daos(
...
+            args[4] = NULL; 
+        } else if (mfu_dst_file->type == HDF5) {
+            args[1] = "--src-vol-name=native";
+            args[2] = strdup(argpaths[0]);
+            args[3] = strdup(dst_cont_uuid);
+            args[4] = NULL; 
+        } else {
+            MFU_LOG(MFU_LOG_ERR, "Error parsing DAOS HDF5 arguments"
+                    MFU_ERRF, MFU_ERRP(-MFU_ERR_INVAL_ARG));
+            mfu_free(&args[3]);
```
daltonbohning, line 1384: @dsikich We also need to free args[2]

src/common/mfu_daos.c
```diff
@@ -1313,3 +1315,120 @@ int mfu_flist_copy_daos(
...
+        /* Child process */
+        if ((child = fork()) == 0) {
+
+            /* start h5repack */
+            execvp(args[0], args);
+
+            /* if child process reaches this point then execvp must have failed */
+            MFU_LOG(MFU_LOG_ERR, "execvp on h5repack failed: "
+                    MFU_ERRF, MFU_ERRP(-MFU_ERR_DCP));
+            mfu_free(&args[3]);
```
daltonbohning, line 1398: @dsikich We also need to free args[2]

src/common/mfu_daos.c
```diff
@@ -1313,3 +1315,120 @@ int mfu_flist_copy_daos(
...
+            rc = 1;
+            return rc;
+        /* Parent process */
+        } else {
+            if (child == (pid_t)(-1)) {
+                MFU_LOG(MFU_LOG_ERR, "fork failed: "
+                        MFU_ERRF, MFU_ERRP(-MFU_ERR_DCP));
+                mfu_free(&args[3]);
```
daltonbohning, line 1406: @dsikich We also need to free args[2]

src/common/mfu_daos.c
```diff
@@ -1313,3 +1315,120 @@ int mfu_flist_copy_daos(
...
+                /* parent will wait for child to complete */
+                child_pid = wait(&child_status);
+            }
+        }
+
+        /* if child status is 0, then copy was successful */
+        if (child_status != 0) {
+            MFU_LOG(MFU_LOG_ERR, "errors occurred while copying to/from container: "
+                    MFU_ERRF, MFU_ERRP(-MFU_ERR_DCP));
+            mfu_free(&args[3]);
```
daltonbohning, line 1419: @dsikich We also need to free args[2]

src/common/mfu_daos.c
```diff
@@ -1313,3 +1315,120 @@ int mfu_flist_copy_daos(
...
+            rc = 1;
+            return rc;
+        } else {
+            if (dst_cont_uuid[0] != '\0') {
+                MFU_LOG(MFU_LOG_INFO, "Successfully copied to DAOS HDF5 container: %s",
+                        dst_cont_uuid);
+            } else {
```
daltonbohning, line 1426: @dsikich Could we instead just check `mfu_dst_file->type`?

src/common/mfu_daos.c
```diff
@@ -1313,3 +1315,120 @@ int mfu_flist_copy_daos(
...
+                MFU_LOG(MFU_LOG_INFO, "Successfully copied to HDF5 file: %s",
+                        argpaths[1]);
+            }
+        }
+        mfu_free(&args[3]);
```
daltonbohning, line 1431: @dsikich We also need to free args[2]

src/common/mfu_daos.c
```diff
@@ -331,6 +332,8 @@ static int daos_set_api_cont_type(
 
     if (cont_type == DAOS_PROP_CO_LAYOUT_POSIX) {
         mfu_file->type = DFS;
+    } else if (cont_type == DAOS_PROP_CO_LAYOUT_HDF5) {
+        mfu_file->type = HDF5;
```
daltonbohning, line 336: @dsikich Just a note that we'll need to incorporate DAOS_PROP_CO_LAYOUT_HDF5 into the dest logic rework. I don't expect there to be any "conflicts", but some of the logic might need some adjustments.

daltonbohning:
This PR should be merged after #457

src/common/mfu_daos.c
```diff
@@ -1313,3 +1315,120 @@ int mfu_flist_copy_daos(
...
+    }
+
+    if (mfu_src_file->type == HDF5 || mfu_dst_file->type == HDF5) {
+        char src_cont_uuid[130];
+        char dst_cont_uuid[130];
+        pid_t child;
+        int child_status;
+        pid_t child_pid;
+        char *args[4];
+        args[0] = "h5repack";
```
daltonbohning, line 1345: @dsikich After the serialization changes, we can use
```suggestion
        args[0] = HDF5_BIN_DIR"/h5repack";
```

src/dcp/dcp.c
```diff
@@ -364,7 +364,21 @@ int main(int argc, char** argv)
         MPI_Finalize();
         return 1;
     }
-    
+   
+    /* check if source or destination is an HDF5 container, if that is
+     * the case then fork/exec h5repack */
+    rc = mfu_daos_hdf5_copy(mfu_src_file, mfu_dst_file, argpaths, daos_args);
```
daltonbohning, line 370: @dsikich We should only call this if either the source or dst is HDF5. Otherwise, we'll never get to the POSIX or DAOS copies

src/common/mfu_daos.c
```diff
@@ -1313,3 +1315,120 @@ int mfu_flist_copy_daos(
...
+            rc = 1;
+            return rc;
+        } else {
+            if (dst_cont_uuid[0] != '\0') {
+                MFU_LOG(MFU_LOG_INFO, "Successfully copied to DAOS HDF5 container: %s",
+                        dst_cont_uuid);
+            } else {
+                MFU_LOG(MFU_LOG_INFO, "Successfully copied to HDF5 file: %s",
+                        argpaths[1]);
+            }
```
daltonbohning, line 1429: @dsikich I think this is more explicit. What do you think?
```suggestion
            if (mfu_dst_file->type == HDF5) {
                MFU_LOG(MFU_LOG_INFO, "Successfully copied to DAOS HDF5 container: %s",
                        dst_cont_uuid);
            } else {
                MFU_LOG(MFU_LOG_INFO, "Successfully copied to HDF5 file: %s",
                        argpaths[1]);
            }
```

src/common/mfu_daos.c
daltonbohning, line 1431: @dsikich
```suggestion
        mfu_free(&args[2]);
        mfu_free(&args[3]);
```

src/dcp/dcp.c
```diff
@@ -364,7 +364,21 @@ int main(int argc, char** argv)
...
+    if (rc != 0) {
+        mfu_finalize();
+        MPI_Finalize();
+        return 1;
+    } else {
+        /* h5repack was successful, parent process can return  */
+        mfu_finalize();
+        MPI_Finalize();
+        return 0;
+    }
```
daltonbohning, line 380: @dsikich We should only call this if either the source or dst is HDF5. Otherwise, we'll never get to the POSIX or DAOS copies. Also, maybe we can simplifythe if/else statement.
```suggestion
    if (mfu_src_file->type == HDF5 || mfu_dst_file->type == HDF5) {
        rc = mfu_daos_hdf5_copy(mfu_src_file, mfu_dst_file, argpaths, daos_args);
        if (rc != 0) {
            rc = 1;
        }
        mfu_finalize();
        MPI_Finalize();
        return rc;
    }
```

daltonbohning (commented):
@dsikich Updated my previous comments to use commit suggestions, at your discretion.

src/common/mfu_daos.c
```diff
@@ -1313,3 +1317,102 @@ int mfu_flist_copy_daos(
...
+        } else {
+            /* this must be a DAOS->DAOS copy */
+            args[1] = "";
+            args[2] = strdup(argpaths[0]);
+            args[3] = strdup(argpaths[1]);
+            args[4] = NULL;
+        }
+
+        /* Child process */
+        if ((child = fork()) == 0) {
```
adammoody, line 1376: Just a heads up that some MPI libraries do a poor job supporting fork() calls.  That means some users may encounter problems if their MPI library or their system happens to be one that has trouble with fork.  On the first IBM BlueGene systems, fork() wasn't an option at all.  On Infiniband networks, users may need to ``export IBV_FORK_SAFE=1`` to get things to work properly.

I don't know of a good substitute that can be used instead.  Just something to be aware of.  Perhaps add a note in our documentation where it would come into play.

adammoody: Oh, if HDF offers an API/library for this, that might be better than invoking their binary via fork().

daltonbohning: This is good to know. Whatever we ultimately end up doing, I'll be sure to run tests with the different systems and configurations that DAOS supports.

src/common/mfu_daos.c
dsikich: @daltonbohning I think that h5repack should handle creating the destination. I am about to test some of this but this should work once  
DAOS-6574 duns: some bug fixes and features to DUNS is merged into DAOS.

dsikich:
Leaving a note here that the functionality in this PR is only supported if commit dc6306dbed9698d9f6e218c5b3ec15c7f9d9044c from the HDF5 DAOS VOL is present. Also, this PR needs commit 506a8bda1ce81dd039e10e809f716c7f2d544a9b for DAOS.

src/common/mfu_daos.c
dsikich: @daltonbohning after testing, h5repack does not accept daos://$pool/ without a generated UUID, so I think we will just need to generate a UUID for the destination in this case like we do the other tools.

dsikich: @daltonbohning but we do not need to create the container

daltonbohning: @dsikich I imagine that's probably because duns_resolve_path currently fails for daos://pool.
I wonder if Mohamad has plans to support daos://pool after his DAOS PR? https://github.com/daos-stack/daos/pull/4798#event-4414016922
But I guess worst case, we can just call duns_resolve_path, check for missing container, generate uuid if needed, rebuild the daos:// path, and pass to h5repack

dsikich: @daltonbohning that is the PR I was using to run the h5repack, so daos://$pool/ will not work for h5repack even using it I guess. So we will still need to generate the cont UUID to pass to h5repack, and that handles container creation.

daltonbohning: @dsikich As a reference, I have similar logic for doing this in my PR:
https://github.com/hpc/mpifileutils/pull/468/commits/2cfc511d5c1bd185a2e415d826954cc1ab3991e9
(scroll to line ~420)

For now, I would say to just do what you need to and I can handle any conflicts on my end, since my PR doesn't contain "necessary" changes.

dsikich:
@daltonbohning I ran this with the new changes to hdf5 and DAOS and it seems to be working. I removed any code that connected/created/cleaned up pools or containers, as h5repack should handle that. This can be merged after daos-stack/daos#4798 lands, which looks like it should be soon.

src/common/mfu_daos.c
```diff
@@ -4766,4 +4768,176 @@ int daos_cont_deserialize_hdlr(int rank, daos_args_t *da, const char *h5filename
...
+
+    /* check for DAOS path and set VOL Name accordingly */
+    rc = duns_resolve_path(argpaths[0], &src_dattr);
+    if (rc == 0) {
+        src_daos = true;
+        uuid_copy(daos_args->src_pool_uuid, src_dattr.da_puuid);
+        uuid_copy(daos_args->src_cont_uuid, src_dattr.da_cuuid);
+        if (!daos_uuid_valid(daos_args->src_cont_uuid)) {
+            uuid_generate(daos_args->src_cont_uuid);
+        }
```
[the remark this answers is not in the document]
dsikich: @daltonbohning yeah, good catch. I just meant to have this for the destination container.

daltonbohning (approved):
@dsikich Just minor things. Otherwise, LGTM.

src/common/mfu_daos.c
```diff
@@ -4766,4 +4768,173 @@ int daos_cont_deserialize_hdlr(int rank, daos_args_t *da, const char *h5filename
...
+    } else {
+        /* destination path is POSIX */
+        rc = 0;
+    }
+
+    pid_t child;
+    int child_status;
+    pid_t child_pid;
+    char *args[4];
+    args[0] = "h5repack";
```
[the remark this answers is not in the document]
dsikich: @daltonbohning yes, I agree.

src/common/mfu_daos.c
```diff
@@ -973,7 +975,7 @@ int daos_setup(
     bool have_src_cont  = daos_uuid_valid(da->src_cont_uuid);
     bool have_dst_pool  = daos_uuid_valid(da->dst_pool_uuid);
     bool have_dst_cont  = daos_uuid_valid(da->dst_cont_uuid);
-
+
```
adammoody, line 978: stray spaces or tab here?

src/common/mfu_daos.c
```diff
@@ -4766,4 +4768,176 @@ int daos_cont_deserialize_hdlr(int rank, daos_args_t *da, const char *h5filename
...
+            MFU_LOG(MFU_LOG_INFO, "Successfully copied to %s", argpaths[1]);
+        }
+    }
+
+    if (src_daos && dst_daos) {
+        mfu_free(&args[1]);
+    } else {
+        mfu_free(&args[3]);
+    }
+    mfu_free(&args[2]);
```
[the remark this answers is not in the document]
adammoody: Yeah, I was thinking of how to simplify this too.  Another option is to just always strdup and always free which you can do so long as all pointers are initialize to NULL if not being used:
```
args[1] = strdup("--dst-vol-name=native");
...
mfu_free(&args[1]);
mfu_free(&args[2]);
mfu_free(&args[3]);
mfu_free(&args[4]);
```
That works, because this is valid:
```
args[4] = NULL;
mfu_free(&args[4]);
```

src/dcp/dcp.c
```diff
@@ -415,7 +422,7 @@ int main(int argc, char** argv)
             mfu_param_path_free_all(numpaths, paths);
             mfu_free(&paths);
 #ifdef DAOS_SUPPORT
-	    daos_cleanup(daos_args, mfu_src_file, mfu_dst_file);
+	        daos_cleanup(daos_args, mfu_src_file, mfu_dst_file);
```
adammoody, line 425: Maybe a tab in here?

adammoody (approved):

adammoody:
This looks good.  I made a couple comments, but nothing to hold up the merge.  Thanks.

src/common/mfu_daos.c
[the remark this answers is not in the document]
daltonbohning: Ah, yes. I think that's even simpler!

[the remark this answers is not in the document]
dsikich: @adammoody @daltonbohning ok, I updated this

dsikich:
> This looks good. I made a couple comments, but nothing to hold up the merge. Thanks.

@adammoody thanks!

PR #460, opened 2021-02-11
serialize and deserialize almost all props

daltonbohning:


src/common/mfu_daos.c
```diff
@@ -16,6 +16,9 @@
 
 #ifdef HDF5_SUPPORT
 #include <hdf5.h>
+#if H5_VERS_MAJOR == 1 && H5_VERS_MINOR < 12
+#define H5Sencode1 H5Sencode
+#endif
```
daltonbohning, line 21: @dsikich Please let me know if this breaks anything for you

src/common/mfu_daos.c
```diff
@@ -2434,21 +2441,27 @@ static int cont_serialize_prop_opaque(struct hdf5_args *hdf5,
...
 {
     int rc = 0;
     hid_t status = 0;
 
+    if (entry == NULL || entry->dpe_str == NULL) {
+        MFU_LOG(MFU_LOG_ERR, "Property %s not found", prop_str);
+        rc = 1;
+        goto out;
+    }
+
```
daltonbohning, line 2456: @dsikich I found that printing the error here simplified the caller

src/common/mfu_daos.c
```diff
@@ -2556,38 +2575,160 @@ static int cont_serialize_props(struct hdf5_args *hdf5,
...
+    prop_query->dpp_entries[5].dpe_type = DAOS_PROP_CO_CSUM_SERVER_VERIFY;
+    prop_query->dpp_entries[6].dpe_type = DAOS_PROP_CO_REDUN_FAC;
+    prop_query->dpp_entries[7].dpe_type = DAOS_PROP_CO_REDUN_LVL;
+    prop_query->dpp_entries[8].dpe_type = DAOS_PROP_CO_SNAPSHOT_MAX;
+    prop_query->dpp_entries[9].dpe_type = DAOS_PROP_CO_COMPRESS;
+    prop_query->dpp_entries[10].dpe_type = DAOS_PROP_CO_ENCRYPT;
+    prop_query->dpp_entries[11].dpe_type = DAOS_PROP_CO_OWNER;
+    prop_query->dpp_entries[12].dpe_type = DAOS_PROP_CO_OWNER_GROUP;
+    prop_query->dpp_entries[13].dpe_type = DAOS_PROP_CO_DEDUP;
+    prop_query->dpp_entries[14].dpe_type = DAOS_PROP_CO_DEDUP_THRESHOLD;
```
daltonbohning, line 2605: @dsikich I chose to "hardcode" the indexes so it is more explicit when referenced below, since we can't use `daos_prop_entry_get`

src/common/mfu_daos.c
```diff
@@ -2556,38 +2575,160 @@ static int cont_serialize_props(struct hdf5_args *hdf5,
...
+
+    /* TODO DAOS these as well? */
+    //prop_query->dpp_entries[15].dpe_type = DAOS_PROP_CO_STATUS;
+    //prop_query->dpp_entries[16].dpe_type = DAOS_PROP_CO_ALLOCED_OID;
```
daltonbohning, line 2609: @dsikich These might be new since you started the serialization work

src/common/mfu_daos.c
```diff
@@ -2556,38 +2575,160 @@ static int cont_serialize_props(struct hdf5_args *hdf5,
...
 
-    prop_query->dpp_entries[0].dpe_type = DAOS_PROP_CO_LAYOUT_TYPE;
     rc = daos_cont_query(cont, NULL, prop_query, NULL);
     if (rc != 0) {
-        MFU_LOG(MFU_LOG_ERR, "failed to query container: %d", rc);
+        MFU_LOG(MFU_LOG_ERR, "failed to query container "DF_RC, DP_RC(rc));
```
daltonbohning, line 2613: @dsikich Notice we can use DF_RC and DP_RC for printing a -DER

src/common/mfu_daos.c
```diff
@@ -2556,38 +2575,160 @@ static int cont_serialize_props(struct hdf5_args *hdf5,
...
+    if (rc && rc != -DER_NO_PERM) {
+        MFU_LOG(MFU_LOG_ERR, "failed to query container ACL "DF_RC, DP_RC(rc));
         rc = 1;
         goto out;
     }
-    printf("cont_str: %s\n", cont_str);
-    rc = cont_serialize_prop_uint(hdf5, &prop_query->dpp_entries[0].dpe_val,
-                                  cont_str);
+
+    /* TODO DAOS maybe pass static strings to cont_serialize_prop_str, etc? */
```
daltonbohning, line 2626: @dsikich If we just pass a static string like this:
```
cont_serialize_prop_str(hdf5, entry, "DAOS_PROP_CO_LABEL")
```
Then we don't need the local `cont_str` at all and we don't need to use `strncpy`.
But maybe there was a reason why you chose to not do it this way?

daltonbohning:
@dsikich I didn't see any logic for deserializing the properties. Maybe I overlooked that somewhere?

dsikich:
@daltonbohning right now I am deserializing the cont layout and the version. Deserializing is quite a bit simpler. You can find the cont layout deserialization in cont_deserialize_hdlr.

dsikich:
@daltonbohning The properties get set when the container is created

daltonbohning:
> cont_deserialize_hdlr

@dsikich Not sure how I missed that before... I'll add in the rest of the deserialization. Thanks!

src/common/mfu_daos.c
dsikich: @daltonbohning no I started doing it that way, then realized I did not need it, and can be removed.

src/common/mfu_daos.c
dsikich: @daltonbohning yeah those are newer ones

src/common/mfu_daos.c
dsikich: @daltonbohning yeah, that makes sense since we cannot use those.

src/common/mfu_daos.c
daltonbohning: @dsikich I'll add them as well

src/common/mfu_daos.c
daltonbohning: @dsikich Great, I'll simplify this then. Thanks!

daltonbohning:
@dsikich I added the deserialization, but I wasn't able to properly test because of the container create issue we talked about before.
Also, we could probably simplify the serialization and deserialization of the properties by just looping over each property and using a `case` statement to decide whether to use `str`, `uint` or `opaque`. Maybe we can think about that once it is functionally working.

PR #468, opened 2021-03-08
daos updates

daltonbohning:


daltonbohning:
@dsikich This is ready for review now. Particularly, I would really appreciate feedback on commit https://github.com/hpc/mpifileutils/pull/468/commits/e29062142cbccd402729d080813c35d38f8a4d81. The idea is that we start with a static buffer to avoid unnecessary allocations. But we double the size of the buffer if we get DER_KEY2BIG. The real tricky part is that you can't `free` a static buffer, so we have to sort of tip-toe around that. Please let me know if you have a simpler/cleaner approach than what I implemented here.

An alternative is that maybe we could just have the CALLER allocate a buffer and pass the buffer and size to this function. That way, we could keep re-using the same buffer over and over (and the static vs alloc isn't as big of a deal, because we aren't always allocating EVERY time). But a repercussion is that if we allocate a large buffer at any point, then the buffer will remain large until the copy operation is complete (because the caller passes the same buffer). But maybe that's okay. And maybe in some cases if any of the keys are large for a particular object, then maybe other objects have large keys too? It's really a tradeoff between space/time.

daltonbohning:
Tabling this PR for now until it is better thought out

PR #469, opened 2021-03-09
daos serialize: decrease chunk size

daltonbohning:
Decrease the chunk size from 128MB to 1K.
This greatly reduces the space requirements.

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -2289,7 +2289,10 @@ static int init_recx_data(struct hdf5_args *hdf5)
     hdf5->single_dims[0] = 1;
     hdf5->rx_dims[0] = 0;
     hdf5->rx_max_dims[0] = H5S_UNLIMITED;
-    hdf5->rx_chunk_dims[0] = 131072;
+
+    /* TODO consider other chunk sizes or possibly use different
+     * chunk sizes for different dkeys/akeys. */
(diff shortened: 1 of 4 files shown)


dsikich (approved):

src/common/mfu_daos.c
```diff
@@ -109,48 +103,73 @@ static void mfu_daos_copy_stats_print(mfu_daos_copy_stats_t* stats)
...
+        MFU_LOG(MFU_LOG_INFO, "Bytes written : %.3lf %s (%" PRId64 " bytes)",
+                write_size_tmp, write_size_units, stats->bytes_written);
+    }
+
+    if (print_read_rate) {
+        /* Compute read rate */
+        double read_rate = 0;
+        if (stats->bytes_read > 0) {
+            read_rate = (double) stats->bytes_read / rel_time;
+        }
```
adammoody, line 132: It's not likely, but I always get paranoid about a possible divide by zero in places like this.  To be safe, you can add in a ``rel_time > 0`` check.

daltonbohning: Oh, that's actually what I _**meant**_ to do to begin with:
```
if (rel_tim > 0) {
    read_rate = (double) stats->bytes_read / rel_time;
}
```
Since it doesn't matter if `bytes_read` if 0. (that would just be 0)
Thanks for noticing this!

PR #470, opened 2021-03-09
daos: simplify and enhance path parsing

daltonbohning:
Fixed a bug in daos_parse_path where -1 would
be returned for a POSIX path, but should be 1.

Simplified daos_parse_path to always require the daos://
prefix for daos paths.

Remove custom pool uuid parsing, which is now handled by
duns_resolve_path.

Improve some path free's.


dsikich (approved):
LGTM. Looks like this simplifies the parsing a good amount!

adammoody (approved):
Looks good.  Thanks, @daltonbohning

PR #472, opened 2021-03-16
daos: use internal IO functions

daltonbohning:
- Replace the dfs_* IO calls with the new dfs_sys_* API.
- Return ENOSYS when not DAOS_SUPPORT.
- Added mfu_errno2rc to simplify returns from dfs_* calls.

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -1165,12 +1165,18 @@ static int mfu_dfs_mount(
   daos_handle_t* poh,
   daos_handle_t* coh)
 {
-    /* Mount dfs */
-    int rc = dfs_mount(*poh, *coh, O_RDWR, &mfu_file->dfs);
-    if (rc != 0) {
-        MFU_LOG(MFU_LOG_ERR, "Failed to mount DAOS filesystem (DFS): "
-                MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS));
+    /* Mount dfs_sys */
+    int rc = dfs_sys_mount(*poh, *coh, O_RDWR, DFS_SYS_NO_LOCK, &mfu_file->dfs_sys);
+    if (rc !=0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to mount DAOS filesystem (DFS): %s", strerror(rc));
         rc = -1;
+    }
+    rc = dfs_sys2base(mfu_file->dfs_sys, &mfu_file->dfs);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to get DAOS filesystem (DFS) base: %s", strerror(rc));
+	rc = -1;
+	dfs_sys_umount(mfu_file->dfs_sys);
+	mfu_file->dfs_sys = NULL;
     }
 
     return rc;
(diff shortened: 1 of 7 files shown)


src/common/mfu_daos.c
```diff
@@ -1114,13 +1114,19 @@ static int mfu_dfs_mount(
...
-                MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS));
+    /* Mount dfs_sys */
+    int rc = dfs_sys_mount(*poh, *coh, O_RDWR, DFS_SYS_NO_LOCK, &mfu_file->dfs_sys);
+    if (rc !=0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to mount DAOS filesystem (DFS): %s", strerror(rc));
         rc = -1;
     }
+    rc = dfs_sys2base(mfu_file->dfs_sys, &mfu_file->dfs);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to get DAOS filesystem (DFS) base: %s", strerror(rc));
```
dsikich, line 1125: @daltonbohning looks like the formatting tabs/spaces got messed up here

src/common/mfu_flist_copy.c
```diff
@@ -3220,8 +3220,8 @@ mfu_file_t* mfu_file_new(void)
     mfile->fd         = -1;
 #ifdef DAOS_SUPPORT
     mfile->obj        = NULL;
+    mfile->dfs_sys    = NULL;
     mfile->dfs        = NULL;
```
dsikich, line 3224: @daltonbohning is mfile->dfs still used? It looks like it has been changed to mfile->dfs_sys

src/common/mfu_param_path.h
```diff
@@ -58,10 +58,10 @@ typedef struct {
...
 #ifdef DAOS_SUPPORT
     /* DAOS specific variables for I/O */
-    daos_off_t           offset;
-    dfs_obj_t*           obj;
-    dfs_t*               dfs;
-    struct d_hash_table* dfs_hash;
+    daos_off_t           offset;  /* file offset */
+    dfs_obj_t*           obj;     /* open object handle */
+    dfs_sys_t*           dfs_sys; /* handle for high-level file operations */
+    dfs_t*               dfs;     /* handle for lower-level file operations */
```
dsikich, line 64: @daltonbohning curious how dfs_t is still used?

dsikich:
@daltonbohning this looks ok to me. Most of it looks like it is just replacing the daos versions of the io functions with the dfs_sys library version in DAOS. Are all of the tests passing with this version?

src/common/mfu_param_path.h
daltonbohning: @dsikich It's used for any functions that either don't have a dfs_sys implementation, or for functions that a dfs_sys implementation would just be a direct wrapper for dfs. Right now, it's just truncate:
https://github.com/hpc/mpifileutils/blob/2a8e5c413c0aa15657cffc56b5173e8fef9d724f/src/common/mfu_io.c#L1120
But more generally, the idea is that we have full access to DFS, so we aren't restricted to the dfs_sys API. I.e. instead of _**replacing**_ the DFS API, the DFS Sys API is built _**on top of**_ DFS

daltonbohning:
> @daltonbohning this looks ok to me. Most of it looks like it is just replacing the daos versions of the io functions with the dfs_sys library version in DAOS. Are all of the tests passing with this version?

@dsikich Almost, with a few edge-case exceptions that will be handled internally (patch in review in DAOS).

src/common/mfu_daos.c
```diff
@@ -1530,7 +1529,7 @@ int daos_cleanup(
     MPI_Barrier(MPI_COMM_WORLD);
 
     /* Unmount destination DFS container */
-    if ((mfu_dst_file != NULL) && (mfu_dst_file->dfs != NULL)) {
```
dsikich, line 1533: @daltonbohning is the first NULL check no longer necessary?

daltonbohning: @dsikich Not sure why I did that. I think it's safer to keep the first NULL check. I'll fix that, and also add it for the src_file above

daltonbohning: @dsikich I moved the NULL checks to the helper function and also reorganized a bit so we could remove some barriers.

dsikich (approved):

daltonbohning:
@adammoody This is working and ready for review now.

adammoody (approved):
Thanks @daltonbohning and @dsikich .  Wow, that cleans up a ton of code!  That's really nice.  It's also somehow sad to see all of that first draft of the code go away, though, knowing the effort you two put into it.  Thanks!  Looks good to me.

daltonbohning:
> Thanks @daltonbohning and @dsikich . Wow, that cleans up a ton of code! That's really nice. It's also somehow sad to see all of that first draft of the code go away, though, knowing the effort you two put into it. Thanks! Looks good to me.

Thanks @adammoody ! We actually took this draft of the IO functions and created an new API out of them that can be re-used in other applications, so it wasn't a waste!

PR #473, opened 2021-03-18
daos: drm, dfind, and dwalk support

daltonbohning:
- Add support to drm for DAOS POSIX containers.
- Add support to dfind for DAOS POSIX containers.
- Add support to daos internals for a single path.

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -182,11 +182,18 @@ bool daos_uuid_valid(const uuid_t uuid)
 static int daos_check_args(
     int rank,
     char** argpaths,
+    int numpaths,
     daos_args_t* da,
     int* flag_daos_args)
 {
-    char* src_path = argpaths[0];
-    char* dst_path = argpaths[1];
+    char* src_path = NULL;
+    char* dst_path = NULL;
+    if (numpaths > 0) {
+        src_path = argpaths[0];
+    }
+    if (numpaths > 1) {
+        dst_path = argpaths[1];
+    }
 
     bool have_src_path  = src_path != NULL;
     bool have_dst_path  = dst_path != NULL;
@@ -317,9 +324,11 @@ int daos_parse_path(
 static int daos_set_paths(
   int rank,
   char** argpaths,
+  int numpaths,
   daos_args_t* da)
 {
     int     rc = 0;
+    bool    have_dst = numpaths > 1;
     bool    prefix_on_src = false;
     bool    prefix_on_dst = false;
     char*   prefix_path = NULL;
@@ -370,12 +379,14 @@ static int daos_set_paths(
             argpaths[0] = da->src_path;
         }
 
-        /* Check if the prefix matches the destination */
-        prefix_on_dst = daos_check_prefix(argpaths[1], da->dfs_prefix, &da->dst_path);
-        if (prefix_on_dst) {
-            uuid_copy(da->dst_pool_uuid, prefix_p_uuid);
-            uuid_copy(da->dst_cont_uuid, prefix_c_uuid);
-            argpaths[1] = da->dst_path;
+        if (have_dst) {
+            /* Check if the prefix matches the destination */
+            prefix_on_dst = daos_check_prefix(argpaths[1], da->dfs_prefix, &da->dst_path);
+            if (prefix_on_dst) {
+                uuid_copy(da->dst_pool_uuid, prefix_p_uuid);
+                uuid_copy(da->dst_cont_uuid, prefix_c_uuid);
+                argpaths[1] = da->dst_path;
+            }
         }
 
         if (!prefix_on_src && !prefix_on_dst) {
@@ -419,7 +430,7 @@ static int daos_set_paths(
         }
     }
 
-    if (!prefix_on_dst) {
+    if (have_dst && !prefix_on_dst) {
         size_t dst_len = strlen(argpaths[1]);
         dst_path = strndup(argpaths[1], dst_len);
         if (dst_path == NULL) {
@@ -521,6 +532,8 @@ static int daos_set_api_compat(
     daos_args_t* da,
     char** argpaths)
 {
+    bool have_dst = mfu_dst_file != NULL;
+
     /* Check whether we have pool/cont uuids */
     bool have_src_pool  = daos_uuid_valid(da->src_pool_uuid);
     bool have_src_cont  = daos_uuid_valid(da->src_cont_uuid);
@@ -554,13 +567,15 @@ static int daos_set_api_compat(
             MFU_LOG(MFU_LOG_ERR, "Cannot use path with non-POSIX container.");
             return 1;
         }
-        if (mfu_dst_file->type != DAOS && mfu_dst_file->type != DFS) {
-            MFU_LOG(MFU_LOG_ERR, "Cannot copy non-POSIX container outside DAOS.");
-            return 1;
+        if (have_dst) {
+            if (mfu_dst_file->type != DAOS && mfu_dst_file->type != DFS) {
+                MFU_LOG(MFU_LOG_ERR, "Cannot copy non-POSIX container outside DAOS.");
+                return 1;
+            }
+            mfu_dst_file->type = DAOS;
         }
-        mfu_dst_file->type = DAOS;
     }
-    if (mfu_dst_file->type == DAOS) {
+    if (have_dst && mfu_dst_file->type == DAOS) {
         if (!dst_path_is_root || !src_path_is_root) {
             MFU_LOG(MFU_LOG_ERR, "Cannot use path with non-POSIX container.");
             return 1;
@@ -906,11 +921,13 @@ int daos_parse_epc_str(
 int daos_setup(
     int rank,
     char** argpaths,
+    int numpaths,
     daos_args_t* da,
     mfu_file_t* mfu_src_file,
     mfu_file_t* mfu_dst_file)
 {
     int tmp_rc;
+    bool have_dst = (numpaths > 1) && (mfu_dst_file != NULL);
 
     /* Each process keeps track of whether it had any DAOS errors.
      * If there weren't any daos args, then ignore daos_init errors.
@@ -928,7 +945,7 @@ int daos_setup(
 
     /* Do a preliminary check on the DAOS args */
     if (!local_daos_error) {
-        tmp_rc = daos_check_args(rank, argpaths, da, &flag_daos_args);
+        tmp_rc = daos_check_args(rank, argpaths, numpaths, da, &flag_daos_args);
         if (tmp_rc != 0) {
             MFU_LOG(MFU_LOG_ERR, "Invalid DAOS args: "
                     MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS_INVAL_ARG));
@@ -941,7 +958,7 @@ int daos_setup(
      * prefix since the path is mapped to the root
      * of the container in the DAOS DFS mount */
     if (!local_daos_error) {
-        tmp_rc = daos_set_paths(rank, argpaths, da);
+        tmp_rc = daos_set_paths(rank, argpaths, numpaths, da);
         if (tmp_rc != 0) {
             MFU_LOG(MFU_LOG_ERR, "Invalid DAOS args: "
                     MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS_INVAL_ARG));
@@ -951,7 +968,7 @@ int daos_setup(
 
     /* Re-check the required DAOS arguments (if any) */
     if (!local_daos_error) {
-        tmp_rc = daos_check_args(rank, argpaths, da, &flag_daos_args);
+        tmp_rc = daos_check_args(rank, argpaths, numpaths, da, &flag_daos_args);
         if (tmp_rc != 0) {
             MFU_LOG(MFU_LOG_ERR, "Invalid DAOS args: "
                     MFU_ERRF, MFU_ERRP(-MFU_ERR_DAOS_INVAL_ARG));
@@ -1071,11 +1088,13 @@ int daos_setup(
             local_daos_error = true;
             goto out;
         }
-        /* Set the dst api based on the container type */
-        tmp_rc = daos_set_api_cont_type(mfu_dst_file, da->dst_cont_type, da->api);
-        if (tmp_rc != 0) {
-            local_daos_error = true;
-            goto out;
+        if (have_dst) {
+            /* Set the dst api based on the container type */
+            tmp_rc = daos_set_api_cont_type(mfu_dst_file, da->dst_cont_type, da->api);
+            if (tmp_rc != 0) {
+                local_daos_error = true;
+                goto out;
+            }
         }
     }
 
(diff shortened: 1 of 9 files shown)


src/drm/drm.c
```diff
@@ -197,28 +222,69 @@ int main(int argc, char** argv)
...
+    }
+
     /* create new mfu_file objects */
     mfu_file_t* mfu_file = mfu_file_new();
 
-    /* paths to walk come after the options */
-    int numpaths = 0;
+    char** argpaths = &argv[optind];
+
+    /* The remaining arguments are treated as paths */
```
dsikich, line 240: @daltonbohning is this going to work if multiple paths are passed in with different pools?

daltonbohning: @dsikich Whether they are in the same pool or different pools, mfu_flist_walk_param_paths would print:
"ERROR: Only one source can be specified when using DAOS".
And only the "source" mfu_file is being passed to daos_setup, so daos_setup only looks at the first path.
Ultimately, we'll probably want to be able to support a LIST of daos paths in a general sense for some of the utilities, but for now only a single daos path is supported per mfu_file.

dsikich: @daltonbohning ok, I thought it was setup for only one but was just making sure in case someone passes multiple. I asked at one point early on if we should support multiple paths, and at the time they were not interested and just thought they could run the tool multiple times for each path. It is something we can ask about again as we get into more of the refinements we've been talking about, since it doesn't really hurt to have it as an option.

daltonbohning: @dsikich I agree it would be nice to have, but we would probably need some amount of refactoring, etc. to get it working fully.

daltonbohning:
@dsikich I also added dfind support for POSIX containers and I'm going to add dwalk soon. At this point it's really just copy-paste for dfind and dwalk. I think by doing this now, we won't be as likely to cut off support for these tools if we refactor the backend.
I.e. we can see what these tools need from the backend in order to work. Turns out they really just need to call setup :)

dsikich:
> @dsikich I also added dfind support for POSIX containers and I'm going to add dwalk soon. At this point it's really just copy-paste for dfind and dwalk. I think by doing this now, we won't be as likely to cut off support for these tools if we refactor the backend.
> I.e. we can see what these tools need from the backend in order to work. Turns out they really just need to call setup :)

@daltonbohning I think I want to keep in mind that we may not want to always support every single tool. It is easy enough for the POSIX case, but does this open up that we need to support them for non-POSIX containers as well? I think we just want to keep that in mind.

daltonbohning:
> @daltonbohning I think I want to keep in mind that we may not want to always support every single tool. It is easy enough for the POSIX case, but does this open up that we need to support them for non-POSIX containers as well? I think we just want to keep that in mind.

@dsikich Yeah, I see what you mean. I initially added `drm` support because it was asked for. I know `dfind` is something that was desired in some way. And `dwalk` is really just `dfind` slimmed down. For the most part, supporting POSIX containers is just a matter of calling the existing backend.
I don't think we have any plans per se to support these for non-POSIX. But I was thinking we could just not "advertise" support for these tools until someone "formally" requests support. And there are conceptual limitations to supporting these for non-posix as well. For example, what does a "walk" really mean for non-POSIX? Would that mean listing the oids/dkeys/akeys? And even then, there is some rudimentary support through the "daos obj" tool.

adammoody:
Some of the tools already have limitations even in pure POSIX land.  For example, some tools only partially work on NFS, like dtar and dbz2, and some tools are specific to Lustre like dstripe.  We could create a section in our documentation that lists each major storage type: Lustre, GPFS, NFS, DAOS and then lists the supported tools and partially supported tools for each one.  We could either spell out limitations there, or just say something like "see man pages for details" if it is well documented.

daltonbohning:
> Some of the tools already have limitations even in pure POSIX land. For example, some tools only partially work on NFS, like dtar and dbz2, and some tools are specific to Lustre like dstripe. We could create a section in our documentation that lists each major storage type: Lustre, GPFS, NFS, DAOS and then lists the supported tools and partially supported tools for each one. We could either spell out limitations there, or just say something like "see man pages for details" if it is well documented.

That's a good point. I think it would be good if the man pages are well documented in this respect, but also some users might prefer the "cleaner" rendering of a Readme, markdown, etc. that details the support. If you have an idea of where you would want such a section to live, I can fill in the DAOS section appropriately.

dsikich:
> > @daltonbohning I think I want to keep in mind that we may not want to always support every single tool. It is easy enough for the POSIX case, but does this open up that we need to support them for non-POSIX containers as well? I think we just want to keep that in mind.
> 
> @dsikich Yeah, I see what you mean. I initially added `drm` support because it was asked for. I know `dfind` is something that was desired in some way. And `dwalk` is really just `dfind` slimmed down. For the most part, supporting POSIX containers is just a matter of calling the existing backend.
> I don't think we have any plans per se to support these for non-POSIX. But I was thinking we could just not "advertise" support for these tools until someone "formally" requests support. And there are conceptual limitations to supporting these for non-posix as well. For example, what does a "walk" really mean for non-POSIX? Would that mean listing the oids/dkeys/akeys? And even then, there is some rudimentary support through the "daos obj" tool.



> > @daltonbohning I think I want to keep in mind that we may not want to always support every single tool. It is easy enough for the POSIX case, but does this open up that we need to support them for non-POSIX containers as well? I think we just want to keep that in mind.
> 
> @dsikich Yeah, I see what you mean. I initially added `drm` support because it was asked for. I know `dfind` is something that was desired in some way. And `dwalk` is really just `dfind` slimmed down. For the most part, supporting POSIX containers is just a matter of calling the existing backend.
> I don't think we have any plans per se to support these for non-POSIX. But I was thinking we could just not "advertise" support for these tools until someone "formally" requests support. And there are conceptual limitations to supporting these for non-posix as well. For example, what does a "walk" really mean for non-POSIX? Would that mean listing the oids/dkeys/akeys? And even then, there is some rudimentary support through the "daos obj" tool.

@daltonbohning yeah, for those tools POSIX support is easy enough, and makes sense. I just don't want to get into a situation where we commit to a lot of extra work too early. I agree that we should not plan to support the non-POSIX cases of these tools until it is formally requested.

daltonbohning:
> @daltonbohning yeah, for those tools POSIX support is easy enough, and makes sense. I just don't want to get into a situation where we commit to a lot of extra work too early. I agree that we should not plan to support the non-POSIX cases of these tools until it is formally requested.

@dsikich Right. I think that as long as we (1) don't advertise these tools as having non-POSIX support and (2) document the existing support properly, then we should be okay. At one point I mentioned this to Scott Peirce and he didn't seem to have any reservations about the non-posix support as long as it didn't require too much effort.

dsikich:
> > @daltonbohning yeah, for those tools POSIX support is easy enough, and makes sense. I just don't want to get into a situation where we commit to a lot of extra work too early. I agree that we should not plan to support the non-POSIX cases of these tools until it is formally requested.
> 
> @dsikich Right. I think that as long as we (1) don't advertise these tools as having non-POSIX support and (2) document the existing support properly, then we should be okay. At one point I mentioned this to Scott Peirce and he didn't seem to have any reservations about the non-posix support as long as it didn't require too much effort.

@daltonbohning I don't have reservations, I just think it should be very low on the list of priorities, unless it is specifically needed by someone.

daltonbohning:
> @daltonbohning I don't have reservations, I just think it should be very low on the list of priorities, unless it is specifically needed by someone. We should focus on what we currently have before adding new features/tools until it is completely stabilized.

@dsikich Sorry, I meant reservations about POSIX container support. I agree that non-posix support for these is low on the list of priorities.

dsikich:
> > @daltonbohning I don't have reservations, I just think it should be very low on the list of priorities, unless it is specifically needed by someone. We should focus on what we currently have before adding new features/tools until it is completely stabilized.
> 
> @dsikich Sorry, I meant reservations about POSIX container support. I agree that non-posix support for these is low on the list of priorities.

@daltonbohning yeah, I maybe wasn't clear enough either, that is what I meant as well. The POSIX support here makes sense to me 👍

src/common/mfu_daos.c
```diff
@@ -370,12 +385,14 @@ static int daos_set_paths(
             argpaths[0] = da->src_path;
         }
 
-        /* Check if the prefix matches the destination */
-        prefix_on_dst = daos_check_prefix(argpaths[1], da->dfs_prefix, &da->dst_path);
-        if (prefix_on_dst) {
-            uuid_copy(da->dst_pool_uuid, prefix_p_uuid);
-            uuid_copy(da->dst_cont_uuid, prefix_c_uuid);
-            argpaths[1] = da->dst_path;
+        if (have_dst) {
```
dsikich, line 388: @daltonbohning I know that dcp checks to make sure that numpaths > 2 (for a source and destination) before moving forward, but we should make sure none of the other tools that take two paths get to this point. I don't remember if dsync and dcmp error out if they have less than two paths before calling this function.

daltonbohning: @dsikich dcmp does error out before calling daos_setup:
https://github.com/hpc/mpifileutils/blob/b0d402dea9d054c5321e15f61a5f70ef8f591bf6/src/dcmp/dcmp.c#L2260-L2265

And so does dsync:
https://github.com/hpc/mpifileutils/blob/b0d402dea9d054c5321e15f61a5f70ef8f591bf6/src/dsync/dsync.c#L3190-L3203

daltonbohning: @dsikich Now I'm wondering if maybe we should more strictly enforce this. If mfu_src_file is given, then numargs should be EXACTLY 1. And if mfu_src_file and mfu_dst_file is given, then numargs should be EXACTLY 2.
ANd if mfu_src_file is not given at all then that's an error

daltonbohning: @dsikich Actually, enforcing exactly 1 or exactly 2 doesn't work for dcp and dsync, because it might just be multiple posix paths, which is valid. Anyway, I think each tool does error out if at least 1 or at least 2 paths are not given.

src/dfind/dfind.c
```diff
@@ -516,13 +538,20 @@ int main (int argc, char** argv)
     	        exit(1);
             }
     	    break;
-
+#ifdef DAOS_SUPPORT
+            case 'X':
+                daos_args->dfs_prefix = MFU_STRDUP(optarg);
+                break;
+#endif
```
dsikich, line 545: @daltonbohning its too bad we still have to keep adding this daos_prefix to more tools since we will be eventually getting rid of it

daltonbohning: @dsikich Yeah :/
I wonder if it's okay to just not add it? Do you know when we plan to have support such that we won't need it?

dsikich: @daltonbohning I think it is supposed to be in the 2.0 release but not 100 percent sure. It is probably not necessary to have it in all of these since there is a way to copy subsets using the daos:// direct format, and we will have to go back and remove it as part of the interface in the future.

daltonbohning: @dsikich Okay, I do agree with that. I removed prefix from drm, dwalk, dfind. Now we only have it in dcp, dsync, dcmp (it was already in those).
Worst case, someone will specifically request it and we can add it. But like you said, the direct format allows it as well (and is easier IMO)

dsikich: @daltonbohning yeah, I think using the prefix is awkward.

src/dfind/dfind.c
```diff
@@ -297,13 +308,20 @@ int main (int argc, char** argv)
...
-        {"output",    1, 0, 'o'},
-        {"verbose",   0, 0, 'v'},
-        {"quiet",     0, 0, 'q'},
-        {"help",      0, 0, 'h'},
+        {"input",       1, 0, 'i'},
+        {"output",      1, 0, 'o'},
+        {"text",        0, 0, 't'},
+        {"verbose",     0, 0, 'v'},
+        {"quiet",       0, 0, 'q'},
+        {"help",        0, 0, 'h'},
```
dsikich, line 324: @daltonbohning I do not see the type option added here but it is added below?

src/dfind/dfind.c
```diff
@@ -338,7 +356,7 @@ int main (int argc, char** argv)
     int usage = 0;
     while (1) {
         int c = getopt_long(
-                    argc, argv, "i:o:vqh",
+                    argc, argv, "i:o:tvqh",
```
dsikich, line 359: @daltonbohning The lowercase 't' option is added here but not 'T'?

dsikich (approved):

src/dfind/dfind.c
daltonbohning: @dsikich Right. That's because the uppercase 'T' is for --type. But the command line only takes the long option for --type, like with these:
https://github.com/hpc/mpifileutils/blob/2fc898231298484692385e12512ba8189c0c6cc8/src/dfind/dfind.c#L328-L347

The 'T' is only used internally for parsing the arguments.
Basically, if the print_usage doesn't show the short options, then the short options don't appear in this list because they aren't exposed through the command line.

dsikich: @daltonbohning oh ok it only takes the long option I missed that. thanks! LGTM.

daltonbohning: No worries :)

src/common/mfu_daos.c
```diff
@@ -317,9 +324,11 @@ int daos_parse_path(
 static int daos_set_paths(
   int rank,
   char** argpaths,
+  int numpaths,
   daos_args_t* da)
 {
     int     rc = 0;
+    bool    have_dst = numpaths > 1;
```
adammoody, line 331: Style edit for readability:
```suggestion
    bool    have_dst = (numpaths > 1);
```

src/common/mfu_daos.c
```diff
@@ -521,6 +532,8 @@ static int daos_set_api_compat(
     daos_args_t* da,
     char** argpaths)
 {
+    bool have_dst = mfu_dst_file != NULL;
```
adammoody, line 535:
```suggestion
    bool have_dst = (mfu_dst_file != NULL);
```

src/dfind/dfind.c
```diff
@@ -542,25 +562,89 @@ int main (int argc, char** argv)
...
+    if (numpaths < 1) {
+        if (rank == 0) {
+            MFU_LOG(MFU_LOG_ERR, "At least one path is needed");
+            print_usage();
+        }
+        mfu_file_delete(&mfu_file);
+        mfu_finalize();
+        MPI_Finalize();
+        return 1;
+    }
```
adammoody, line 594: Some tools can work from just an input file -- the input file replaces the need to walk a path.  I think dfind should be one of those, e.g.,
```
dwalk --output list.mfu /path/to/walk
dfind --input list.mfu --output filesonly.mfu --type f
```
I haven't checked this logic in detail, but would this require a path?

daltonbohning: You're right. This check was preventing `--input` from being used with `--output`. I instead moved a check into `daos_setup` to make sure we have enough paths there. Otherwise, the path checking falls into the pre-existing logic for each tool.
Thanks for catching this!

adammoody (approved):
Thanks @daltonbohning and @dsikich

PR #474, opened 2021-03-22
mfu_flist_walk: handle trailing slash in dirname

daltonbohning:
The main motivation for this is that DAOS often has root paths "/", since that is the root of a container. When the paths are built, this happens:
```
dir = "/"
name = "something"
path = "//something"
```
Although DAOS/DFS handles the double "//", it creates a slight performance and caching problem.
The path building was also modified to use snprintf, which is more performant. Even with the additional "/" check, this should should be _**slightly**_ faster.

### DIFF
--- a/src/common/mfu_flist_walk.c
+++ b/src/common/mfu_flist_walk.c
@@ -93,6 +93,35 @@ static void reduce_fini(const void* buf, size_t size)
     MFU_LOG(MFU_LOG_INFO, "Walked %llu items in %.3lf secs (%.3lf items/sec) ...", val, secs, rate);
 }
 
+/****************************************
+ * Global helper functions
+ ***************************************/
+
+/** Build a full path from a dirname and basename in the form:
+ * <dir> + '/' + <name> + '/0'
+ * up to path_len long.
+ * Returns 0 on success and -1 if the new path is too long. */
+static int build_path(char* path, size_t path_len, const char* dir, const char* name)
+{
+    size_t dir_len = strlen(dir);
+
+    /* Only separate with a '/' if the dir does not have a trailing slash.
+     * Builds a path to at most path_len long. */
+    int new_len;
+    if ((dir_len > 0) && (dir[dir_len - 1] == '/')) {
+        new_len = snprintf(path, path_len, "%s%s", dir, name);
+    } else {
+        new_len = snprintf(path, path_len, "%s/%s", dir, name);
+    }
+    if (new_len > path_len) {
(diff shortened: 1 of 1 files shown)


src/common/mfu_flist_walk.c
```diff
@@ -93,6 +93,35 @@ static void reduce_fini(const void* buf, size_t size)
...
+     * Builds a path to at most path_len long. */
+    int new_len;
+    if ((dir_len > 0) && (dir[dir_len - 1] == '/')) {
+        new_len = snprintf(path, path_len, "%s%s", dir, name);
+    } else {
+        new_len = snprintf(path, path_len, "%s/%s", dir, name);
+    }
+    if (new_len > path_len) {
+        MFU_LOG(MFU_LOG_ERR, "Path name is too long, %lu chars exceeds limit %lu: '%s/%s'",
+                new_len, path_len, dir, name);
```
dsikich, line 118: @daltonbohning do I have this right that this will remove an extra "/" if one is already after the directory name?

daltonbohning: @dsikich Right. Currently, this is what happens:
```
// Duplicate // in path
"/dirname/" + "/" + "filename" = "/dirname//filename"
```
But this PR changes that to:
```
// No duplicate //
"/dirname/" + "filename" = "/dirname/filename"
"/dirname" + "/" + "filename" = "/dirname/filename"
```

Realistically, this was only happening when dirname was "/" (root)

dsikich (approved):

adammoody:
Thanks.  Looks good to me.

Some related functions are in ``mfu_path_reduce()`` or ``mfu_path_strdup_reduce_str()`` both from ``common/mfu_path.c``.  Using those are probably overkill here, but this reminded me of them.  They also process out any ``.`` and ``..`` entries.

adammoody (approved):

daltonbohning:
> Thanks. Looks good to me.
> 
> Some related functions are in `mfu_path_reduce()` or `mfu_path_strdup_reduce_str()` both from `common/mfu_path.c`. Using those are probably overkill here, but this reminded me of them. They also process out any `.` and `..` entries.

Thanks @adammoody !

I agree that they are overkill, since the walk already ignores "." and ".." during readdir, and those `mfu_` functions operate on a `mfu_path` (list of paths) type, which we don't otherwise need here.

But seeing those functions does make me wonder if there is room for improvement in some way (either here or there)... Another time maybe :)

PR #476, opened 2021-04-01
daos: fix root path bug

daltonbohning:
- In `mfu_param_path_copy_dest`, don't cut the root path.
- In parse_filename, account for when dir and name are both "/",
  which can happen depending on the version of basename used.


daltonbohning:
@adammoody @dsikich Simple one here. It seems I didn't fully test #474

dsikich (approved):

adammoody (approved):

PR #477, opened 2021-04-01
serialization optimizations: write single val to akey dset

dsikich:
Instead of creating a whole dataset for single value records,
just store the single value inside of the akey dataset. For
array record types this field is set to NULL, and the length
is set to zero.

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -1950,6 +1950,7 @@ static int serialize_recx_single(struct hdf5_args *hdf5,
                                  daos_key_t *dkey,
                                  daos_handle_t *oh,
                                  daos_iod_t *iod,
+                                 uint64_t *ak_index,
                                  mfu_daos_stats_t* stats)
 {
     /* if iod_type is single value just fetch iod size from source
@@ -1960,6 +1961,8 @@ static int serialize_recx_single(struct hdf5_args *hdf5,
     d_iov_t     iov;
     int         rc;
     herr_t      err;
+    hvl_t       *single_val;
+
     buf = MFU_CALLOC(1, buf_len);
 
     /* set sgl values */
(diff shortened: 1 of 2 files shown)


src/common/mfu_daos.c
```diff
@@ -2439,17 +2437,20 @@ static int serialize_akeys(struct hdf5_args *hdf5,
...
             if ((int)iod.iod_size == 0) {
+                /* set single value field to NULL, 0 for array types */
+                (*hdf5->ak)[*ak_index].rec_single_val.p = NULL;
+                (*hdf5->ak)[*ak_index].rec_single_val.len = 0;
+
+                /* create a record dset only for array types */
+                memset(&rec_name, FILENAME_LEN, sizeof(rec_name));
+                path_len = snprintf(rec_name, FILENAME_LEN, "%lu", *ak_index);
+                if (path_len > FILENAME_LEN) {
+                    MFU_LOG(MFU_LOG_ERR, "record name too long");
```
daltonbohning, line 2449: @dsikich A couple points here:
1. rec_name is declared above as rec_name[32], and FILENAME_LEN is defined as 1024, so we might get some buffer overflow.
2. I think the memset should be:
    memset(&rec_name, 0, sizeof(rec_name));  // set all bytes to 0
3. In the future we should probably support sizes > 1024, since DAOS doesn't enforce a size restriction.

dsikich: @daltonbohning well the rec_name specifically is just the index number of the record. I don't think it has anything to do with (3). Each record dataset just needs a name, it could be anything really. I will update (2).

daltonbohning: @dsikich Ah, I see. In that case, I wonder if something like this is more "proper":
```
size_t rec_name_len = 32;
char* rec_name[rec_name_len];
...
snprintf(rec_name, rec_name_len, ...);
```
It's probably unlikely that we will overflow 32 characters with an index, but if for some reason any of these things change in the future, the buffer size is tied to the buffer itself, since really the size passed to snprintf should be the size of the buffer.

dsikich: @daltonbohning yeah I think that is probably better. I can update.

daltonbohning (approved):
LGTM. This is also passing my local, internal, serialization test case.

dsikich:
@adammoody does this look ok to you?

adammoody:
Thanks @dsikich .  I should be able to take a look by tomorrow.

dsikich:
> Thanks @dsikich . I should be able to take a look by tomorrow.

@adammoody thanks! if you have time it would help if you could review #482 as well.

adammoody (approved):

adammoody:
Looks good to me.

PR #478, opened 2021-04-05
daos: copy user attrs and cont props on create

daltonbohning:
When creating a new DAOS container, copy the user attributes
and container properties from the src container.

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -610,6 +610,300 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
     return 0;
 }
 
+/*
+ * Get the user attributes for a container in a format similar
+ * to what daos_cont_set_prop expects.
+ * The last entry is the ACL and is conditionally set only if
+ * the user has permissions.
+ * The properties should remain in the order expected by serialization.
+ * Returns 1 on error, 0 on success.
+ */
+static int cont_get_props(daos_handle_t coh, daos_prop_t** _props, bool get_oid)
+{
+    int             rc;
+    daos_prop_t*    props = NULL;
+    daos_prop_t*    prop_acl = NULL;
+    daos_prop_t*    props_merged = NULL;
+    uint32_t        num_props = 15;
+
+    if (get_oid) {
+        num_props++;
+    }
+
+    /* Allocate space for all props except ACL. */
+    props = daos_prop_alloc(num_props);
+    if (props == NULL) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to allocate container properties.");
+        rc = 1;
+        goto out;
+    }
+
+    /* The order of properties MUST match the order expected by serialization. */
+    props->dpp_entries[0].dpe_type = DAOS_PROP_CO_LABEL;
+    props->dpp_entries[1].dpe_type = DAOS_PROP_CO_LAYOUT_TYPE;
+    props->dpp_entries[2].dpe_type = DAOS_PROP_CO_LAYOUT_VER;
+    props->dpp_entries[3].dpe_type = DAOS_PROP_CO_CSUM;
+    props->dpp_entries[4].dpe_type = DAOS_PROP_CO_CSUM_CHUNK_SIZE;
+    props->dpp_entries[5].dpe_type = DAOS_PROP_CO_CSUM_SERVER_VERIFY;
+    props->dpp_entries[6].dpe_type = DAOS_PROP_CO_REDUN_FAC;
+    props->dpp_entries[7].dpe_type = DAOS_PROP_CO_REDUN_LVL;
+    props->dpp_entries[8].dpe_type = DAOS_PROP_CO_SNAPSHOT_MAX;
+    props->dpp_entries[9].dpe_type = DAOS_PROP_CO_COMPRESS;
+    props->dpp_entries[10].dpe_type = DAOS_PROP_CO_ENCRYPT;
+    props->dpp_entries[11].dpe_type = DAOS_PROP_CO_OWNER;
+    props->dpp_entries[12].dpe_type = DAOS_PROP_CO_OWNER_GROUP;
+    props->dpp_entries[13].dpe_type = DAOS_PROP_CO_DEDUP;
+    props->dpp_entries[14].dpe_type = DAOS_PROP_CO_DEDUP_THRESHOLD;
+
+    /* Conditionally get the OID. Should always be true for serialization. */
+    if (get_oid) {
+        props->dpp_entries[15].dpe_type = DAOS_PROP_CO_ALLOCED_OID;
+    }
+
+    /* Get all props except ACL first. */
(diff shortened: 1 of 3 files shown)


src/common/mfu_daos.c
```diff
@@ -610,6 +610,290 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+        MFU_LOG(MFU_LOG_ERR, "Failed to query container: "DF_RC, DP_RC(rc));
+        rc = 1;
+        goto out;
+    }
+
+    /* Fetch the ACL separately in case user doesn't have access */
+    rc = daos_cont_get_acl(coh, &prop_acl, NULL);
+    if (rc == 0) {
+        props_merged = daos_prop_merge(props, prop_acl);
+        if (props_merged == NULL) {
```
dsikich, line 666: @daltonbohning so you are merging the all into the existing property list? Do you know if this gets appended to the list as the last entry?

src/common/mfu_daos.c
```diff
@@ -3479,8 +3584,8 @@ static int cont_serialize_props(struct hdf5_args *hdf5,
     }
 
     /* serialize ACL */
-    if (prop_acl != NULL) {
-        entry = &prop_acl->dpp_entries[0];
+    if (prop_query->dpp_nr > 16) {
+        entry = &prop_query->dpp_entries[16];
```
dsikich, line 3588: @daltonbohning this doesn't totally make sense to me yet. I saw you merged the props and the prop_acl when the properties were retrieved, but it is hard for me to tell if the acl is put into dpp_entries[16].

src/common/mfu_daos.c
daltonbohning: @dsikich Yes, it gets appended. I had to look at the source code for this one:
https://github.com/daos-stack/daos/blob/adb2578ac6659ff510309de63aaa10b660f65301/src/common/prop.c#L104
Basically, this is what the merge does:
```
prop_a = [A=0, B=1, C=2]
prop_b = [B=4, D=5]
merge(prop_a, prop_b) = [A=0, B=4, C=2, D=5]
```
I guess it's _**possible**_ that this implementation could change, but probably unlikely for our use case, since both property lists are distinct (no overlap).

src/common/mfu_daos.c
daltonbohning: @dsikich I'll leave a comment explaining this

dsikich (approved):

src/common/mfu_daos.c
dsikich: @daltonbohning ok, that makes sense you had to look at the source code. I looked at the header and it didn't say.

src/common/mfu_daos.c
```diff
@@ -610,6 +610,291 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+    props->dpp_entries[6].dpe_type = DAOS_PROP_CO_REDUN_FAC;
+    props->dpp_entries[7].dpe_type = DAOS_PROP_CO_REDUN_LVL;
+    props->dpp_entries[8].dpe_type = DAOS_PROP_CO_SNAPSHOT_MAX;
+    props->dpp_entries[9].dpe_type = DAOS_PROP_CO_COMPRESS;
+    props->dpp_entries[10].dpe_type = DAOS_PROP_CO_ENCRYPT;
+    props->dpp_entries[11].dpe_type = DAOS_PROP_CO_OWNER;
+    props->dpp_entries[12].dpe_type = DAOS_PROP_CO_OWNER_GROUP;
+    props->dpp_entries[13].dpe_type = DAOS_PROP_CO_DEDUP;
+    props->dpp_entries[14].dpe_type = DAOS_PROP_CO_DEDUP_THRESHOLD;
+    props->dpp_entries[15].dpe_type = DAOS_PROP_CO_ALLOCED_OID;
```
dsikich, line 652: @daltonbohning this property should not be set for POSIX containers.

daltonbohning: @dsikich Thanks! Specifically, it should not be set when we use DFS. I think we should still set it for POSIX -> POSIX with the Object API, just not DFS. It seemed the simplest way was to just pass a `get_oid` flag

dsikich: > @dsikich Thanks! Specifically, it should not be set when we use DFS. I think we should still set it for POSIX -> POSIX with the Object API, just not DFS. It seemed the simplest way was to just pass a `get_oid` flag

@daltonbohning yeah, I agree 👍

dsikich (approved):
@daltonbohning looks good. There will eventually be a DFS version of serialization/deserialization, so I can just update the flag when the container is serialized for that case whenever it gets added.

dsikich:
@adammoody can you take a look at this one? I am working on an option/feature that builds on this PR.

src/common/mfu_daos.c
```diff
@@ -610,6 +610,300 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+        goto out;
+    }
+
+    /* Get the attribute names */
+    rc = daos_cont_list_attr(coh, name_buf, &total_size, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "failed to list user attributes "DF_RC, DP_RC(rc));
+        rc = 1;
+        goto out;
+    }
```
adammoody, line 773: Just curious about the ``daos_cost_list_attr`` interface.  Is ``total_size`` used as an input parameter here to specify the buffer size?

If so, I'm guessing this function might return an error if the attribute list size has increased, since the first time this was called.

And I suspect it's valid for the attribute list size to be less than what it was?

adammoody (approved):

adammoody:
Thanks @daltonbohning and @dsikich .  This looks good to me.

src/common/mfu_daos.c
dsikich: @adammoody I believe it is valid for it to be less than it was, but I will note that we should return an error if the attribute size has increased since the first time it was called.

daltonbohning: @adammoody @dsikich Yes, it is possible for the actual size to change between the first call to `daos_cont_list_attr` (which just gets the size) and the second call (which gets the names).
https://github.com/daos-stack/daos/blob/011b78cb28aaf162af635f37d1fcabc62076ef0b/src/include/daos_cont.h#L382-L399
I **_think_** this is how it works, based on other DAOS paradigms:
- If size is less than the size passed in
  - The output size set accordingly, since it's passed by reference, and the buffer contains all the data
- If size is more than the size passed in
  - The output size is set accordingly, and the buffer is **_truncated_**

The "foolproof" way to handle this would be to have a retry loop, and realloc and fetch every time the size is **_more_** than the size of the buffer.

daltonbohning:
> Thanks @daltonbohning and @dsikich . This looks good to me.

Thanks @adammoody !

PR #482, opened 2021-04-20
serialization optimizations for DAOS_OF_KV_FLAT

dsikich:
* add daos-gen tool to generate a DAOS_OF_KV_FLAT objects
    * skip akey iteration on serialize and deserialize for DAOS_OF_KV_FLAT

If an object has the DAOS_OF_KV_FLAT feature set, then there is no
need to do any sort of akey iteration. The daos_kv API is used
because all that is needed is oid, dkey, and the value. The
obj API and the daos_kv API cannot be mixed, or it will cause
errors. The KV value is stored inside of the dkey dataset.


dsikich:
@daltonbohning @adammoody should not be merged before #477

src/common/mfu_daos.c
```diff
@@ -1973,6 +1973,51 @@ static inline void init_hdf5_args(struct hdf5_args *hdf5)
...
+    rc = daos_kv_get(*oh, DAOS_TX_NONE, 0, dkey_val, &size, buf, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to fetch object");
+        goto out;
+    }
+    buf = MFU_CALLOC(1, size);
+    if (buf == NULL) {
+        rc = ENOMEM;
+        goto out;
+    }
```
adammoody, line 1999: It's fine to leave this check for NULL, but it probably would not trigger if actually out of memory.  The MFU allocate functions will typically abort if they fail to acquire memory, e.g.,

https://github.com/hpc/mpifileutils/blob/55255604d87d581eb52a710c2accad31581136e7/src/common/mfu_util.c#L105

src/common/mfu_daos.c
```diff
@@ -4075,122 +4199,138 @@ static int cont_deserialize_keys(struct hdf5_args *hdf5,
...
+                memset(akey, 0, sizeof(akey));
+                memcpy(akey, hdf5->akey_data[ak_off + k].akey_val.p,
+                       hdf5->akey_data[ak_off + k].akey_val.len);
+                d_iov_set(&aiov, (void*)hdf5->akey_data[ak_off + k].akey_val.p,
+                          hdf5->akey_data[ak_off + k].akey_val.len);
+
+                /* read record data for each akey */
+                index = ak_off + k;
+                len = snprintf(NULL, 0, "%lu", index);
+                char dset_name[len + 1];
```
adammoody, line 4230: I think this needs to be malloc'd right?

src/common/mfu_daos.c
```diff
@@ -4075,122 +4199,138 @@ static int cont_deserialize_keys(struct hdf5_args *hdf5,
...
-                    MFU_LOG(MFU_LOG_ERR, "failed to deserialize recx");
+            stats->bytes_written += kv_single_size;
+        } else {
+            for(k = 0; k < total_akeys_this_dkey; k++) {
+                memset(&aiov, 0, sizeof(aiov));
+                memset(akey, 0, sizeof(akey));
+                memcpy(akey, hdf5->akey_data[ak_off + k].akey_val.p,
+                       hdf5->akey_data[ak_off + k].akey_val.len);
+                d_iov_set(&aiov, (void*)hdf5->akey_data[ak_off + k].akey_val.p,
+                          hdf5->akey_data[ak_off + k].akey_val.len);
```
adammoody, line 4225: For readability and to minimize the chance for typos, I always find it to be help to define a local variable to simplify long names, especially when the same long string is used multiple times.  Having a local variable also helps when viewing values under a debugger.

I don't know the exact type needed here, but one could simplify with something like:
```suggestion
                type* akeyval = &hdf5->akey_data[ak_off + k].akey_val;
                memcpy(akey, akeyval->p, akeyval->len);
                d_iov_set(&aiov, (void*)akeyval->p, akeyval->len);
```
It's much easier to see that all four references map to the same struct this way, and under TotalView, we can just double-click ``akeyval`` to immediate see its ``p`` and ``len`` values rather than diving through a more complicated structure.

src/daos-gen/daos-gen.c
```diff
@@ -0,0 +1,290 @@
...
+        MPI_Finalize();
+        return 1;
+    }
+
+    /* set default number of objects and keys per object,
+     * if one is not passed in */
+    if (num_objects == 0)
+        num_objects = 10;
+    if (keys_per_object == 0)
+        keys_per_object = 20;
```
adammoody, line 166: Just a style thing to always include braces.
```suggestion
    if (num_objects == 0) {
        num_objects = 10;
    }
    if (keys_per_object == 0) {
        keys_per_object = 20;
    }
```

src/daos-gen/daos-gen.c
```diff
@@ -0,0 +1,290 @@
...
+
+    daos_obj_id_t       oid[num_objects];
+	daos_handle_t	    oh[num_objects];
+	char		        buf[BUFLEN];
+    const char          *key_fmt = "key%d";
```
adammoody, line 171: Have some tabs and spaces mixed in this part.
```suggestion
    daos_obj_id_t       oid[num_objects];
    daos_handle_t        oh[num_objects];
    char    	        buf[BUFLEN];
    const char          *key_fmt = "key%d";
```

adammoody:
This generally looks good to me.  Of course, I can't do any runtime testing, so that limits how deeply I can review things.  Just a few high-level things I saw:
- Some mixing of tabs and spaces
- Check that all if/while blocks have braces, even for single lines blocks
- There are several spots where I see ``char foo[size]`` where ``size`` is computed at runtime.  I think those probably need to use a ``malloc(size)/free()`` instead in some cases.

src/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ ADD_SUBDIRECTORY(drm)
 ADD_SUBDIRECTORY(dstripe)
 ADD_SUBDIRECTORY(dsync)
 IF(ENABLE_DAOS AND ENABLE_HDF5)
+  ADD_SUBDIRECTORY(daos-gen)
```
daltonbohning, line 19: @dsikich Should we put `daos-gen` under experimental since it's really just being used for testing purposes?

src/daos-gen/daos-gen.c
```diff
@@ -0,0 +1,290 @@
...
+
+    /* TODO: generate different types of data, and different for each key */
+	memset(buf, 'A', BUFLEN);
+    for (i = 0; i < num_objects; i++) {
+        oid[i] = dts_oid_gen(0);
+                                   
+        rc = daos_obj_generate_oid(daos_args->src_coh, &oid[i], ofeats, OC_RP_XSF, 0, 0);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "Failed to generate oid: "DF_RC, DP_RC(rc));
+            return 1;
```
daltonbohning, line 222: @dsikich return from main here doesn't properly finalize MPI and MFU

src/daos-gen/daos-gen.c
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+    	/** create the KV store */
+        rc = daos_kv_open(daos_args->src_coh, oid[i], DAOS_OO_RW, &oh[i], NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "Failed to open kv object: "DF_RC, DP_RC(rc));
+            return 1;
```
daltonbohning, line 229: @dsikich return from main here doesn't properly finalize MPI and MFU

src/daos-gen/daos-gen.c
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+        /* insert keys */
+        for (j = 0; j < keys_per_object; j++) {
+            sprintf(key, key_fmt, j);
+            rc = daos_kv_put(oh[i], DAOS_TX_NONE, 0, key, BUFLEN, buf, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to put kv object: "DF_RC, DP_RC(rc));
+                return 1;
```
daltonbohning, line 238: @dsikich return from main here doesn't properly finalize MPI and MFU

src/daos-gen/daos-gen.c
```diff
@@ -0,0 +1,290 @@
...
+            }
+        }
+
+        rc = daos_kv_close(oh[i], NULL);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "Failed to close kv object: "DF_RC, DP_RC(rc));
+            return 1;
```
daltonbohning, line 245: @dsikich return from main here doesn't properly finalize MPI and MFU

src/daos-gen/daos-gen.c
```diff
@@ -0,0 +1,290 @@
...
+
+    /* calculate total objects and num_keys_per object, each rank
+     * generates same amount */
+    if (rank == 0) {
+        int total_num_objects = size * num_objects; 
+        uuid_unparse(daos_args->src_cont_uuid, uuid_str);
+        printf("Container UUID: %s\n\ttotal objects:%d\n"
+               "\tkeys per object:%d\n", uuid_str, total_num_objects, keys_per_object);
+    }
+
```
daltonbohning, line 257: @dsikich I'd recommend putting a `label` here so error conditions above can `goto` for proper cleanup.

daltonbohning (commented):
I left some comments just for daos-gen. Since this is only being used internally for testing purposes, I don't think it's as important to be "perfect", especially if we move to to experimental tools.

src/common/mfu_daos.c
```diff
@@ -1973,6 +1973,51 @@ static inline void init_hdf5_args(struct hdf5_args *hdf5)
...
+
+    /* get the size of the value */
+    rc = daos_kv_get(*oh, DAOS_TX_NONE, 0, dkey_val, &size, buf, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to fetch object");
+        goto out;
+    }
+    buf = MFU_CALLOC(1, size);
+    if (buf == NULL) {
+        rc = ENOMEM;
```
daltonbohning, line 1997: @dsikich Since the daos_ functions return -DER codes, we should probably do the same:
```suggestion
        rc = -DER_NOMEM;
```

src/common/mfu_daos.c
```diff
@@ -1973,6 +1973,51 @@ static inline void init_hdf5_args(struct hdf5_args *hdf5)
...
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to fetch object");
+        goto out;
+    }
+
+    stats->bytes_read += size;
+    kv_val = &(*hdf5->dk)[*dk_index].rec_kv_val;
+    kv_val->p = MFU_CALLOC(1, (uint64_t)size);
+    if (kv_val->p == NULL) {
+        rc = ENOMEM;
```
daltonbohning, line 2011: @dsikich Similarly:
```suggestion
        rc = -DER_NOMEM;
```

src/common/mfu_daos.c
```diff
@@ -2575,14 +2623,22 @@ static int serialize_dkeys(struct hdf5_args *hdf5,
...
-                                dkey_kds, &dkey_sgl, &dkey_anchor,
-                                NULL);
-        if (rc != 0) {
-            MFU_LOG(MFU_LOG_ERR, "failed to list dkeys: %d", rc);
-            goto out;
+        if (ofeat & DAOS_OF_KV_FLAT) {
+            rc = daos_kv_list(*oh, DAOS_TX_NONE, &dkey_number,
+                              dkey_kds, &dkey_sgl, &dkey_anchor, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "failed to list dkeys: %d", rc);
```
daltonbohning, line 2630: @dsikich Formatted daos error
```suggestion
                MFU_LOG(MFU_LOG_ERR, "failed to list dkeys: "DF_RC, DP_RC(rc));
```

src/common/mfu_daos.c
```diff
@@ -2575,14 +2623,22 @@ static int serialize_dkeys(struct hdf5_args *hdf5,
...
+                goto out;
+            }
+        } else {
+            rc = daos_obj_list_dkey(*oh, DAOS_TX_NONE, &dkey_number,
+                                    dkey_kds, &dkey_sgl, &dkey_anchor, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "failed to list dkeys: %d", rc);
```
daltonbohning, line 2637: @dsikich formatted daos error
```suggestion
                MFU_LOG(MFU_LOG_ERR, "failed to list dkeys: "DF_RC, DP_RC(rc));
```

src/common/mfu_daos.c
```diff
@@ -3580,24 +3676,50 @@ int daos_cont_serialize_hdlr(int rank, struct hdf5_args *hdf5, char *output_dir,
...
+         * a better way to organize this with swtich statements, or
+         * creating "daos_obj" wrappers, etc. */
+        ofeat = (oid.hi & OID_FMT_FEAT_MASK) >> OID_FMT_FEAT_SHIFT;
+        if (ofeat & DAOS_OF_KV_FLAT) {
+            rc = daos_kv_open(da->src_coh, oid, DAOS_OO_RW, &oh, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "failed to open kv object: "DF_RC, DP_RC(rc));
+                goto out;
+            }
+            /* TODO: for DAOS_OF_KV_FLAT use daos_kv.h and daos_kv_list to list dkeys */
```
daltonbohning, line 3691: @dsikich Is this TODO resolved?

src/common/mfu_daos.c
```diff
@@ -3580,24 +3676,50 @@ int daos_cont_serialize_hdlr(int rank, struct hdf5_args *hdf5, char *output_dir,
...
-        rc = daos_obj_close(oh, NULL);
-        if (rc != 0) {
-            MFU_LOG(MFU_LOG_ERR, "failed to close object: "DF_RC, DP_RC(rc));
-            goto out;
-        }
 
+        /* TODO: DAOS_OF_KV_FLAT uses daos_kv_* functions, and
+         * other object types use daos_obj_* functions. Maybe there is
+         * a better way to organize this with swtich statements, or
+         * creating "daos_obj" wrappers, etc. */
```
daltonbohning, line 3683: @dsikich perhaps we could have open and close wrappers that check the oid features and call the corresponding kv_ or obj_ functions

src/common/mfu_daos.c
```diff
@@ -4075,122 +4199,138 @@ static int cont_deserialize_keys(struct hdf5_args *hdf5,
...
+                MFU_LOG(MFU_LOG_ERR, "failed to put kv object: "DF_RC, DP_RC(rc));
                 goto out;
             }
-            if (num_attrs > 0) {
-                rc = cont_deserialize_recx(hdf5, oh, diov,
-                                           num_attrs, ak_off, k, stats);
-                if (rc != 0) {
-                    MFU_LOG(MFU_LOG_ERR, "failed to deserialize recx");
+            stats->bytes_written += kv_single_size;
+        } else {
```
daltonbohning, line 4218: @dsikich Instead of the large if/else blocks, I wonder if we could just abstract each to a function.

src/common/mfu_daos.c
dsikich: @daltonbohning yeah, I was thinking we could do that, but I am not sure if it is worth making wrappers since I don't think we will be adding more API's?

daltonbohning: @dsikich Yeah, maybe not. I think right now open/close is only used in like 2 different places. So there're only 2 copies of each, and it's very simple

daltonbohning: @dsikich But it could help simplify the if/else if we do this for dcp/dsync as well

src/common/mfu_daos.c
dsikich: @adammoody yeah, indexing into those datasets was getting very long

src/CMakeLists.txt
dsikich: @daltonbohning yeah, I suppose my issue while writing the code I have to build all of the experimental tools

src/common/mfu_daos.c
dsikich: @daltonbohning yeah, this function is becoming rather large. I can move them to their own functions.

dsikich:
@daltonbohning @adammoody addressed the comments and broke up one of the deserialization functions

src/common/mfu_daos.c
```diff
@@ -4012,6 +4141,141 @@ static int cont_deserialize_recx(struct hdf5_args *hdf5,
...
+    } else {
+        memset(&sgl, 0, sizeof(sgl));
+        memset(&iov, 0, sizeof(iov));
+        memset(&iod, 0, sizeof(iod));
+        single_tsize = rec_single_val->len;
+        if (single_tsize == 0) {
+            rc = 1;
+            MFU_LOG(MFU_LOG_ERR, "failed to get size of type in single "
+                    "record datatype");
+        }
```
daltonbohning, line 4242: @dsikich should we `goto out` here?

src/common/mfu_daos.c
```diff
@@ -4724,14 +4895,27 @@ int daos_cont_deserialize_hdlr(int rank, daos_args_t *da, const char *h5filename
...
+        if ((ofeat & DAOS_OF_KV_FLAT) &&
+            !(ofeat & DAOS_OF_ARRAY_BYTE) && !(ofeat & DAOS_OF_ARRAY)) {
+            is_kv_flat = true;
+        }
+        if (is_kv_flat) {
+            rc = daos_kv_open(da->src_coh, oid, DAOS_OO_RW, &oh, NULL);
+            if (rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "Failed to open kv object: "DF_RC, DP_RC(rc));
+                rc = 1;
+            }
```
daltonbohning, line 4912: @dsikich Should we `goto out` here?

src/daos-gen/daos-gen.c
```diff
@@ -0,0 +1,302 @@
...
+static uint32_t obj_id_gen = 1;
+
+/** Print a usage message. */
+void print_usage(void)
+{
+    printf("\n");
+    printf("Usage: daos-gen [options]\n");
+    printf("\n");
+    printf("Options:\n");
+    printf("  -p, --pool               - pool uuid for containers\n");
```
daltonbohning, line 30: @dsikich If we want to include the other options.
```suggestion
    printf("  -p, --pool               - pool uuid for containers\n");
    printf("  -o, --num-objects  - number of objects to generate\n");
    printf("  -k, --keys-per-object - number of keys per object\n");
```

daltonbohning (commented):
@dsikich I ran our existing internal serialization tests to verify this didn't break anything. And then I used `daos-gen` + `dsync` to verify the optimization works.
Functionally, this seems to be working.

dsikich:
@daltonbohning I've updated based on your review, thanks.

daltonbohning (approved):

dsikich:
@adammoody is this good to merge now?

adammoody (approved):

adammoody:
Yep, looks good to me, too.  Thanks, @dsikich

PR #485, opened 2021-05-03
travis: use default xenial instead of trusty

daltonbohning:
The "trusty' distro seems to be missing a public key.

The Travis documentation recommends not using Trusty, as it is EOL:
- https://docs.travis-ci.com/user/reference/trusty/

For reference, this is the difference:
- https://docs.travis-ci.com/user/reference/xenial/#differences-from-the-trusty-images

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: xenial
 
 language: c
 


daltonbohning:
@adammoody Very simple change. This fixes the build errors that have been occurring, since trusty is EOL.

adammoody (approved):
Yes, thanks @daltonbohning

PR #486, opened 2021-05-05
add daos-preserve option for preserving daos metadata

dsikich:
This option depends on the hdf5 library, and can only be used
when moving data to or from a POSIX filesystem. If a user
moves data from a POSIX DAOS Container to NFS or Lustre they
can write the container properties and user attributes to
an hdf5 file. When a user is ready to move their data back
into a container they can pass the metadata file into the
copy and the properties will be read from the hdf5 file.
The container properties and user attributes are set on
container creation in the case the the data is moving from
a POSIX filesystem to a DAOS POSIX container.

If a user is moving data from one DAOS container to another
the metadata is already preserved from the source to destination,
and passing the option is not necessary.

This option can be used by setting --daos-preserve <path/to/use> where the filename will be appended to the end of the path. 


src/common/mfu_daos.h
```diff
@@ -47,6 +47,8 @@ typedef struct {
     bool allow_exist_dst_cont;          /* whether to allow the dst container to exist for DAOS API */
     enum daos_cont_props src_cont_type; /* type of the source container */
     enum daos_cont_props dst_cont_type; /* type of the destination container */
+    bool daos_preserve;                 /* preserve daos cont props and user attrs */
+    char *daos_preserve_path;           /* set path to write daos props and user attrs */
```
daltonbohning, line 51: @dsikich I wonder if we could simplify this by removing `daos_preserve` and using `daos_preserve_path != NULL` to mean `true`

src/common/mfu_daos.c
```diff
@@ -992,10 +993,32 @@ int daos_connect(
     int             valid = 0; /* assume failure until otherwise */
     int             rc;
     daos_prop_t*    props = NULL;
+    struct daos_prop_entry *entry;
+    char *filename;
+#ifdef HDF5_SUPPORT
+    struct hdf5_args hdf5;
+#endif
```
daltonbohning, line 1000: @dsikich Should `entry` and `filename` also be within `HDF5_SUPPORT`?

src/common/mfu_daos.c
```diff
@@ -992,10 +993,32 @@ int daos_connect(
...
 
     /* have rank 0 connect to the pool and container,
      * we'll then broadcast the handle ids from rank 0 to everyone else */
     if (rank == 0) {
+#ifdef HDF5_SUPPORT
+        /* initialization for hdf5 file for preserve option */
+        if (preserve) {
+            rc = asprintf(&filename, "%s%s%s", da->daos_preserve_path,
+                          "/", "daos_metadata.hdf5");
```
daltonbohning, line 1009: @dsikich Maybe we should allow the user to pass in the path to the file itself. I can see how users might have several of these metadata files for different scenarios, and they might want to pick a specific one.

src/common/mfu_daos.c
```diff
@@ -1041,6 +1063,20 @@ int daos_connect(
              * create a POSIX container. Otherwise, create a container of
              * the same type as the source. */
             if (!have_src_cont || (da->src_cont_type == DAOS_PROP_CO_LAYOUT_POSIX)) {
+#ifdef HDF5_SUPPORT
+                /* read container properties in hdf5 file when moving data
+                 * back to DAOS if the preserve option has been set */
+                if (preserve) {
+                    /* TODO: should always be called daos_metadata.hdf5? */
```
daltonbohning, line 1070: @dsikich I think the simplest approach is for the user to provide the filename. But if we want to get "smart" about it, we could:
1. If daos_preserve_path is a directory, look for daos_metadata.hdf5
2. If daos_perserve_path is a file, use the file.

We could do this both for serializing and deserializing the metadata; though, I'm not sure if we want that complexity

src/common/mfu_daos.c
```diff
@@ -1041,6 +1063,20 @@ int daos_connect(
...
+                    daos_cont_layout_t ctype = DAOS_PROP_CO_LAYOUT_POSIX;
+                    printf("Reading metadata file: %s\n", filename);
+                    rc = cont_deserialize_all_props(&hdf5, props, entry, &ctype);
```
daltonbohning, line 1073: @dsikich What is the purpose of passing `entry` here? It doesn't seem to be used outside of `cont_deserialize_all_props`

src/common/mfu_daos.c
```diff
@@ -1244,6 +1309,90 @@ int daos_parse_epc_str(
...
+        goto out;
+    }
+
+    /* TODO: much of this HDF5 setup should get moved into the HDF5 user
+     * interface */
+    hdf5.file = H5Fcreate(filename, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
+    if (hdf5.file < 0) {
+        MFU_LOG(MFU_LOG_ERR, "failed to create hdf5 file");
+        goto out;
+    }
```
daltonbohning, line 1333: @dsikich I think we should set `rc = 1` here so the caller properly knows it failed.

src/common/mfu_daos.c
```diff
@@ -1244,6 +1309,90 @@ int daos_parse_epc_str(
...
+    if (status < 0) {
+        MFU_LOG(MFU_LOG_ERR, "failed to insert user attr name");
+        goto out;
+    }
+    status = H5Tinsert(hdf5.usr_attr_memtype, "Attribute Value",
+                       HOFFSET(usr_attr_t, attr_val), hdf5.usr_attr_val_vtype);
+    if (status < 0) {
+        MFU_LOG(MFU_LOG_ERR, "failed to insert user attr val");
+        goto out;
+    }
```
daltonbohning, line 1372: @dsikich I think we should set rc = 1 for these failed hdf5 calls so the caller properly knows it failed.

src/common/mfu_daos.c
```diff
@@ -1397,19 +1547,32 @@ int daos_setup(
...
 
+        /* TODO: pass daos_preserve bool here to daos_connect=true */
         create_cont = true;
+        /* do check that src is POSIX and since dst has a pool,
+         * then the dst *should* always be DFS, but this is not set
+         * on the destintaion until after daos_connect is called, and
+         * we should read the properties *before* the container is
+         * created. We do not have the destination container type
+         * here yet which is why extra check is used then passed
+         * to daos_connect */
```
daltonbohning, line 1558: @dsikich Hopefully we'll have a test case for this in the future to fully verify all cases work

src/common/mfu_daos.c
```diff
@@ -1397,19 +1547,32 @@ int daos_setup(
...
+        if (da->daos_preserve) {
+            if (mfu_src_file->type == POSIX) {
+                preserve = true; 
+            }
+        }
```
daltonbohning, line 1563: @dsikich I wonder if we DO want to allow `--daos-preserve` for DAOS->DAOS?
Something like:
```
if preserve:
    use preserve file for dst cont create
else:
    use src cont for dst cont create
```

Otherwise, I think we should probably error here, so the user knows that the `--daos-preserve` they passed in isn't going to be used

src/common/mfu_daos.c
```diff
@@ -1464,8 +1627,22 @@ int daos_setup(
         }
     }
 
+    /* check daos_preserve
+     * if src cont is DFS an dst is POSIX, then write
```
daltonbohning, line 1631:
```suggestion
     * if src cont is DFS and dst is POSIX, then write 
```

src/common/mfu_daos.h
dsikich: @daltonbohning yeah that would remove the extra variable.

src/common/mfu_daos.c
dsikich: @daltonbohning I went back and forth on this, and wasn't sure if I should just create the name of the file or expect the user to pass in a name. I think what you suggested below is good where we only create the name if the user passes in a directory.

src/common/mfu_daos.c
dsikich: @daltonbohning ok, yeah it is probably simplest to just expect the user to provide a filename

src/common/mfu_daos.c
dsikich: @daltonbohning hmm.. I will have to take a look but I just remember the function required passing it in

src/common/mfu_daos.h
dsikich: @daltonbohning this doesn't totally work actually, because I need to pass different values to daos connect depending on if DAOS is the source or destination.

daltonbohning: @dsikich Ah, that makes sense!

src/common/mfu_daos.c
daltonbohning: @dsikich It might be worth asking about this in the datamover WG to get other opinions

src/common/mfu_daos.c
dsikich: @daltonbohning I removed it since it didn't need to be passed around

dsikich:
@daltonbohning I've updated based on your comments, and my changes are in the "updates" commit so you can see what I've updated.

src/dcp/dcp.c
```diff
@@ -82,6 +82,7 @@ void print_usage(void)
 #ifdef DAOS_SUPPORT
     printf("      --daos-prefix        - DAOS prefix for unified namespace path\n");
     printf("      --daos-api           - DAOS API in {DFS, DAOS} (default uses DFS for POSIX containers)\n");
+    printf("      --daos-preserve      - preserve DAOS container properties and user attributes\n");
```
daltonbohning, line 85: @dsikich
1. Should we also wrap this in #ifdef HDF5_SUPPORT?
2. Or maybe at least mention HDF5 in the help text?
3. We should mention that a path is expected so the user doesn't think this is just a flag.
4. We should probably also update the doc: https://github.com/hpc/mpifileutils/blob/master/doc/rst/dcp.1.rst

src/common/mfu_daos.c
```diff
@@ -1041,6 +1055,20 @@ int daos_connect(
              * create a POSIX container. Otherwise, create a container of
              * the same type as the source. */
             if (!have_src_cont || (da->src_cont_type == DAOS_PROP_CO_LAYOUT_POSIX)) {
+#ifdef HDF5_SUPPORT
+                /* read container properties in hdf5 file when moving data
+                 * back to DAOS if the preserve option has been set */
+                if (preserve) {
+                    /* TODO: should always be called daos_metadata.hdf5? */
+                    daos_cont_layout_t ctype = DAOS_PROP_CO_LAYOUT_POSIX;
+                    printf("Reading metadata file: %s\n", da->daos_preserve_path);
```
daltonbohning, line 1064: @dsikich We should probably use MFU_LOG instead of printf

src/common/mfu_daos.c
```diff
@@ -1244,6 +1300,93 @@ int daos_parse_epc_str(
...
 
+#ifdef HDF5_SUPPORT
+static int serialize_daos_metadata(char *filename,
+                                   daos_args_t* da)
+{
+    int    rc = 0;
+    hid_t  status = 0;
+    struct hdf5_args hdf5;
+
+    printf("Writing metadata file: %s\n", da->daos_preserve_path);
```
daltonbohning, line 1311: @dsikich We should probably use MFU_LOG instead of printf

src/common/mfu_daos.c
```diff
@@ -1464,8 +1627,22 @@ int daos_setup(
...
+     * cont props and user attrs */
+#ifdef HDF5_SUPPORT
+    char *filename = NULL;
+    if (da->daos_preserve && rank == 0) {
+        if (mfu_src_file->type == DFS && mfu_dst_file->type == POSIX) {
+            tmp_rc = serialize_daos_metadata(filename, da);
+            if (tmp_rc != 0) {
+                MFU_LOG(MFU_LOG_ERR, "failed serialize DAOS metadata: "DF_RC,
+                        DP_RC(tmp_rc));
+            }
```
daltonbohning, line 1641: @dsikich Should this set rc and/or local_daos_error?

daltonbohning (approved):
@dsikich Functionally, this seems to be working (for some simple manual tests, anyway).
I just had some comments about errors/docs/etc.

dsikich:
> @dsikich Functionally, this seems to be working (for some simple manual tests, anyway).
> I just had some comments about errors/docs/etc.

@daltonbohning thanks! I updated the docs and addressed your comments in my latest commit.

daltonbohning (approved):

adammoody:
Hi all, I'll take a look at this by Monday.

src/common/mfu_daos.c
```diff
@@ -1041,6 +1055,20 @@ int daos_connect(
              * create a POSIX container. Otherwise, create a container of
              * the same type as the source. */
             if (!have_src_cont || (da->src_cont_type == DAOS_PROP_CO_LAYOUT_POSIX)) {
+#ifdef HDF5_SUPPORT
+                /* read container properties in hdf5 file when moving data
+                 * back to DAOS if the preserve option has been set */
+                if (preserve) {
+                    /* TODO: should always be called daos_metadata.hdf5? */
+                    daos_cont_layout_t ctype = DAOS_PROP_CO_LAYOUT_POSIX;
+                    MFU_LOG(MFU_LOG_INFO, "Reading metadata file: %s\n", da->daos_preserve_path);
```
adammoody, line 1064: MFU_LOG should add the newline for us.
```suggestion
                    MFU_LOG(MFU_LOG_INFO, "Reading metadata file: %s", da->daos_preserve_path);
```

src/common/mfu_daos.c
```diff
@@ -1087,6 +1132,12 @@ int daos_connect(
 bcast:
     if (rank == 0) {
         daos_prop_free(props);
+#ifdef HDF5_SUPPORT
+        /* only close if handle is open */
+        if (hdf5.file > 0) {
+            H5Fclose(hdf5.file);
+        }
+#endif
```
adammoody, line 1140: Should we guard this with ``if (preserve)`` like the other sections?

src/common/mfu_daos.c
```diff
@@ -992,10 +993,24 @@ int daos_connect(
     int             valid = 0; /* assume failure until otherwise */
     int             rc;
     daos_prop_t*    props = NULL;
+#ifdef HDF5_SUPPORT
+    struct hdf5_args hdf5;
```
adammoody, line 997: Is there any kind of "NULL" value we can initialize this to?

src/common/mfu_daos.c
```diff
@@ -1244,6 +1300,93 @@ int daos_parse_epc_str(
...
 
+#ifdef HDF5_SUPPORT
+static int serialize_daos_metadata(char *filename,
+                                   daos_args_t* da)
+{
+    int    rc = 0;
+    hid_t  status = 0;
+    struct hdf5_args hdf5;
+
+    MFU_LOG(MFU_LOG_INFO, "Writing metadata file: %s\n", da->daos_preserve_path);
```
adammoody, line 1311:
```suggestion
    MFU_LOG(MFU_LOG_INFO, "Writing metadata file: %s", da->daos_preserve_path);
```

src/common/mfu_daos.c
```diff
@@ -1068,6 +1096,23 @@ int daos_connect(
...
+             * copy */
+            if (preserve) { 
+                /* deserialize and set the user attributes if they exist */
+                htri_t usr_attrs_exist = H5Lexists(hdf5.file, "User Attributes", H5P_DEFAULT);
+                if (usr_attrs_exist > 0) {
+                    rc = cont_deserialize_usr_attrs(&hdf5, *coh);
+                    if (rc != 0) {
+                        rc = 1;
+                        goto bcast;
+                    }
```
adammoody, line 1111: For ``cont_deserialize_usr_attrs`` we set ``rc = 1``, but we do not set ``rc = 1`` for ``cont_deserialize_all_props`` above.  In both cases, I see they use ``rc`` to hold the return value from the function call, so I'm guessing either way is valid.  Can we make them consistent?

adammoody (approved):
Thanks @dsikich .  I found a few minor things to check, otherwise this looks good to me.

dsikich:
> Thanks @dsikich . I found a few minor things to check, otherwise this looks good to me.



> Thanks @dsikich . I found a few minor things to check, otherwise this looks good to me.

thanks @adammoody I addressed your comments.

PR #487, opened 2021-05-17
daos: remove prefix from documentation and update call to duns_resolve_path

dsikich:
This just removes the daos-prefix option from the documentation
because it is no longer necessary after a recent DAOS PR
that allows passing a subset of a UNS path to duns_resolve_path.
The option is removed from the documentation but not removed
from the code in order to preserve backwards compatibility with
the DAOS tests for the datamover.

The call to duns_resolve_path was updated in the case that the
last directory or file in the UNS path does not yet exist, since
it doesn't recurse the rest of the paths in the directory tree
in that case.

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -14,6 +14,7 @@
 #include <daos_uns.h>
 #include <gurt/common.h>
 #include <gurt/hash.h>
+#include <libgen.h>
 
 #ifdef HDF5_SUPPORT
 #include <hdf5.h>
@@ -323,8 +324,13 @@ int daos_parse_path(
 {
     struct duns_attr_t  dattr = {0};
     int                 rc;
+    char*               tmp_path1 = NULL;
+    char*               path_dirname = NULL;
+    char*               tmp_path2 = NULL;
+    char*               path_basename = NULL;
+    char*               tmp = NULL;
 
-    /* Check if this path represents a daos pool and/or container. */
+    /* check first if duns_resolve_path succeeds on regular path */
     rc = duns_resolve_path(path, &dattr);
     if (rc == 0) {
         /* daos:// or UNS path */
@@ -335,17 +341,66 @@ int daos_parse_path(
         } else {
             strncpy(path, dattr.da_rel_path, path_len);
         }
-    } else if (strncmp(path, "daos:", 5) == 0) {
-        /* Actual error, since we expect a daos path */
-        rc = -1;
     } else {
-        /* We didn't parse a daos path,
-         * but we weren't necessarily looking for one */
-        rc = 1;
+        /* If basename does not exist yet then duns_resolve_path will fail even
+        * if dirname is a UNS path */
+
+        /* get dirname */
+        tmp_path1 = strdup(path);
+        if (tmp_path1 == NULL) {
+            rc = -ENOMEM;
+            goto out;
+        }
+        path_dirname = dirname(tmp_path1);
+
+        /* reset before calling duns_resolve_path with new string */
+        memset(&dattr, 0, sizeof(struct duns_attr_t));
+
+        /* Check if this path represents a daos pool and/or container. */
+        rc = duns_resolve_path(path_dirname, &dattr);
+        /* if it succeeds get the basename and append it to the rel_path */
+        if (rc == 0) {
+            /* if duns_resolve_path succeeds then concat basename to 
+            * da_rel_path */
+            tmp_path2 = strdup(path);
+            if (tmp_path2 == NULL) {
+                rc = -ENOMEM;
+                goto out;
+            }
+            path_basename = basename(tmp_path2);
+   
+            /* dirname might be root uns path, if that is the case,
+             * then da_rel_path might be NULL */
+            if (dattr.da_rel_path == NULL) {
+                tmp = MFU_CALLOC(path_len, sizeof(char));
+            } else {
+                tmp = realloc(dattr.da_rel_path, path_len);
+            }
+            if (tmp == NULL) {
+                rc = -ENOMEM;
+                goto out;
+            }
+            dattr.da_rel_path = tmp;
+            strcat(dattr.da_rel_path, "/");
+            strcat(dattr.da_rel_path, path_basename);
+
+            snprintf(*pool_str, DAOS_PROP_LABEL_MAX_LEN + 1, "%s", dattr.da_pool);
+            snprintf(*cont_str, DAOS_PROP_LABEL_MAX_LEN + 1, "%s", dattr.da_cont);
+            strncpy(path, dattr.da_rel_path, path_len);
+        } else if (strncmp(path, "daos:", 5) == 0) {
+            /* Actual error, since we expect a daos path */
+            rc = -1;
+        } else {
+            /* We didn't parse a daos path,
+            * but we weren't necessarily looking for one */
+            rc = 1;
+        }
     }
-
+out:
+    mfu_free(&tmp_path1);
+    mfu_free(&tmp_path2);
     mfu_free(&dattr.da_rel_path);
-
+    duns_destroy_attr(&dattr);
     return rc;
 }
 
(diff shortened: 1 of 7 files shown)


daltonbohning:
@dsikich Doing some testing of the new duns_resolve_path updates....
I'm seeing a hole in the way we are using duns_resolve_path.
A problem is that this path:
/path/to/uns1/new_dir
Does NOT resolve, because new_dir does not yet exist. duns_resolve_path doesn't bother recursing up the path if it doesn't exist.
I'm thinking we'll probably want to call duns_resolve_path on the dirname(path). And if that resolves, then concat basename(path) to the resolved path.
E.g.
```
tmp_path1 = strdup(path)
path_dirname = dirname(tmp_path1)
rc = duns_resolve_path(path_dirname)
if rc == 0
    tmp_path2 = strdup(path)
    path_basename = basename(tmp_path2)
    concat attr.da_rel_path / path_basename
```
The strdup usage for dirname and basename is necessary because those function implementation are wonky.

So maybe we should do something like this before removing `daos-prefix` from the documentation?

dsikich:
> @dsikich Doing some testing of the new duns_resolve_path updates....
> I'm seeing a hole in the way we are using duns_resolve_path.
> A problem is that this path:
> /path/to/uns1/new_dir
> Does NOT resolve, because new_dir does not yet exist. duns_resolve_path doesn't bother recursing up the path if it doesn't exist.
> I'm thinking we'll probably want to call duns_resolve_path on the dirname(path). And if that resolves, then concat basename(path) to the resolved path.
> E.g.
> 
> ```
> tmp_path1 = strdup(path)
> path_dirname = dirname(tmp_path1)
> rc = duns_resolve_path(path_dirname)
> if rc == 0
>     tmp_path2 = strdup(path)
>     path_basename = basename(tmp_path2)
>     concat attr.da_rel_path / path_basename
> ```
> 
> The strdup usage for dirname and basename is necessary because those function implementation are wonky.
> 
> So maybe we should do something like this before removing `daos-prefix` from the documentation?

@daltonbohning I thought that duns_resolve_path would succeed for something like /path/to/uns1/new_dir, because I thought it checked each level and if one succeeds then returns true. Since you tested it and this happens for directories that don't exist I can update. What is happening for regular files?

daltonbohning:
> @daltonbohning I thought that duns_resolve_path would succeed for something like /path/to/uns1/new_dir, because I thought it checked each level and if one succeeds then returns true. Since you tested it and this happens for directories that don't exist I can update. Did you try this for regular files?

If the path given doesn't exist at all, then it doesn't check the higher levels (parents). I.e. if stat fails, it doesn't check if the parent is a UNS entry:
https://github.com/daos-stack/daos/blob/861c98b25df27c3676f1367903da9eb3aacaae90/src/client/dfs/duns.c#L419-L425

It seems the behavior is the same for regular files: if it doesn't already already exist then resolve fails, but if it does exist, then resolve works as expected. But I don't think our logic needs to explicitly handle directories vs files. I think it can just blindly use dirname and then append basename to the rel_path

dsikich:
> > @daltonbohning I thought that duns_resolve_path would succeed for something like /path/to/uns1/new_dir, because I thought it checked each level and if one succeeds then returns true. Since you tested it and this happens for directories that don't exist I can update. Did you try this for regular files?
> 
> If the path given doesn't exist at all, then it doesn't check the higher levels (parents). I.e. if stat fails, it doesn't check if the parent is a UNS entry:
> https://github.com/daos-stack/daos/blob/861c98b25df27c3676f1367903da9eb3aacaae90/src/client/dfs/duns.c#L419-L425
> 
> It seems the behavior is the same for regular files: if it doesn't already already exist then resolve fails, but if it does exist, then resolve works as expected. But I don't think our logic needs to explicitly handle directories vs files. I think it can just blindly use dirname and then append basename to the rel_path

@daltonbohning I see, yeah passing dirname to duns_resolve_path then appending should work. I will update this today.

dsikich:
@daltonbohning the calling of duns_resolve_path has been updated in the case that the last file/dir does not exist

src/common/mfu_daos.c
```diff
@@ -293,10 +294,37 @@ int daos_parse_path(
...
+        path_basename = basename(tmp_path2);
+        
+        /* make sure there is enough memory allocated to concat basename */
+        tmp = realloc(dattr.da_rel_path, path_len);
+        if (tmp == NULL) {
+            rc = ENOMEM;
+        }
+        dattr.da_rel_path = tmp;
+        strcat(dattr.da_rel_path, "/");
+        strcat(dattr.da_rel_path, path_basename);
```
daltonbohning, line 326: @dsikich If tmp==NULL here then it wouldn't be safe to call these strcat functions. And the caller expects a negative rc to be an actual error.
To remove the need for this extra error handling, maybe instead of calling realloc at all, we could just append basename directly below? (line 331) Either with strcat, asprintf, etc.

dsikich: @daltonbohning not sure what you mean, I don't think appending directly to da_rel_path is safe? Thats fine, I can return there.

daltonbohning: @dsikich What I mean is that instead of creating a copy of rel_path and appending the basename, and then doing this:
https://github.com/hpc/mpifileutils/blob/f6d2143d5dc669f002c4bffebc426df827771a71/src/common/mfu_daos.c#L331-L335
We could instead just directly append the basename to `path`.

src/common/mfu_daos.c
```diff
@@ -319,7 +319,7 @@ int daos_parse_path(
         /* make sure there is enough memory allocated to concat basename */
         tmp = realloc(dattr.da_rel_path, path_len);
         if (tmp == NULL) {
-            rc = ENOMEM;
+            return -ENOMEM;
```
daltonbohning, line 322: @dsikich We would need to also make sure we free everything alloc'ed in this function before returning. Maybe we can add a `goto` to the cleanup at the bottom of this function?

src/common/mfu_daos.c
dsikich: @daltonbohning copying over the path string seems to work ok with strncpy but using strcat causes memory corruption errors

dsikich: @daltonbohning I updated this again though, this way also seems to work but it requires using strncpy

daltonbohning: @dsikich 
> copying over the path string seems to work ok with strncpy but using strcat causes memory corruption errors

Oh, right. Because the string isn't null terminated (or zero'ed out).

src/common/mfu_daos.c
```diff
@@ -323,9 +324,58 @@ int daos_parse_path(
...
+            strcat(dattr.da_rel_path, "/");
+            strcat(dattr.da_rel_path, path_basename);
+        } else if (strncmp(path, "daos:", 5) == 0) {
+            /* Actual error, since we expect a daos path */
+            rc = -1;
+        } else {
+            /* We didn't parse a daos path,
+            * but we weren't necessarily looking for one */
+            rc = 1;
+        }
```
daltonbohning, line 377: Since this logic is here now, we don't need the other `else if..else` below, right?

daltonbohning (commented):
I think we should add an internal test for this scenario, and for fs copy if/once it has the same support

src/common/mfu_daos.c
dsikich: @daltonbohning well if the first duns_resolve_path succeeds then we still need to do the copy, but I removed some of the redundant code in my latest commit.

dsikich:
> @daltonbohning the calling of duns_resolve_path has been updated in the case that the last file/dir does not exist

@daltonbohning yeah I agree. I need to update mpifileutils RPM after this is merged. Then I plan to update fs copy but I guess I will just add the test for fs copy when I submit the PR for that if the RPM isn't merged yet. If it is merged I can add them both to the same PR.

daltonbohning (approved):

dsikich:
@adammoody does this look ok to you?

adammoody (approved):

adammoody:
Looks good to me.  Thanks @dsikich and @daltonbohning .  Happy Thanksgiving to you both!

dsikich:
> Looks good to me. Thanks @dsikich and @daltonbohning . Happy Thanksgiving to you both!

@adammoody thanks you too!

daltonbohning:
> Looks good to me. Thanks @dsikich and @daltonbohning . Happy Thanksgiving to you both!

You as well!

PR #489, opened 2021-06-24
daos: cleanup error reporting in serialize and deserialize

dsikich:
daos-serialize was using a broadcast to report the rc code
on rank zero, which should be an Allreduce. Also, daos-deserialize
was printing serialize instead of deserialize when reporting an error.


daltonbohning (approved):
Thanks @dsikich !

adammoody (approved):
Nice catch, @dsikich .

PR #492, opened 2021-07-26
daos: serialize new props and check cont status property

dsikich:
Check DAOS_PROP_CO_STATUS and do not serialize or copy
properties if status is unhealthy. Add force option
to serialize even if status is unhealthy, but never
copy with unhealthy status.

Serialize DAOS_PROP_CO_ROOTS and DAOS_PROP_CO_EC_CELL_SZ,
which were added after initial implementaion of
daos-serialize was completed. Also, remove calling of
dfs_cont_create in deserialization to avoid serialized
properties being overwritten. Since DFS API is not used
in serialization or deserialization it is better to instead
use daos_cont_create and pass in preserved properties to
the deserialization.

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -492,6 +492,42 @@ static int daos_get_cont_type(
     return 0;
 }
 
+/* check if cont status is unhealthy */
+static int daos_check_cont_status(daos_handle_t coh, bool *status_healthy)
+{
+    daos_prop_t*            prop = daos_prop_alloc(1);
+    struct daos_prop_entry  *entry;
+    struct daos_co_status   stat = {0};
+    int                     rc = 0;
+    
+    if (prop == NULL) {
+        MFU_LOG(MFU_LOG_ERR, "Failed to allocate prop: "DF_RC, DP_RC(rc));
+        rc = ENOMEM;
+        goto out;
+    }
+
+    prop->dpp_entries[0].dpe_type = DAOS_PROP_CO_STATUS;
+
+    rc = daos_cont_query(coh, NULL, prop, NULL);
+    if (rc) {
+        MFU_LOG(MFU_LOG_ERR, "daos_cont_query() failed "DF_RC, DP_RC(rc));
+        rc = 1;
+        goto out;
+    }
+
+    entry = &prop->dpp_entries[0];
+    daos_prop_val_2_co_status(entry->dpe_val, &stat);                            
+    if (stat.dcs_status == DAOS_PROP_CO_HEALTHY) {
+        *status_healthy = true;
+    } else {
+        *status_healthy = false;
+    }
+
+out:
+    daos_prop_free(prop);
+    return rc;
+}
+
 /*
  * Try to set the file type based on the container type,
  * using api as a guide.
@@ -619,16 +655,20 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
  * Returns 1 on error, 0 on success.
  */
 static int cont_get_props(daos_handle_t coh, daos_prop_t** _props,
-                          bool get_oid, bool get_label)
+                          bool get_oid, bool get_label, bool get_roots)
 {
     int             rc;
     daos_prop_t*    props = NULL;
     daos_prop_t*    prop_acl = NULL;
     daos_prop_t*    props_merged = NULL;
-    uint32_t        num_props = 14;
+    /* total amount of properties to allocate */
+    uint32_t        total_props = 15;
+    /* minimum number of properties that are always allocated/used to start
+     * count */
+    int             prop_index = 15;
 
     if (get_oid) {
-        num_props++;
+        total_props++;
     }
 
     /* container label is required to be unique, so do not
@@ -636,11 +676,15 @@ static int cont_get_props(daos_handle_t coh, daos_prop_t** _props,
      * serialization, but only deserialized if the label
      * no longer exists in the pool */
     if (get_label) {
-        num_props++;
+        total_props++;
+    }
+
+    if (get_roots) {
+        total_props++;
     }
 
     /* Allocate space for all props except ACL. */
-    props = daos_prop_alloc(num_props);
+    props = daos_prop_alloc(total_props);
     if (props == NULL) {
         MFU_LOG(MFU_LOG_ERR, "Failed to allocate container properties.");
         rc = 1;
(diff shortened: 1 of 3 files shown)


src/common/mfu_daos.c
```diff
@@ -619,18 +654,22 @@ static int daos_any_error(int rank, bool local_daos_error, int flag_daos_args)
...
+    uint32_t        num_props = 15;
 
     if (get_oid) {
         num_props++;
     }
 
+    if (get_roots) {
+        num_props++;
+    }
+
```
adammoody, line 672: Maybe move this block to come just after the ``if (get_label) num_props++;`` to keep with the same order of ``get_oid, get_label, get_roots`` that is used when setting values below.

adammoody (approved):

src/common/mfu_daos.c
```diff
@@ -4781,6 +4931,110 @@ static int cont_deserialize_prop_uint(struct hdf5_args* hdf5,
...
+        MFU_LOG(MFU_LOG_ERR, "failed to close attribute datatype");
+        rc = 1;
+        goto out;
+    }
+    status = H5Sclose(attr_dspace);
+    if (status < 0) {
+        MFU_LOG(MFU_LOG_ERR, "failed to close attribute dataspace");
+        rc = 1;
+        goto out;
+    }
```
daltonbohning, line 5032: Similar to above, we need to try to close in case of other failures

src/common/mfu_daos.c
```diff
@@ -5210,23 +5478,10 @@ int daos_cont_deserialize_connect(daos_args_t *daos_args,
...
-        }
-    } else {
-        rc = daos_cont_create(daos_args->src_poh, daos_args->src_cont_uuid, prop, NULL);
-        if (rc != 0) {
-            MFU_LOG(MFU_LOG_ERR, "failed to create container: %d", rc);
-            goto out;
-        }
+    rc = daos_cont_create(daos_args->src_poh, daos_args->src_cont_uuid, prop, NULL);
+    if (rc != 0) {
+        MFU_LOG(MFU_LOG_ERR, "failed to create container: %d", rc);
```
daltonbohning, line 5483:
```suggestion
        MFU_LOG(MFU_LOG_ERR, "failed to create container: "DF_RC, DP_RC(rc));
```

src/daos-serialize/daos-serialize.c
```diff
@@ -27,6 +27,7 @@ void print_usage(void)
     printf("Options:\n");
     printf("  -o  --output-path        - path to output serialized hdf5 files\n");
     printf("  -v, --verbose            - verbose output\n");
+    printf("  -q, --force              - force serialization even if container has unhealthy status\n");
```
daltonbohning, line 30:
```suggestion
    printf("  -f, --force              - force serialization even if container has unhealthy status\n");
```

src/daos-serialize/daos-serialize.c
```diff
@@ -150,9 +156,11 @@ int main(int argc, char** argv)
...
     tmp_rc = daos_connect(rank, daos_args, daos_args->src_pool_uuid,
                           daos_args->src_cont_uuid, &daos_args->src_poh,
-                          &daos_args->src_coh, true, false, false, false, NULL);
+                          &daos_args->src_coh, force_serialize, true,
+                          false, false, false, NULL);
     if (tmp_rc != 0) {
-        rc = 1;
+        /* don't continue if connect was unsuccessful */      
+        return 1;
     }
```
daltonbohning, line 164: We still need to call daos_fini(), mfu_finalize(), and MPI_Finalize();

src/common/mfu_daos.c
```diff
@@ -4781,6 +4931,110 @@ static int cont_deserialize_prop_uint(struct hdf5_args* hdf5,
...
+        goto out;
+    }
+    root_oids = MFU_CALLOC(attr_dims[0], sizeof(obj_id_t));
+    if (root_oids == NULL) {
+        rc = ENOMEM;
+        goto out;
+    }
+    /* freed in daos_prop_free in daos_cont_deserialize_connect */
+    roots = MFU_CALLOC(1, sizeof(struct daos_prop_co_roots));
+    if (roots == NULL) {
```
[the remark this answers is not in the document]
dsikich: @daltonbohning the memory is allocated and the entry->due_val_ptr points to the memory allocated here (roots struct). If I free in this function then when the props is passed anywhere the due_val_ptr for the roots property will not be pointing to anything. I can free it only after the prop has been passed to daos_cont_create.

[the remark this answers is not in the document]
daltonbohning: @dsikich Ah, I missed that assignment below

src/common/mfu_daos.c
```diff
@@ -662,14 +701,19 @@ static int cont_get_props(daos_handle_t coh, daos_prop_t** _props,
...
     }
 
     if (get_label) {
-        props->dpp_entries[15].dpe_type = DAOS_PROP_CO_LABEL;
+        props->dpp_entries[16].dpe_type = DAOS_PROP_CO_LABEL;
+    }
+
+    if (get_roots) {
+        props->dpp_entries[17].dpe_type = DAOS_PROP_CO_ROOTS;
     }
```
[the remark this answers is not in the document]
dsikich: @daltonbohning yeah this is a good point.

src/common/mfu_daos.c
```diff
@@ -1183,18 +1191,22 @@ int daos_connect(
 
         /* check container status, and if unhealthy do not continue unless --force
          * option is used (only for serialization, never for copies) */
-        rc = daos_check_cont_status(*coh);
-        if (rc != 0 && !force_serialize) {
+        bool status_healthy;
+        rc = daos_check_cont_status(*coh, &status_healthy);
+        if (rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "Checking DAOS container status failed: %s", strerror(rc));
```
daltonbohning, line 1197: This log might print an unhelpful strerror, since daos_check_cont_status returns 1 on most errors

src/common/mfu_daos.c
```diff
@@ -5427,7 +5423,7 @@ int cont_deserialize_all_props(struct hdf5_args *hdf5,
     }
 
     entry = &prop->dpp_entries[17];
-    rc = cont_deserialize_prop_roots(hdf5, entry, "DAOS_PROP_CO_ROOTS");
+    rc = cont_deserialize_prop_roots(hdf5, entry, "DAOS_PROP_CO_ROOTS", roots);
```
daltonbohning, line 5426: Does this function use both entry and roots?

dsikich: @daltonbohning yes, the entry->dpe_val_ptr points to the roots struct. For structs in the DAOS properties you have point these to some other memory (the special property struct like acl, roots, etc) you can't store it directly in properties like you can with dpe_val.

src/common/mfu_daos.c
dsikich: @daltonbohning would it be better to return the DAOS error code? Seems like a lot of the functions just return 1 on error, which yeah is not very helpful I guess.

daltonbohning (approved):
@dsikich Just FYI, I didn't run any of our functional tests against this branch.

src/common/mfu_daos.c
daltonbohning: @dsikich It's tricky, because it isn't always simple / straightforward to aggregate error codes across processes. And even then, we would need to adjust every call all the way up to daos-serialize.c to NOT just return 1.
In this case, maybe it makes sense for daos_check_cont_status itself to log a message with the corresponding error code, but still return 1. Then, this part of the code can just return without printing anything, or at least without printing an error code.

dsikich:
> @dsikich Just FYI, I didn't run any of our functional tests against this branch.

@daltonbohning so far I just tested copies and serializations manually because I had an issue running functional tests locally recently, but I found a workaround in one of the channels to try. I will try that and then run serialization and dcp ones.

src/common/mfu_daos.c
daltonbohning: Ultimately the idea is that the user shouldn't be confused about what the problem is

dsikich: @daltonbohning I updated this to not print an error code, and daos_check_cont_status prints the DAOS error that occurred but the function returns 1.

dsikich:
@daltonbohning @adammoody merging this I just needed to rebase into one commit, functional tests in DAOS were ran on dcp and the serialization tools for verification as well.

daltonbohning:
> @daltonbohning @adammoody merging this I just needed to rebase into one commit, functional tests in DAOS were ran on dcp and the serialization tools for verification as well.

@dsikich What you can also do when merging is "Squash and merge". It's confusing, because "Squash and merge" doesn't actually _merge_, it rebases! It does this:
1. Squash all commits into one commit
2. Push that commit to master (basically rebase, so no merge commits)

PR #498, opened 2021-09-16
dsync: report when sync is complete

ofaaland:
At INFO debug level, it would be useful to report when the dsync
finishes.  This allows the user to extract timing data after the fact,
for example when many dsyncs are run through some automated system.


src/dsync/dsync.c
```diff
@@ -3402,5 +3406,7 @@ int main(int argc, char **argv)
     mfu_finalize();
     MPI_Finalize();
 
+    MFU_LOG(MFU_LOG_INFO, "Completed sync");
+
```
adammoody, line 3410: Let's guard this with an ``if (rank == 0)`` check so that only one rank prints the message.

And to be a bit safer, let's move this to come before the ``mfu_finalize`` call.  I think the logging will still work after that call, but it's conceivable we might data structures in the logging that would be freed by the finalize call.

adammoody:
Thanks, @ofaaland .

src/dsync/dsync.c
ofaaland: Ah, I assumed that after MPI_Finalize that only one process existed.  But I definitely don't actually know that.

Thanks for the review, I've updated the patch.

PR #499, opened 2021-09-28
daos: negative test case bug fixes

dsikich:
This fixes bugs found in a test case that is currently being
skippin the DAOS CI. There were a couple of regressions that
need fixing. If a daos source string without a container was
passed in dcp was not reporting invalid arguments were passed.
Also, it should not matter if the source/dest pool or cont UUID
are upper or lower case. It was incorrectly reporting that the
DAOS source and destination were the same in that case.

    * properly check if a source pool and source cont are passed
    * allow passing in upper or lower case UUIDs for src and dst


src/common/mfu_daos.c
```diff
@@ -218,10 +218,32 @@ static int daos_check_args(
...
-    bool same_pool = (strcmp(da->src_pool, da->dst_pool) == 0);
-    bool same_cont = same_pool && (strcmp(da->src_cont, da->dst_cont) == 0);
+    int rc1, rc2 = 0;
+    bool same_pool = false;
+    uuid_t src_pool;
+    uuid_t dst_pool;
+    rc1 = uuid_parse(da->src_pool, src_pool); 
+    rc2 = uuid_parse(da->dst_pool, dst_pool); 
+    if (rc1 == 0 && rc2 == 0) {
+        same_pool = (strcasecmp(da->src_pool, da->dst_pool) == 0);
```
daltonbohning, line 232: Suggestion: For this case, since we know they are both UUIDs, and we've already parsed them, maybe it's clearer to use uuid_compare? https://linux.die.net/man/3/uuid_compare
But I'm okay with either.

src/common/mfu_daos.c
```diff
@@ -218,10 +218,32 @@ static int daos_check_args(
...
+    bool same_cont = false;
+    uuid_t src_cont;
+    uuid_t dst_cont;
+    rc1 = uuid_parse(da->src_cont, src_cont); 
+    rc2 = uuid_parse(da->dst_cont, dst_cont); 
+    if (rc1 == 0 && rc2 == 0) {
+        same_cont = same_pool && (strcasecmp(da->src_cont, da->dst_cont) == 0);
+    } else {
+        same_cont = same_pool && (strcmp(da->src_cont, da->dst_cont) == 0);
+    }
```
daltonbohning, line 246: Seeing this logic duplicated for both pool and container makes we wonder if we should have a helper function. But the code snippet is short enough, so unless we need this elsewhere I'm okay with it.

daltonbohning (approved):

dsikich:
@daltonbohning added a fix for when program was continuing to write to a DAOS container when it was out of space. All datamover/negative.py are passing now with these fixes.

daltonbohning (approved):

src/common/mfu_daos.c
```diff
@@ -218,10 +218,32 @@ static int daos_check_args(
         return 0;
     }
     
+    /* if passed in values for src/dst are both uuids ignore the case when comparing */
+
     /* Determine whether the source and destination
      * use the same pool and container */
-    bool same_pool = (strcmp(da->src_pool, da->dst_pool) == 0);
-    bool same_cont = same_pool && (strcmp(da->src_cont, da->dst_cont) == 0);
+    int rc1, rc2 = 0;
```
adammoody, line 225: It's a bit strange to initialize rc2 but not rc1 here.  Maybe initialize both or neither?  I'm fine either way.

adammoody (approved):
Looks good to merge, @dsikich

src/common/mfu_daos.c
dsikich: @adammoody oh yeah I just changed it to neither.

PR #502, opened 2021-10-08
daos: remove WITH_CART_PREFIX build option

dsikich:
The WITH_CART_PREFIX build option was added when the
DAOS and CART libraries were built under seperate paths
by DAOS. The libraries are now built under the same
path so this build variable is no longer necessary.


daltonbohning (approved):
LGMT. Just FYI that we'll want to remove WITH_CART_PREFIX from our internal spec as well.

PR #503, opened 2021-10-09
dsync and dcp: allow user to control syncing of xattrs

ofaaland:
Rather than always copying xattrs, introduce a new option
"--xattrs | -e" to allow the user to control how this is done.

Options are
  "none"        Copy no xattrs
  "all"         Copy all xattrs (prior default)
  "non-lustre"  Copy non-lustre xattrs (new default)
  "libattr"        Copy xattrs not indicated as "skip" in /etc/xattr.conf

Copying xattrs is now independent of --preserve for owner, perms,
etc.

Lustre uses xattrs to record important information such as how the file
data should be distributed among the Lustre servers.  As a result,
copying the lustre xattrs prevents certain default values from taking
effect.

If "non-lustre" option is in effect, xattrs named in lustre
source file lustre_idl.h as of this writing are not synced (see
XATTR_NAME_SOM and friends).  In addition, xattrs with prefix
"lustre" are not synced.

Signed-off-by: Olaf Faaland <faaland1@llnl.gov>

### DIFF
--- a/src/common/mfu_flist_copy.c
+++ b/src/common/mfu_flist_copy.c
@@ -17,6 +17,8 @@
 #include <getopt.h>
 #include <time.h> /* asctime / localtime */
 #include <regex.h>
+#include <attr/libattr.h>
+#include <attr/error_context.h>
 
 /* These headers are needed to query the Lustre MDS for stat
  * information.  This information may be incomplete, but it
@@ -330,18 +332,40 @@ static int mfu_copy_xattrs(
     /* iterate over list and copy values to new object lgetxattr/lsetxattr */
     if(got_list) {
         char* name = list;
+        struct error_context ctx;
 
         while(name < list + list_size) {
             /* start with a reasonable buffer,
              * allocate something bigger as needed */
             size_t val_bufsize = 1024;
             void* val = (void*) MFU_MALLOC(val_bufsize);
+            int copy_xattr;
 
             /* lookup value for name */
             ssize_t val_size;
             int got_val = 0;
 
-            while(! got_val) {
+            copy_xattr = 1; /* copy unless indicated below not to */
+            if (copy_opts->copy_xattrs == XATTR_USE_LIBATTR) {
+                if (attr_copy_action(name, &ctx) == ATTR_ACTION_SKIP) {
+                    copy_xattr = 0;
+                }
+            } else if (copy_opts->copy_xattrs == XATTR_SKIP_LUSTRE) {
+                /* ignore xattrs lustre treats specially */
+                /* list from lustre source file lustre_idl.h */
+                if (    strncmp(name,"lustre.",strlen("lustre.")) == 0 ||
+                        strcmp(name,"trusted.som") == 0 || strcmp(name,"trusted.lov") == 0 ||
+                        strcmp(name,"trusted.lma") == 0 || strcmp(name,"trusted.lmv") == 0 ||
+                        strcmp(name,"trusted.dmv") == 0 || strcmp(name,"trusted.link") == 0 ||
+                        strcmp(name,"trusted.fid") == 0 || strcmp(name,"trusted.version") == 0 ||
+                        strcmp(name,"trusted.hsm") == 0 || strcmp(name,"trusted.lfsck_bitmap") == 0 ||
+                        strcmp(name,"trusted.dummy") == 0)
+                {
+                    copy_xattr = 0;
+                }
+            }
+
+            while(! got_val && copy_xattr) {
                 errno = 0;
                 if (copy_opts->dereference) {
                     /* getxattr of dereferenced symbolic links */
@@ -390,7 +414,7 @@ static int mfu_copy_xattrs(
             }
 
             /* set attribute on destination object */
-            if(got_val) {
+            if(got_val && copy_xattr) {
                 errno = 0;
                 /* lsetxattr of symbolic link itself. No need to dereference here */
                 int setrc = mfu_file_lsetxattr(dest_path, name, val, (size_t) val_size, 0, mfu_dst_file);
@@ -923,9 +947,9 @@ static int mfu_copy_set_metadata_dirs(
 
 /* creates dir in destpath for specified item, identifies source path
  * that contains source dir, computes relative path to dir under source path,
- * and creates dir at same relative path under destpath, copies xattrs
- * when preserving permissions, which contains file striping info on Lustre,
- * returns 0 on success and -1 on error */
+ * and creates dir at same relative path under destpath, optionally copies
+ * xattrs (which contain striping information under Lustre), optionally
+ * preserves permissions, returns 0 on success and -1 on error */
 static int mfu_create_directory(
     mfu_flist list,                 /* flist holding target directory */
     uint64_t idx,                   /* index of target directory within its list */
@@ -993,7 +1017,7 @@ static int mfu_create_directory(
      * creating / striping files in the directory */
 
     /* copy extended attributes on directory */
-    if (copy_opts->preserve) {
+    if (copy_opts->copy_xattrs != XATTR_COPY_NONE) {
         int tmp_rc = mfu_copy_xattrs(list, idx, dest_path, copy_opts, mfu_src_file, mfu_dst_file);
         if (tmp_rc < 0) {
             rc = -1;
@@ -1166,8 +1190,8 @@ static int mfu_create_link(
         }
     }
 
-    /* set permissions on link */
-    if (copy_opts->preserve) {
+    /* set xattrs on link */
+    if (copy_opts->copy_xattrs != XATTR_COPY_NONE) {
         int xattr_rc = mfu_copy_xattrs(list, idx, dest_path, copy_opts, mfu_src_file, mfu_dst_file);
         if (xattr_rc < 0) {
             rc = -1;
@@ -1185,9 +1209,9 @@ static int mfu_create_link(
 
 /* creates inode in destpath for specified file, identifies source path
  * that contains source file, computes relative path to file under source path,
- * and creates file at same relative path under destpath, copies xattrs
- * when preserving permissions, which contains file striping info on Lustre,
- * returns 0 on success and -1 on error */
+ * and creates file at same relative path under destpath, optionally copies
+ * xattrs (which contain striping information under Lustre), optionally
+ * preserves permissions, returns 0 on success and -1 on error */
 static int mfu_create_file(
     mfu_flist list,
     uint64_t idx,
@@ -1240,7 +1264,7 @@ static int mfu_create_file(
     /* copy extended attributes, important to do this first before
      * writing data because some attributes tell file system how to
      * stripe data, e.g., Lustre */
-    if (copy_opts->preserve) {
+    if (copy_opts->copy_xattrs != XATTR_COPY_NONE) {
         int tmp_rc = mfu_copy_xattrs(list, idx, dest_path, copy_opts, mfu_src_file, mfu_dst_file);
         if (tmp_rc < 0) {
             rc = -1;
@@ -3253,6 +3277,9 @@ mfu_copy_opts_t* mfu_copy_opts_new(void)
 
     /* By default, don't bother to preserve all attributes. */
     opts->preserve = false;
+
+    /* By default, do not copy special to Lustre (which set striping) */
+    opts->copy_xattrs = XATTR_SKIP_LUSTRE;
 
     /* By default, don't dereference source symbolic links. 
      * This is not a perfect opposite of no_dereference */
--- a/src/common/mfu_param_path.c
+++ b/src/common/mfu_param_path.c
@@ -303,6 +303,31 @@ static void mfu_unpack_param(const char** pptr, mfu_param_path* param)
     return;
 }
 
+/*
+ * Parse an option string provided by the user to determine
+ * which xattrs to copy from source to destination.
+ */
+attr_copy_t parse_copy_xattrs_option(char *optarg)
+{
+    if (strcmp(optarg,"none") == 0) {
+        return XATTR_COPY_NONE;
+    }
+
+    if (strcmp(optarg,"non-lustre") == 0) {
+        return XATTR_SKIP_LUSTRE;
+    }
+
+    if (strcmp(optarg,"libattr") == 0) {
+        return XATTR_USE_LIBATTR;
+    }
+
+    if (strcmp(optarg,"all") == 0) {
+        return XATTR_COPY_ALL;
+    }
+
+    return XATTR_COPY_INVAL;
+}
+
 /**
  * Analyze all file path inputs and place on the work queue.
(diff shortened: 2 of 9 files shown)


adammoody:
@ofaaland , there are too many issues to track.  No problem that you weren't aware.  I just wanted to list them here because your work here will help address them.

We can also add these separately in dcp if you want.

We still have dcp1 and dcp because they implement two different algorithms.  Depending on the situation, one can be faster than the other.  Eventually, it'd be nice to roll both algorithms into one tool, but that's still future work.

src/dcp/dcp.c
```diff
@@ -150,6 +151,7 @@ int main(int argc, char** argv)
         {"daos-preserve"        , required_argument, 0, 'D'},
         {"input"                , required_argument, 0, 'i'},
         {"chunksize"            , required_argument, 0, 'k'},
+        {"copy-xattrs"          , required_argument, 0, 'e'},
```
adilger, line 154: minor nit - both tar and rsync use "--xattrs" (also short option '-X' for rsync) for this option.  Any reason not to be consistent?

src/dcp/dcp.c
```diff
@@ -231,6 +234,23 @@ int main(int argc, char** argv)
                     }
                 }
                 break;
+            case 'e':
+                copy_xattrs_used = 1;
+                if (!strcmp(optarg,"none")) {
```
adilger, line 239: Not sure what the coding guidelines are for MPIFileutils, but (IMHO) the usage of "!strcmp()" can be misread, as it looks like it is checking "if not a good string compare".  For Lustre we prefer to use "strcmp() == 0", but by no means am I trying to push my code style upon you, just a thought.

src/dcp/dcp.c
```diff
@@ -231,6 +234,23 @@ int main(int argc, char** argv)
...
+                    mfu_copy_opts->copy_xattrs = XATTR_COPY_NONE;
+                } else if (!strcmp(optarg,"non-lustre")) {
+                    mfu_copy_opts->copy_xattrs = XATTR_SKIP_LUSTRE;
+                } else if (!strcmp(optarg,"libattr")) {
+                    mfu_copy_opts->copy_xattrs = XATTR_USE_LIBATTR;
+                } else if (!strcmp(optarg,"all")) {
+                    mfu_copy_opts->copy_xattrs = XATTR_COPY_ALL;
+                } else {
+                    if (rank == 0) {
+                        MFU_LOG(MFU_LOG_ERR, "Unrecognized option for --copy-xattrs");
```
adilger, line 249: It error message would be more useful if it printed the actual bad option inside single quotes, in case it is accidentally a filename or similar.

src/dcp/dcp.c
```diff
@@ -332,6 +352,14 @@ int main(int argc, char** argv)
         }
     }
 
+    /* copy_xattrs only vaid with preserve */
+    if (mfu_copy_opts->preserve == false && copy_xattrs_used == 1) {
+        if (rank == 0) {
+            MFU_LOG(MFU_LOG_ERR, "Copy-xattrs may only be used with --preserve");
```
adilger, line 358: IMHO, errors like this are counter productive for users.  I just means if the user _wants_ to copy xattrs, they will hit an error, and then have to fix what the program could have done itself - enable the preserve option if --xattrs is used.

src/dcp/dcp.c
```diff
@@ -231,6 +234,23 @@ int main(int argc, char** argv)
                     }
                 }
                 break;
+            case 'e':
+                copy_xattrs_used = 1;
+                if (!strcmp(optarg,"none")) {
+                    mfu_copy_opts->copy_xattrs = XATTR_COPY_NONE;
+                } else if (!strcmp(optarg,"non-lustre")) {
```
adilger, line 241: For code like this, I've found it more useful to have an array of { value, "string" } for all of the options (with an empty entry at the end), and then loop over the array looking for the matching string to get the value (possibly in a small helper function), or in the case of the usage message loop over the array and print all of the strings.  That avoids duplicating the string/value mapping a few times (as is done here and dsync.c), and avoids future errors if a new type is added but the usage message is not updated, or not updated in both programs, or similar.

adilger (commented):
None of my comments are particularly related to the functionality being added here (which looks fine to me, the patch seems rather straight forward).  They are mostly generic comments about code style, and trying to minimize/avoid potential future bugs related to the changes being made here.

src/dcp/dcp.c
ofaaland: Yes, --xattrs / -X would be better

man/dcp.1
```diff
@@ -60,6 +60,17 @@ The default chunksize is 64MB.
...
+Copy extended attributes ("xattrs") from source files to target files.  WHICH
+determines which xattrs are copied.  Options are to copy no xattrs, all xattrs,
+xattrs not excluded by /etc/xattr.conf, or all xattrs except those which have
+special meaning to lustre.  Certain xattrs control lustre features on a
+file-by-file basis, such as how the file data is distributed across Lustre
+servers.  Values must be in {none, all, libattr, non-lustre}.  The default is
+non-lustre.
+.UNINDENT
+.INDENT 0.0
+.TP
```
adammoody, line 73: Would you please move this to ``doc/rst/dcp1.rst``.  We auto-generate these man pages from the rst source files.

Let's also capitalize "Lustre" in the user-facing text.

src/common/mfu_flist_copy.c
```diff
@@ -330,18 +332,44 @@ static int mfu_copy_xattrs(
...
             int got_val = 0;
 
-            while(! got_val) {
+            if (copy_opts->copy_xattrs == XATTR_COPY_ALL) {
+                copy_xattr = 1;
+            } else if (copy_opts->copy_xattrs == XATTR_COPY_NONE) {
+                copy_xattr = 0;
+            } else if (copy_opts->copy_xattrs == XATTR_USE_LIBATTR) {
+                if (attr_copy_action(name, &ctx) == ATTR_ACTION_SKIP)
+                    copy_xattr = 0;
```
adammoody, line 354: Coding style: we try to always use braces, even for one-liners.
```suggestion
                if (attr_copy_action(name, &ctx) == ATTR_ACTION_SKIP) {
                    copy_xattr = 0;
                }
```

man/dcp.1
adammoody: I can help generate the man pages from there.

src/common/mfu_param_path.c
```diff
@@ -303,6 +303,27 @@ static void mfu_unpack_param(const char** pptr, mfu_param_path* param)
...
+        return XATTR_COPY_NONE;
+
+    if (strcmp(optarg,"non-lustre") == 0)
+        return XATTR_SKIP_LUSTRE;
+
+    if (strcmp(optarg,"libattr") == 0)
+        return XATTR_USE_LIBATTR;
+
+    if (strcmp(optarg,"all") == 0)
+        return XATTR_COPY_ALL;
```
adammoody, line 322: style: let's include braces through here, too

src/dcp/dcp.c
```diff
@@ -79,6 +79,7 @@ void print_usage(void)
 #endif
     printf("  -b, --bufsize <SIZE>     - IO buffer size in bytes (default " MFU_BUFFER_SIZE_STR ")\n");
     printf("  -k, --chunksize <SIZE>   - work size per task in bytes (default " MFU_CHUNK_SIZE_STR ")\n");
+    printf("  -e, --xattrs <OPT>       - copy xattrs (none, all, non-lustre, libattr)\n");
```
adammoody, line 82: Do you mean to use ``-X`` now?
```suggestion
    printf("  -X, --xattrs <OPT>       - copy xattrs (none, all, non-lustre, libattr)\n");
```

src/dcp/dcp.c
```diff
@@ -90,7 +91,7 @@ void print_usage(void)
     printf("  -i, --input <file>       - read source list from file\n");
     printf("  -L, --dereference        - copy original files instead of links\n");
     printf("  -P, --no-dereference     - don't follow links in source\n");
-    printf("  -p, --preserve           - preserve permissions, ownership, timestamps, extended attributes\n");
+    printf("  -p, --preserve           - preserve permissions, ownership, timestamps (see also --copy-xattrs)\n");
```
adammoody, line 94: Do you mean this instead?
```suggestion
    printf("  -p, --preserve           - preserve permissions, ownership, timestamps (see also --xattrs)\n");
```

src/dsync/dsync.c
```diff
@@ -3047,7 +3049,7 @@ int main(int argc, char **argv)
 
     while (1) {
         int c = getopt_long(
-            argc, argv, "b:cDso:LPSvqh",
+            argc, argv, "b:cDe:so:LPSvqh",
```
adammoody, line 3052: Do you mean to have ``X:`` in here instead of ``e:``?

adammoody:
Thanks again @ofaaland .  Sorry to be so tardy on this.  Great work.  I found a few things to check, but otherwise it looks good.

We use sphinx to generate our man pages from the rst files.  I can help you set that up.

man/dcp.1
ofaaland: Cool, I didn't notice the rst files.  So much easier to read and edit than nroff or whatever that is.

src/common/mfu_flist_copy.c
ofaaland: Done

src/dcp/dcp.c
ofaaland: Yes

src/dcp/dcp.c
ofaaland: Yep

src/dsync/dsync.c
ofaaland: Yes

src/common/mfu_param_path.c
ofaaland: done

ofaaland:
Hi @adammoody , I fixed up the issues you found, and also added a test.  I ran the test bash file by hand, rather than using nose2 and the python test-runner because I'm not sure how it works - but it should work, I've included a python wrapper copied from the dcp test.

I found doc/README.md and followed the instructions to setup and run Sphinx.  It ran, and it updated the man pages to reflect my .rst updates, but it also made a bunch of markup changes that are unrelated to my work.  So perhaps the version of Sphinx is different or there's some other Sphinx configuration I haven't done.  Here's an example:
```
 .TP
 .B \-b, \-\-blocksize SIZE
 Set the compression block size, from 1 to 9.
-Where 1=100kB <E2><80><A6> and 9=900kB. Default is 9.
+Where 1=100kB ... and 9=900kB. Default is 9.
 .UNINDENT
 .INDENT 0.0
 .TP
```

Because I don't know the reason for those markup changes, I've discarded the changes Sphinx made for now.  Let me know how to proceed.

Thanks!

ofaaland:
dsync test output:
```
Testing dsync
PASSED verify of option none for /mnt/xfs/dest type xfs
PASSED verify of option all for /mnt/xfs/dest type xfs
SKIPPED verify of option non-lustre
PASSED verify of option libattr for /mnt/xfs/dest type xfs
```

adammoody:
Thanks @ofaaland .  That's great!

I just gave you a couple of files that I used to install sphinx and then build the man pages.  It's been a while since I tried myself, so it could just be due to a newer sphinx version like you say.

We could also merge without the changes to the man files and generate the new man pages in a later commit if you want.  I often only update the rst files in my commits and forget all about needing to regenerate the man pages until we stamp a new release.

src/dsync/dsync.c
```diff
@@ -63,6 +63,7 @@ static void print_usage(void)
     printf("  -b  --batch-files <N>   - batch files into groups of N during copy\n");
     printf("      --bufsize <SIZE>    - IO buffer size in bytes (default " MFU_BUFFER_SIZE_STR ")\n");
     printf("      --chunksize <SIZE>  - minimum work size per task in bytes (default " MFU_CHUNK_SIZE_STR ")\n");
+    printf("  -e, --xattrs <OPT>      - copy xattrs (none, all, non-lustre, libattr)\n");
```
adammoody, line 66: We've got ``e`` instead of ``X`` here, too

src/dcp/dcp.c
```diff
@@ -168,7 +170,7 @@ int main(int argc, char** argv)
     int usage = 0;
     while(1) {
         int c = getopt_long(
-                    argc, argv, "b:d:g:i:k:LPpsSvqh",
+                    argc, argv, "b:d:e:g:i:k:LPpsSvqh",
```
adammoody, line 173: And an ``X:`` here?

src/dsync/dsync.c
ofaaland: oops, fixed.

src/dcp/dcp.c
ofaaland: Yep

ofaaland:
OK, I fixed up those issues, removed a couple of whitespace issues that snuck in, and hand-tested dcp to make sure -X works.  I think it's ready to go.

I think it makes sense to update the man pages in a separate commit, in case there are other unrelated man page changes.  But I'll take your sphinx files and see if I still get markup changes.

Thanks @adammoody

ofaaland:
Looks like the markup changes are not due to the process I followed - I get the same results using your scripts.  There is at least one other content change (for DAOS) to the .rst files that isn't yet reflected in the man pages, so I think a man-page-only commit makes sense.

src/common/mfu_flist_copy.c
```diff
@@ -993,7 +1024,7 @@ static int mfu_create_directory(
      * creating / striping files in the directory */
 
     /* copy extended attributes on directory */
-    if (copy_opts->preserve) {
+    if (copy_opts->copy_xattrs) {
```
adammoody, line 1027: It seems like this condition might always be true, since I think ``XATTR_COPY_NONE=1`` in the enum.  I see the function handles NONE correctly, though it might best to avoid calling all together when disabled.  Maybe replace this with ``if (copy_opts->copy_xattrs != XATTR_COPY_NONE)``?

ofaaland: Yes, good catch.  I forgot to change it after I switched copy_xattrs from a boolean to an enum, but the mistake was hidden because the function handled NONE correctly.

I changed that condition there and in the other two places mfu_copy_xattrs() is called as you suggested.  I also simplified the logic within mfu_copy_xattrs(), and then tested to confirm it still worked.

adammoody:
Thanks, @ofaaland .  This is good to go, but we've got a conflict on some rst files due to a change that snuck in.  Would you please update to resolve those, and then we can merge?

ofaaland:
Thanks @adammoody.  Rebased on master and fixed the conflicts.

adammoody:
Really nice addition, @ofaaland !

PR #509, opened 2021-12-06
daos: remove duplicate error message

daltonbohning:
if dfs_sys_mount fails, don't call dfs_sys2base


adammoody (approved):
Thanks, @daltonbohning

dsikich (commented):
@daltonbohning thanks, LGTM.

PR #511, opened 2021-12-30
add cmake check for libattr

adammoody:
This adds a cmake find_package check for libattr, which is used for ``attr_copy_action`` when copying extended attributes.  When xattr support is enabled, libattr is marked as required.  However, this also adds compile-time checks in the code to disable libattr code if is was not found so that it could be optional rather than required.


adammoody:
@ofaaland , would you please review this when you get a chance?  If you have suggestions, please let me know.

In case someone doesn't have libattr installed, I protected the libattr-related lines with a ``HAVE_LIBATTR`` check.  For now, I've left the cmake check as required, but we could likely back that off to make it optional.

I also dropped the input ``ctx`` variable and replaced it with NULL since we weren't using it anyway.  To use, I think we'd want to fill in the fields of the struct with pointers to functions, as in the example here:

https://github.com/philips/attr/blob/1cc88bd4c17ef99ace22c8be362d513f155b1387/examples/copyattr.c#L89

ofaaland (approved):
LGTM
I think we should leave LibAttr as "REQUIRED" for the moment since there is no easy way for the user to tell whether libattr support was built (if they don't know about tools like ldd)

src/common/mfu_flist_copy.c
```diff
@@ -17,8 +17,11 @@
 #include <getopt.h>
 #include <time.h> /* asctime / localtime */
 #include <regex.h>
+
+#ifdef HAVE_LIBATTR
 #include <attr/libattr.h>
 #include <attr/error_context.h>
```
ofaaland, line 23: I think we don't need this include anymore.

ofaaland (changes requested):

src/common/mfu_flist_copy.c
adammoody: Good idea.  I dropped it.

daltonbohning:
@adammoody To be clear, the current (**_and intended_**) behavior is that `libattr` is required when `ENABLE_XATTRS=ON`?

adammoody:
@daltonbohning , yes, that's the goal, but we're assuming that one has libattr available.  If we find common setups that use xattrs but do not have libattr installed, I think we could do more work to make libattr optional.

daltonbohning:
> @daltonbohning , yes, that's the goal, but we're assuming that one has libattr available. If we find common setups that use xattrs but do not have libattr installed, I think we could do more work to make libattr optional.

@adammoody Thanks. Luckily, so far it seems our (DAOS) setups do have libattr installed/available.

PR #518, opened 2022-02-01
prepare for v0.11.1 release

adammoody:
Final commits to prepare for v0.11.1 release.


adammoody:
@daltonbohning , @dsikich , my earlier email may have bounced back to me with an error.  We're working to stamp a v0.11.1 release within the next week or so.  We have a need for the Lustre feature added to ``dsync``.  Let me know if you'd like to push any changes before that.

There is a draft release notes page.  Please feel free to add any DAOS related content there, too.  We can also update that page after the release, so no need to rush if you're busy.

adammoody:
@daltonbohning , and do we need to update the library version numbering?

adammoody:
@ofaaland , we'll aim to make this release in about a week.  I have a draft release notes page started.

daltonbohning:
@adammoody I can't think of anything DAOS-related that _**needs**_ to be in v0.11.1. I'll take a look at the release notes, but I'm not confident I'll have all the content (@dsikich would know better but is away).
As for the library version:
https://github.com/hpc/mpifileutils/blob/86fbacce8b773f8dc7ed42bee9c8a68cf997a4d8/src/common/CMakeLists.txt#L2-L5
I think it should be bumped from `3.0.0` to `4.0.0`. For `mfu_daos.h` at least, some function signatures are not backwards compatible.

adammoody:
Thanks, @daltonbohning .  I bumped the library version to 4.0.0.  In case you have to do that again in the future, we actually have two places to do that now.  There is a second related, but different ``dist/CMakeLists.txt`` for the release tarball.  We might be able to move common code to a cmake include to avoid that in the future, but just so you know...

adammoody:
@dsikich , @daltonbohning , @ofaaland ,  I just posted the v0.11.1 release.  As always, great work putting that together.

PR #531, opened 2022-05-27
DAOS-9902 API: To support empty container lable

wiliamhuang:
Signed-off-by: Lei Huang <lei.huang@intel.com>

### DIFF
--- a/src/common/mfu_daos.c
+++ b/src/common/mfu_daos.c
@@ -3857,10 +3857,12 @@ static int cont_serialize_prop_str(struct hdf5_args* hdf5,
     hid_t   attr_dspace;
     hid_t   usr_attr;
 
-    if (entry == NULL || entry->dpe_str == NULL) {
-        MFU_LOG(MFU_LOG_ERR, "Property %s not found", prop_str);
-        rc = 1;
-        goto out;
+    if (entry == NULL || entry->dpe_str == NULL ) {
+        if ( !(entry != NULL && entry->dpe_str == NULL && entry->dpe_type == DAOS_PROP_CO_LABEL)) {
+            MFU_LOG(MFU_LOG_ERR, "Property %s not found", prop_str);
+            rc = 1;
+            goto out;
+        }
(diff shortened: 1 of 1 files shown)


src/common/mfu_daos.c
```diff
@@ -3857,10 +3857,12 @@ static int cont_serialize_prop_str(struct hdf5_args* hdf5,
...
-    if (entry == NULL || entry->dpe_str == NULL) {
-        MFU_LOG(MFU_LOG_ERR, "Property %s not found", prop_str);
-        rc = 1;
-        goto out;
+    if (entry == NULL || entry->dpe_str == NULL ) {
+        if ( !(entry != NULL && entry->dpe_str == NULL && entry->dpe_type == DAOS_PROP_CO_LABEL)) {
+            MFU_LOG(MFU_LOG_ERR, "Property %s not found", prop_str);
+            rc = 1;
+            goto out;
+        }
```
daltonbohning, line 3865: Really, this was an oversight during original implementation. If DAOS gives us a NULL/empty string for ANY property, then we just trust that. The client code here doesn't need to enforce any business rules. So, we could simplify to something like...
```suggestion
    if (entry == NULL) {
        MFU_LOG(MFU_LOG_ERR, "Property %s not found", prop_str);
        rc = 1;
        goto out;
```

daltonbohning (changes requested):

src/common/mfu_daos.c
wiliamhuang: Thank you! Just made the change you suggested.

daltonbohning (approved):

daltonbohning:
@adammoody I went ahead and landed since the changes are contained to `mfu_daos.c`. In general, are you okay with this, as long DAOS changes are contained to DAOS-only source files?

adammoody:
Yes, that's fine to go ahead a merge daos-specific items.  Thanks for checking, @daltonbohning .

PR #539, opened 2022-10-21
daos: remove --daos-prefix handling

daltonbohning:
This was previously deprecated and is longer needed. The new way is to use a UNS path:
daos://pool/container/path/to/stuff


adammoody (approved):

daltonbohning:
Thanks @adammoody !

PR #546, opened 2023-03-23
Add in initial support for HPSSFS-FUSE

hawartens:
When the destination filesystem is detected to be "fuse" and the subtype is "hpssfs" we know that we are writing to an HPSS archive using HPSSFS-FUSE. HPSS archives have a concept called "Class of Service" (COS) that defines properties for files that live within that COS. Unfortunately a single "Class of Service" for an HPSSFS-FUSE filesystem is defined at mount time, so all files that are written to the filesystem will all go to the same COS. In many cases, sites have COSes that are based on the size of the files (eg. COSA has small files, COSB has medium sized files, and COSC has large files). If we used mpifileutils to write to an HPSSFS-FUSE filesystem without support for HPSS, the files copied into HPSS would then all be added to the same COS and many of the files would not live where they were intended. An HPSS administrator or the HPSS ChangeCOS process would then have to move the data to the correct COS on the backend (meaning this data was copied twice). In order to avoid this second move of the data, this work allows mpifileutils to send a file size hint, via an ioctl() to the HPSS Core Server so that the file will be copied to the correct COS immediately.


CMakeLists.txt
```diff
@@ -57,6 +57,12 @@ IF(ENABLE_LUSTRE)
   # - LUSTRE_STAT
 ENDIF(ENABLE_LUSTRE)
 
+OPTION(ENABLE_HPSS "Enable optimization and features for HPSSFS-FUSE" OFF)
+MESSAGE(STATUS "ENABLE_HPSS: ${ENABLE_HPSS}")
+IF(ENABLE_HPSS)
+  ADD_DEFINITIONS(-DHPSS_SUPPORT)
```
adammoody, line 63: Do we need to add any ``FIND_PACKAGE`` here to set include or lib paths?  It doesn't look like we make any library calls, so perhaps just the header is sufficient.

adammoody:
@hawartens , sorry for the delay.  I took a pass on this today.  These are just style edits and calling library functions for allocating/freeing memory.  I was trying to push the changes back to your original branch, but losing my fight with git.  Instead, I pushed to the main repo on a new branch.  

https://github.com/hpc/mpifileutils/compare/main...hpss-support

When you get a chance, would you please confirm that I didn't break anything?

Do the changes I made to init/finalize look valid to you?

src/common/mfu_util.c
```diff
@@ -1105,6 +1109,284 @@ bool mfu_is_lustre(const char* path)
...
+
+    if (mfu_mi_list_initialized == false) {
+        MFU_LOG(MFU_LOG_INFO, "The filesystem list has not been initialized...");
+        return(false);
+    }
+
+    errno = 0;
+    rc = stat(path, &st);
+    if (rc == -1) {
+        fprintf(stderr, "Could not stat: %s (%d)", path, errno);
```
adammoody, line 1359: Let's use ``MFU_LOG(MFU_LOG_ERR, ...`` here.

src/common/mfu_util.c
```diff
@@ -1105,6 +1109,284 @@ bool mfu_is_lustre(const char* path)
...
+            fclose(mountfp);
+            return(ENOMEM);
+        }
+        memset(itemp, 0, sizeof(struct mfu_mount_info));
+        INIT_LIST_HEAD(&itemp->mfu_mi_linkage);
+
+        /*
+         * Parse one of the lines in /proc/self/mounts
+         */
+        mfu_mount_info_parse_mount_line(line, len, itemp);
```
adammoody, line 1258: Let's handle any parse error here, in case someone has a bad line in their file.  In particular, I think we should avoid calling the subsequent ``stat()`` if we failed to parse the line and maybe skip adding an entry in the list.

src/common/mfu_util.c
hawartens: Sounds good. I missed this!

src/common/mfu_util.c
hawartens: My initial goal here was to try and be as lenient as possible. Basically just wanted to have a simple data structure that included everything in "/proc/self/mounts". If that possibly had a bad entry, then I was okay with it being reflected in the list. I am totally fine with whatever you prefer here for handling parse errors.

hawartens:
> @hawartens , sorry for the delay. I took a pass on this today. These are just style edits and calling library functions for allocating/freeing memory. I was trying to push the changes back to your original branch, but losing my fight with git. Instead, I pushed to the main repo on a new branch.
> 
> [main...hpss-support](https://github.com/hpc/mpifileutils/compare/main...hpss-support)
> 
> When you get a chance, would you please confirm that I didn't break anything?
> 
> Do the changes I made to init/finalize look valid to you?


I think it looks good to me! I have a couple comments. Will try to put them inline in your newer branch.

PR #548, opened 2023-05-11
Add in initial support for HPSSFS-FUSE

adammoody:
From @hawartens :
""
When the destination filesystem is detected to be "fuse" and the subtype is "hpssfs" we know that we are writing to an HPSS archive using HPSSFS-FUSE. HPSS archives have a concept called "Class of Service" (COS) that defines properties for files that live within that COS. Unfortunately a single "Class of Service" for an HPSSFS-FUSE filesystem is defined at mount time, so all files that are written to the filesystem will all go to the same COS. In many cases, sites have COSes that are based on the size of the files (eg. COSA has small files, COSB has medium sized files, and COSC has large files). If we used mpifileutils to write to an HPSSFS-FUSE filesystem without support for HPSS, the files copied into HPSS would then all be added to the same COS and many of the files would not live where they were intended. An HPSS administrator or the HPSS ChangeCOS process would then have to move the data to the correct COS on the backend (meaning this data was copied twice). In order to avoid this second move of the data, this work allows mpifileutils to send a file size hint, via an ioctl() to the HPSS Core Server so that the file will be copied to the correct COS immediately.
""


src/common/mfu_util.c
```diff
@@ -1105,6 +1110,239 @@ bool mfu_is_lustre(const char* path)
...
+                                 mfu_mi_linkage)
+        {
+            list_del_init(&itemp->mfu_mi_linkage);
+            mfu_mount_info_destroy(itemp);
+            mfu_free(&itemp);
+        }
+        return;
+    }
+
+    mfu_mi_list_initialized = true;
```
hawartens, line 1276: I think mfu_mi_list_initialized should be set to true at line 1259. My original code accidentally had it set to true in both places accidentally. I know that the entire list has not been set up properly at that point, but if anything fails, then later on when we call mfu_destroy_filesystem_list(), nothing will actually get cleaned up.

src/common/mfu_util.c
```diff
@@ -1105,6 +1110,239 @@ bool mfu_is_lustre(const char* path)
...
+
+        /*
+         *  Get the device for the mountpoint in question.
+         */
+        errno = 0;
+        struct stat st;
+        int rc = stat(itemp->mfu_mi_mountpoint, &st);
+        if (rc == -1) {
+            fclose(mountfp);
+            fprintf(stderr, "Could not stat: %s (%d)", itemp->mfu_mi_mountpoint, errno);
```
hawartens, line 1234: I think you wanted the fprintf()’s changed to MFU_LOG(MFU_LOG_ERR, …)

src/common/mfu_util.c
```diff
@@ -1105,6 +1110,239 @@ bool mfu_is_lustre(const char* path)
...
+    if (mfu_mi_list_initialized == false) {
+        MFU_LOG(MFU_LOG_INFO, "The filesystem list has not been initialized...");
+        return(false);
+    }
+
+    errno = 0;
+    struct stat st;
+    int rc = stat(path, &st);
+    if (rc == -1) {
+        fprintf(stderr, "Could not stat: %s (%d)", path, errno);
```
hawartens, line 1315: I think you wanted the fprintf()’s changed to MFU_LOG(MFU_LOG_ERR, …)

src/common/mfu_util.c
```diff
@@ -1105,6 +1110,239 @@ bool mfu_is_lustre(const char* path)
...
+                             mfu_mi_linkage)
+    {
+        MFU_LOG(MFU_LOG_DBG, "MFU_MI_DEV: %d FILE_DEV: %d MFU_MI_FSTYPE: %s",
+            itemp->mfu_mi_dev, st.st_dev, itemp->mfu_mi_fstype
+        );
+
+        if ((itemp->mfu_mi_dev == st.st_dev) &&
+            (itemp->mfu_mi_fstype != NULL) &&
+            (itemp->mfu_mi_fssubtype != NULL) &&
+            (strncmp(itemp->mfu_mi_fstype, "fuse", 25) == 0) &&
```
hawartens, line 1333: That 25 is sticking out to me like a sore thumb. I really should have made a constant there. Anyhow I think it all looks great to me! Thanks for reviewing it and make it fit more with mpifileutils style.

src/common/mfu_util.c
adammoody: Good catch.  I'll update that.  And for errors, I also like to print the ``strerror(errno)``.

src/common/mfu_util.c
adammoody: Yep, I'll get that one, too.

src/common/mfu_util.c
adammoody: Ok, I moved this up.  I also set it back to false if the list gets deleted.  Would that still work?

src/common/mfu_util.c
adammoody: Feel free to update this if you want.  I'm good either way.

src/common/mfu_util.c
hawartens: Yes. I that looks fine to me!

src/common/mfu_util.c
hawartens: Okay. I'm good either way as well.

adammoody:
It looks good to me, @hawartens .  Good with you, too?

hawartens:
Yes. Looks good to me as well.

adammoody:
Thanks, @hawartens !

hawartens:
Thanks @adammoody! Much appreciated!

Issue #557, opened 2023-09-14
[Request] run ddup without changing atime


markmoe19:
Could ddup be enhanced for allowing a restore of last accessed time (atime) on a file after it is read for de-duplication comparison purposes?  We find atime useful for alerting users to unused files to clean-up.  We would also like to use ddup but it changes atime.

adammoody:
That's a good idea, and it's related to the ``O_NOATIME`` request in https://github.com/hpc/mpifileutils/pull/534 for reading files in dcp/dsync, which can be used to modifying the atime value.

adammoody:
As a test, can you check whether adding the ``O_NOATIME`` flag to the open call in ``ddup`` works:

https://github.com/hpc/mpifileutils/blob/47918154ea0f4895623f36ccf8cbfe2df477c3ae/src/ddup/ddup.c#L96

so that this line changes to
```
mfu_open(fname, O_RDONLY | O_NOATIME)
```

As noted in https://github.com/hpc/mpifileutils/pull/534, in addition to keeping atime the same, that could also improve read performance.  It cuts out a bunch of atime updates going to the Lustre metadata server.

I like this as a general improvement for ddup, but we I'll need to verify that this doesn't create problems for file systems that might not support the ``O_NOATIME`` flag.  If nothing else, we could enable the flag by default and add a  new command line option to drop it.

markmoe19:
I got the below error.  Looks like src/dcp1/compare.c has O_NOATIME in some commented out code, maybe it was considered there as well at one time?

/project/selene-admin/mpifileutils_debug/mpifileutils-v0.11.1/mpifileutils/src/ddup/ddup.c:102:41: error: ‘O_NOATIME’ undeclared (first use in this function)
  102 |     int fd = mfu_open(fname, O_RDONLY | O_NOATIME);
      |                                         ^~~~~~~~~

adammoody:
Oh, we probably also need to add a ``#define _GNU_SOURCE`` statement before any includes in order to pick up the definition for ``O_NOATIME``.  Can you try again after adding that to the very top of the ``ddup.c`` file?

markmoe19:
Ok, great, I got this to compile now, just need to test ...

Does dsync change atime?  This made we think that O_NOATIME would be nice to ddup, dsync, etc. tools as we don't want to change atime or mtime while looking for old files to clean-up or move.  :)

markmoe19:
[...]

PR #561, opened 2023-10-06
mfu: add --open-noatime to open files with O_NOATIME

adammoody:
This adds an ``--open-noatime`` option to a number of tools, which adds the ``O_NOATIME`` flag when opening files to avoid updating the file last access time.

Many centers use last access time to filter files for purge operations, and they would prefer not to change file atime values when making backup copies with ``dsync`` or scanning the file system for duplicate files with ``ddup``.  Adding this flag may also improve read performance on some file systems.

The ``O_NOATIME`` flag is only allowed when the effective user id matches the owner of the file or when the process is running with the ``CAP_FOWNER`` capability.  A normal user will encounter errors when using ``O_NOATIME`` when reading from a shared directory containing files owned by other users, even if the current user has read access to all files.

The following tools are affected:

``ddup`` - when reading files to compute hash values
``dcp`` and ``dsync`` - when reading source files during a copy
``dcmp`` and ``dsync`` - when reading source and destination files while comparing their contents
``dtar`` - while reading source files when creating an archive

Wheen ``--open-noatime`` is specified with ``ddup``, the tool checks the owner user id of each file and conditionally adds ``O_NOATIME`` if the process effective user id matches.  This allows normal users to specify the ``--open-noatime`` option, even when running ``ddup`` on files that they don't own.  The atime will be updated on files that the user can read but does not own.

For the remaining tools, the current algorithms do not expose the file owner id in a way to allow for an easy check.  In this case, ``O_NOATIME`` is added when opening *all* files.  Normal users will thus encounter an error if the tool attempts to open any file that they do not own.

Resolves:
https://github.com/hpc/mpifileutils/issues/557
https://github.com/hpc/mpifileutils/pull/534

### DIFF
--- a/doc/rst/dcmp.1.rst
+++ b/doc/rst/dcmp.1.rst
@@ -60,6 +60,10 @@ OPTIONS
 
    Use O_DIRECT to avoid caching file data.
 
+.. option:: --open-noatime
+
+   Open files with O_NOATIME flag.
+
 .. option:: --progress N
 
    Print progress message to stdout approximately every N seconds.
--- a/doc/rst/dcp.1.rst
+++ b/doc/rst/dcp.1.rst
@@ -82,6 +82,10 @@ OPTIONS
 
    Use O_DIRECT to avoid caching file data.
 
+.. option:: --open-noatime
+
+   Open files with O_NOATIME flag.
+
 .. option:: -S, --sparse
 
    Create sparse files when possible.
--- a/doc/rst/ddup.1.rst
+++ b/doc/rst/ddup.1.rst
@@ -20,6 +20,10 @@ Multiple sets of duplicate files can be matched using this final reported hash.
 OPTIONS
 -------
 
+.. option:: --open-noatime
+
+   Open files with O_NOATIME flag, if possible.
+
 .. option:: -d, --debug LEVEL
 
    Set verbosity level.  LEVEL can be one of: fatal, err, warn, info, dbg.
--- a/doc/rst/dsync.1.rst
+++ b/doc/rst/dsync.1.rst
@@ -83,6 +83,10 @@ OPTIONS
 
    Use O_DIRECT to avoid caching file data.
 
+.. option:: --open-noatime
+
+   Open files with O_NOATIME flag.
+
 .. option:: --link-dest DIR
 
    Create hardlink in DEST to files in DIR when file is unchanged
--- a/doc/rst/dtar.1.rst
+++ b/doc/rst/dtar.1.rst
@@ -119,6 +119,10 @@ OPTIONS
    Apply recorded flags to extracted files.
    Default does not record or extract flags.
 
+.. option:: --open-noatime
+
+   Open source files with O_NOATIME flag when creating archive.
+
 .. option:: --fsync
 
    Call fsync before closing files after writing.
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -15,6 +15,7 @@ LIST(APPEND libmfu_install_headers
   mfu_param_path.h
   mfu_path.h
   mfu_pred.h
+  mfu_proc.h
   mfu_progress.h
   mfu_util.h
   )
@@ -46,6 +47,7 @@ LIST(APPEND libmfu_srcs
   mfu_param_path.c
   mfu_path.c
   mfu_pred.c
+  mfu_proc.c
   mfu_progress.c
   mfu_util.c
   strmap.c
--- a/src/common/mfu.h
+++ b/src/common/mfu.h
@@ -27,6 +27,7 @@ extern "C" {
 #include "mfu_param_path.h"
 #include "mfu_flist.h"
 #include "mfu_pred.h"
+#include "mfu_proc.h"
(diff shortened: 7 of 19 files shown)


adammoody:
@daltonbohning , I hope you're doing well.

We've had some requests to add ``O_NOATIME`` to some tools.  I've opened this PR to do that.  It'd be good to know whether these changes are valid for DAOS system.

Is there someone who can help us check that?

daltonbohning:
> @daltonbohning , I hope you're doing well.
> 
> We've had some requests to add `O_NOATIME` to some tools. I've opened this PR to do that. It'd be good to know whether these changes are valid for DAOS system.
> 
> Is there someone who can help us check that?

Hey Adam. Yes, I'm doing well. Hope the same for you!

It looks like DAOS/DFS doesn't respect passing `O_NOATIME`, though atime is updated. I tested and it doesn't break anything. Extra bits set in `flags` are just ignored. So this change is safe for us, and I'll create an internal ticket to see if we want to handle that flag when passed.

Thanks for the heads up!

daltonbohning:
DAOS JIRA for reference: https://daosio.atlassian.net/browse/DAOS-14479

daltonbohning:
We discussed this in the context of DAOS, and we don't actually store atime with the file. It's only populated in the stat buf to the greater of mtime or ctime. So handling O_NOATIME wouldn't help anything because it would just get "reset" on the next file open.

daltonbohning:
We discussed this in the context of DAOS, and we don't actually store atime with the file. It's only populated in the stat buf to the greater of mtime or ctime. So handling O_NOATIME wouldn't help anything because it would just get "reset" on the next file open.

daltonbohning:
We discussed this in the context of DAOS, and we don't actually store atime with the file. It's only populated in the stat buf to the greater of mtime or ctime. So handling O_NOATIME wouldn't help anything because it would just get "reset" on the next file open.

adammoody:
TODO: we'll need to be a bit more clever when copying files that are readable but not owned by the user.

From ``man 2 open``:

> O_NOATIME (since Linux 2.6.8)
>   Do not update the file last access time (st_atime in the inode) when the file is read(2).
> 
>   This flag can be employed only if one of the following conditions is true:
>   *  The effective UID of the process matches the owner UID of the file.
>   *  The calling process has the CAP_FOWNER capability in its user namespace and the owner UID of the file has a mapping in the namespace.
> 
>   This  flag is intended for use by indexing or backup programs, where its use can significantly reduce the amount of disk activity.  This flag may not be effective on all filesystems.  One example is NFS, where the server maintains the access time.

and potential error:

> EPERM  The O_NOATIME flag was specified, but the effective user ID of the caller did not match the owner of the file and the caller was not privileged.

adammoody:
> We discussed this in the context of DAOS, and we don't actually store atime with the file. It's only populated in the stat buf to the greater of mtime or ctime. So handling O_NOATIME wouldn't help anything because it would just get "reset" on the next file open.

Thanks, @daltonbohning .  And thanks for your super fast response!

adammoody:
It sounds like ``tar`` updates source file atimes by default but one can _attempt_ to preserve atime with an option:

https://www.gnu.org/software/tar/manual/html_section/Attributes.html

> When tar reads files, it updates their access times. To avoid this, use the ‘--atime-preserve[=METHOD]’ option, which can either reset the access time retroactively or avoid changing it in the first place.

adammoody:
For the ability to use ``O_NOATIME``, we do a similar check in ``mfu_flist_chmod()``:

https://github.com/hpc/mpifileutils/blob/47918154ea0f4895623f36ccf8cbfe2df477c3ae/src/common/mfu_flist_chmod.c#L1258-L1260

https://github.com/hpc/mpifileutils/blob/47918154ea0f4895623f36ccf8cbfe2df477c3ae/src/common/mfu_flist_chmod.c#L1286-L1296

https://github.com/hpc/mpifileutils/blob/47918154ea0f4895623f36ccf8cbfe2df477c3ae/src/common/mfu_flist_chmod.c#L1069-L1075

adammoody:
Apparently, rsync v3.2.0 provides the following options for atime:

https://download.samba.org/pub/rsync/rsync.1

```
       --atimes, -U
              This  tells  rsync to set the access (use) times of the destina‐
              tion files to the same value as the source files.

              If repeated, it also sets the --open-noatime option,  which  can
              help you to make the sending and receiving systems have the same
              access times on the transferred files  without  needing  to  run
              rsync an extra time after a file is transferred.

              Note  that  some  older rsync versions (prior to 3.2.0) may have
              been built with a pre-release --atimes patch that does not imply
              --open-noatime when this option is repeated.

       --open-noatime
              This  tells rsync to open files with the O_NOATIME flag (on sys‐
              tems that support it) to avoid changing the access time  of  the
              files  that  are being transferred.  If your OS does not support
              the O_NOATIME flag then rsync will silently ignore this  option.
              Note  also  that  some filesystems are mounted to avoid updating
              the atime on read access even without the O_NOATIME  flag  being
              set.
```

Tip from:  https://unix.stackexchange.com/questions/630228/rsync-keep-access-time-atime-how

Issue #547, opened 2023-04-07
RHEL8 build


MarbolanGos:
Dear HPC mpifileutils tool maintenor,

For a lustre installation, I was asked to build on RHEL8.6 this tool via the specfile you provide.

Here is the file I add to use:
```specfile
Name:           mpifileutils
Version:        v0.11.1
Release:        1%{?dist}
Summary:        File utilities designed for scalability and performance.

Group:          System Environment/Libraries
License:        Copyright and BSD License
URL:            https://hpc.github.io/mpifileutils
Source:         %{name}-%{version}.tar.gz
BuildRoot:      %_topdir/BUILDROOT
Requires: libcircle, lwgrp, dtcmp, libarchive, openssl, openssl-devel

%description
File utilities designed for scalability and performance.

%prep
%setup -q

%build
#topdir=`pwd`
#installdir=$topdir/install

cmake ./ -DWITH_DTCMP_PREFIX=${installdir} -DWITH_LibCircle_PREFIX=${installdir} -DCMAKE_INSTALL_PREFIX=%{buildroot}/usr -DENABLE_LUSTRE=ON -DENABLE_XATTRS=ON
make

%install
rm -rf %{buildroot}
make install
#make install DESTDIR=$RPM_BUILD_ROOT


%files
%defattr(-,root,root,-)
%{_bindir}/*
%{_includedir}/*
%{_libdir}/*
%{_mandir}/*
%{_datadir}/*

%changelog

```` 

Also, I had to use this commandline:
```bash
QA_SKIP_BUILD_ROOT=1  rpmbuild --nodebuginfo   -ba mpifileutils.spec
```

I was a bit in a hurry to get the tool available so removing debuginfo and not having to use QA is not a good idea so maybe it is possible to be better int he compilation... Also, those tips can help others who need to build this tool!

I prefer not asking for a PR as the modifications are not clean but maybe it is possible to work on updating the file to meet the RHEL requirements.

adammoody:
Thank you, @MarbolanGos .  I'll take a look and work to update the spec file.

PR #564, opened 2023-11-08
rpm: use %cmake macros

rezib:
Documentation about these macros is available in Fedora project packaging guidelines:
https://docs.fedoraproject.org/en-US/packaging-guidelines/CMake/

One notable advantage of these macros is that they transparently define compiler and linker flags with distribution defaults.

### DIFF
--- a/mpifileutils.spec
+++ b/mpifileutils.spec
@@ -17,15 +17,12 @@ File utilities designed for scalability and performance.
 %setup -q
 
 %build
-#topdir=`pwd`
-#installdir=$topdir/install
-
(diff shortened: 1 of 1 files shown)


cedeyn:
Seems to properly fix #547

adammoody:
Thank you, @rezib .

adammoody:
One more addition I see in https://github.com/hpc/mpifileutils/issues/547 is that it adds some build-time dependencies.

```
BuildRequires: cmake, libattr-devel, libarchive-devel, bzip2-devel, lustre-client-devel, openssl-devel
```

I'm guessing those are needed in clean build environments.  Would adding those interfere with your build?

rezib:
> I'm guessing those are needed in clean build environments. Would adding those interfere with your build?

Absolutely not! We added these build requirements as well in our in-house spec file actually. We simply didn't contribute this change as we also include other build requirements that may not be relevant in the general case.

Thank you for the merge!

Issue #578, opened 2024-07-30
Building with -DENABLE_XATTRS=OFF fails with "call to undeclared function"


Michael-Hennecke:
Building on SLES 15.5 and Intel MPI fails with "undeclared function" errors when using -DENABLE_XATTRS=OFF.
The respective function calls should probably be wrapped in some sort of "#ifdef XATTR"?

module load gcc intel-toolkit/2024.1.0 intel-mpi/2021.12.0 cmake

cmake ../mpifileutils \
  -DCMAKE_INSTALL_PREFIX=../install \
  -DWITH_DAOS_PREFIX=/usr \
  -DENABLE_DAOS=ON \
  -DENABLE_XATTRS=OFF

make -j install
...
/dss/dsshome1/09/di29puz3/dev/mpifileutils-v0.11.1/mpifileutils/src/common/mfu_io.c:1509:18: error: call to undeclared function 'llistxattr'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
 1509 |     ssize_t rc = llistxattr(path, list, size);
      |                  ^
/dss/dsshome1/09/di29puz3/dev/mpifileutils-v0.11.1/mpifileutils/src/common/mfu_io.c:1546:18: error: call to undeclared function 'listxattr'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
 1546 |     ssize_t rc = listxattr(path, list, size);
      |                  ^
/dss/dsshome1/09/di29puz3/dev/mpifileutils-v0.11.1/mpifileutils/src/common/mfu_io.c:1583:18: error: call to undeclared function 'lgetxattr'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
 1583 |     ssize_t rc = lgetxattr(path, name, value, size);
      |                  ^
/dss/dsshome1/09/di29puz3/dev/mpifileutils-v0.11.1/mpifileutils/src/common/mfu_io.c:1618:18: error: call to undeclared function 'getxattr'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
 1618 |     ssize_t rc = getxattr(path, name, value, size);
      |                  ^
/dss/dsshome1/09/di29puz3/dev/mpifileutils-v0.11.1/mpifileutils/src/common/mfu_io.c:1654:14: error: call to undeclared function 'lsetxattr'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
[...]

PR #582, opened 2024-08-22
MFU-578: add xattr.h header to mfu_io.c

Michael-Hennecke:
include the xattr.h header in mfu_io.c to avoid compilation errors (ISO C99 and later do not support implicit function declarations)

### DIFF
--- a/src/common/mfu_io.c
+++ b/src/common/mfu_io.c
@@ -2,6 +2,7 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/xattr.h>
 #include <unistd.h>
 
 #include <fcntl.h>


daltonbohning (changes requested):
If we do this I think it still needs to be behind the xattr compile flag.

E.g.
https://github.com/hpc/mpifileutils/blob/0a4c530bca5dc2476418a3833a15ab812d636e78/src/common/mfu_flist_archive.c#L35-L37

Then, the real fix to your #578 is to also wrap the uses of the xattr functions.
A workaround is to just build with `-DENABLE_XATTRS=ON`

ofaaland:
@Michael-Hennecke @daltonbohning can we close this since #583 was merged, which added the `#include <sys/xattr.h>` to mfu_io.c?

daltonbohning:
> @Michael-Hennecke @daltonbohning can we close this since #583 was merged, which added the `#include <sys/xattr.h>` to mfu_io.c?

Yep, thanks!

daltonbohning:
@Michael-Hennecke You should just be able to build `main` now with no additional patches

PR #590, opened 2024-10-18
Add option to not update the file last access time with dwalk

GauthierEvd:
This is a proposal to add --no-atime option on dwalk to not update the file last access time while doing a walk without stat.

The motivation behind this patch is that at CEA we have tools to administrate our filesystems and we want to use mpifileutils in these tools. We don't want the atime of a folder to be modified when admins walk a filesystem. For us, updating the atime is a user-side action and should therefore not be modified by our tools.

### DIFF
--- a/src/common/mfu_flist_walk.c
+++ b/src/common/mfu_flist_walk.c
@@ -52,6 +52,7 @@ static int SET_DIR_PERMS;
 static int REMOVE_FILES;
 static int DEREFERENCE;
 static int WALK_RESULT = 0;
+static int NO_ATIME;
 static mfu_file_t** CURRENT_PFILE;
 
 /****************************************
@@ -182,11 +183,15 @@ struct linux_dirent {
 
 static void walk_getdents_process_dir(const char* dir, CIRCLE_handle* handle)
 {
+    int flags = O_RDONLY | O_DIRECTORY;
     char buf[BUF_SIZE];
+
+    if (NO_ATIME)
+        flags |= O_NOATIME;
 
     /* TODO: may need to try these functions multiple times */
     mfu_file_t* mfu_file = *CURRENT_PFILE;
-    mfu_file_open(dir, O_RDONLY | O_DIRECTORY, mfu_file);
+    mfu_file_open(dir, flags, mfu_file);
     if (mfu_file->fd == -1) {
         /* print error */
         MFU_LOG(MFU_LOG_ERR, "Failed to open directory for reading: `%s' (errno=%d %s)", dir, errno, strerror(errno));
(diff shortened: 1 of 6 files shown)


src/common/mfu_flist_walk.c
```diff
@@ -657,11 +668,15 @@ void mfu_flist_walk_paths(uint64_t num_paths, const char** paths,
...
     }
     else {
-        /* walk directories using file types in readdir */
-        CIRCLE_cb_create(&walk_readdir_create);
-        CIRCLE_cb_process(&walk_readdir_process);
-        //        CIRCLE_cb_create(&walk_getdents_create);
-        //        CIRCLE_cb_process(&walk_getdents_process);
+        if (walk_opts->no_atime) {
+            /* walk directories without updating the file last access time */
+            CIRCLE_cb_create(&walk_getdents_create);
```
ofaaland, line 673: I agree this if/else block works, but what's the value in having both `walk_getdents_create()` and `walk_readdir_create()`?  Their code is identical, it seems, except for whether they call `walk_getdents_process_dir()` or `walk_readdir_process_dir()`.

ofaaland (commented):
Thanks for your contribution.  I think I have only one question.

src/common/mfu_flist_walk.c
GauthierEvd: I agree with you, I hadn't realized that the functions `walk_getdents_create()` and `walk_readdir_create()` were identical. The same question applies to the functions `walk_getdents_process` and `walk_readdir_process` which are also identical except for whether they call `walk_readdir_process_dir()` or `walk_getdents_process_dir()`.

I think it's best to make a generic function that will call the desired function depending on whether you have the noatime option. This way, the if/else block is no longer necessary.
`

src/dwalk/dwalk.c
```diff
@@ -432,6 +434,9 @@ int main(int argc, char** argv)
             case 'p':
                 print = 1;
                 break;
+            case 'n':
+                walk_opts->no_atime = 1;
```
ofaaland, line 438: In the man page you say `Must be used with --lite option.`. Do you ensure that's true somewhere below?

ofaaland (commented):
Looks good, just one question

src/dwalk/dwalk.c
GauthierEvd: Yes it is managed by the condition `if(walk_opts->use_stat)` in `mfu_flist_walk_paths()`, if the --lite option is not set, we will not use the path which uses the NO_ATIME for opening directories even if the option has been set.

ofaaland: I see, thanks.  I personally would be in favor of an error or warning to the user, either in dwalk after the `getopt_long()` loop is finished, or perhaps in `mfu_flist_walk_paths()`, if the user runs without `--lite` and with `--no-atime`.  I can see a new user overlooking the "use with -l" admonishment, or forgetting about it.

I can see arguments either way, though, so if you disagree you can just mark this conversation resolved.

ofaaland (approved):
LGTM.  Thanks!

ofaaland:
Further conversation about whether the default should be not to update atime, is here: https://github.com/hpc/mpifileutils/commit/5aac10e45563d9af02b086ee4c9c623b01fe29c7#commitcomment-149752222

PR #592, opened 2024-10-23
build.rst shell script: stop if dependency download fails

ofaaland:
The developer build instructions include some shell scripting to download, build, and install dependencies (e.g. libcircle, lwgrp).  That scripting assumes paths to dependency release tarballs that may be out-of-date.

If a dependency tarball download fails, tell the user which one failed and where to look, and do not proceed to building and installing.

After a release these docs may quickly become out-of-date, and so it can save some developer time to fail faster and give a clearer error message.

### DIFF
--- a/doc/rst/build.rst
+++ b/doc/rst/build.rst
@@ -133,34 +133,51 @@ which can be done with the following commands:
 
    mkdir deps
    cd deps
-     wget https://github.com/hpc/libcircle/releases/download/v0.3/libcircle-0.3.0.tar.gz
-     wget https://github.com/llnl/lwgrp/releases/download/v1.0.4/lwgrp-1.0.4.tar.gz
-     wget https://github.com/llnl/dtcmp/releases/download/v1.1.4/dtcmp-1.1.4.tar.gz
-     wget https://github.com/libarchive/libarchive/releases/download/v3.5.1/libarchive-3.5.1.tar.gz
 
-     tar -zxf libcircle-0.3.0.tar.gz
-     cd libcircle-0.3.0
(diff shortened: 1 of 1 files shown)


doc/rst/build.rst
```diff
@@ -133,10 +133,25 @@ which can be done with the following commands:
...
+     for url in ${urls[*]}
+     do
+             wget $url
+             rc=$?
+             if [[ $rc -ne 0 ]];
+             then
+                     echo
+                     echo FAILED getting $url
+                     echo check for releases under $(echo $url | sed 's/releases.*/releases\//')
+                     exit 1
```
daltonbohning, line 152: Is it intended that the user put these commands in a script and then execute? I have often copy-pasted the commands directly from the site, so `exit 1` will be a logout in that case.

daltonbohning (approved):
I ran as a sanity check and it did work as expected

doc/rst/build.rst
ofaaland: That's a good point, I hadn't thought of the copy-paste.  I'll think of a different way to handle this.

ofaaland: OK, I pushed an updated commit that uses if/else to stop executing once there is a failure to download.  It's now complex enough that I worry it obscures the basic process for installing the deps.
  @daltonbohning or @carbonneau1 do you have an opinion?

carbonneau1 (approved):

doc/rst/build.rst
ofaaland: In another channel, @carbonneau1 said "I don't think it's that bad. I think it is better organized."

PR #593, opened 2024-10-24
fixes to test_dsync/test_xattr.sh and test_dsync/test_xattr.py

ofaaland:
Fixes two issues:
1. The command line generated in test_xattr.py to execute test_xattr.sh and run the test was wrong.  Fix that.
2. The output of "xattr -l" needs to be sorted for the xattr comparison to work consistently

Tested on lustre and xfs

### DIFF
--- a/test/tests/test_dsync/test_xattr.py
+++ b/test/tests/test_dsync/test_xattr.py
@@ -11,6 +11,6 @@ dsync_dest_dir   = "/mnt/lustre2"
 dsync_test_file  = "file_test_xattr_XXX"
 
 def test_xattr():
-        p = subprocess.Popen(["%s %s %s %s %s %s %s" % (dsync_test_bin,
+        p = subprocess.Popen(["%s %s %s %s %s" % (mpifu_path, dsync_test_bin,
           dsync_src_dir, dsync_dest_dir, dsync_test_file)], shell=True,
           executable="/bin/bash").communicate()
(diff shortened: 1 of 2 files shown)


carbonneau1 (approved):
We no longer copy extended attributes by default, right?

ofaaland:
Updated to change the commit messages to clarify this affects a test under test_dsync/

ofaaland:
> We no longer copy extended attributes by default, right?

Based on the manpage `dsync.1` discussion of `--xattr` and the commit message for adding that option, 96a9fe7d2f49a9d6b3e28941bd51f33bef8af8ce, we used to copy all xattrs, and now the default is to copy non-lustre xattrs.

This test doesn't actually check what dsync does if `--xattr` is not provided, so I'm not sure.  I'll look.

ofaaland:
Altered the test so it tests dsync without `--xattr` as well as with `--xattr {none,all,non-lustre}`

ofaaland:
> > We no longer copy extended attributes by default, right?
> 
> Based on the manpage `dsync.1` discussion of `--xattr` and the commit message for adding that option, [96a9fe7](https://github.com/hpc/mpifileutils/commit/96a9fe7d2f49a9d6b3e28941bd51f33bef8af8ce), we used to copy all xattrs, and now the default is to copy non-lustre xattrs.
> 
> This test doesn't actually check what dsync does if `--xattr` is not provided, so I'm not sure. I'll look.

@carbonneau1  I checked, and dsync copies non-lustre xattrs by default.  I modified the test to check that functionality.

carbonneau1 (approved):

PR #594, opened 2024-10-30
mfu: copy much more xattrs

### LINKED ISSUE #553: [Request] add support for project id (i.e. lsattr -p)
dwalk supports group and gid.  Could it also support project and project id?  See "/usr/bin/lsattr -p" for project id linux attribute on a file.  We make use of gid for access and project id is often the same but sometimes different and used more for quota control.  Thank you.

KoyamaSohei:
When copying files using dcp or dsync, extended attributes such as project id are also copied.

Resolves #553.


src/common/mfu_flist_copy.c
```diff
@@ -443,6 +447,37 @@ static int mfu_copy_xattrs(
...
+        int dest_fd = open(dest_path, O_RDONLY);
+        if (dest_fd >= 0) {
+            struct fsxattr attr;
+            /* get src's xattr */
+            rc = ioctl(src_fd, FS_IOC_FSGETXATTR, &attr);
+            if (rc == 0) {
+                /* set dest's xattr */
+                rc = ioctl(dest_fd, FS_IOC_FSSETXATTR, &attr);
+                if (rc != 0) {
+                    MFU_LOG(MFU_LOG_ERR, "ioctl failed: `%s' errno=%d (%s)", dest_path, errno, strerror(errno));
```
ofaaland, line 463: nit: it might help users if the message stated which ioctl failed

src/common/mfu_flist_copy.c
```diff
@@ -443,6 +447,37 @@ static int mfu_copy_xattrs(
...
+                }
+            } else {
+                MFU_LOG(MFU_LOG_ERR, "ioctl failed: `%s' errno=%d (%s)", src_path, errno, strerror(errno));
```
ofaaland, line 466: nit: it might help users if the message stated which ioctl failed

src/common/mfu_flist_copy.c
```diff
@@ -443,6 +447,37 @@ static int mfu_copy_xattrs(
     mfu_free(&list);
     list_bufsize = 0;
 
+    /* open the src */
```
ofaaland, line 450: style: this new code should be moved to a helper function. It will allow the code to be a little simpler, exiting early if one of the syscalls fails, and will simplify reading mfu_copy_xattrs() when someone is looking for a bug unrelated to the FSGETXATTR/FSSETXATTR code.

src/common/mfu_flist_copy.c
```diff
@@ -443,6 +447,37 @@ static int mfu_copy_xattrs(
...
+    if (src_fd >= 0) {
+        /* open the dest */
+        int dest_fd = open(dest_path, O_RDONLY);
+        if (dest_fd >= 0) {
+            struct fsxattr attr;
+            /* get src's xattr */
+            rc = ioctl(src_fd, FS_IOC_FSGETXATTR, &attr);
+            if (rc == 0) {
+                /* set dest's xattr */
+                rc = ioctl(dest_fd, FS_IOC_FSSETXATTR, &attr);
```
ofaaland, line 461: Is it safe to write all of `attr` like this?  For example, are fsx_nextents and fsx_extesize guaranteed not to change, if src_path is on one type of FS, like xfs, and dest_path is on another, like nfs?

Looking at https://man7.org/linux/man-pages/man2/ioctl_xfs_fsgetxattr.2.html it sounds to me like many of the flags and fields may be supported on some file systems but not others, or worse may be interpreted differently on one file system than on another.  e.g.

```
fsx_extsize is the preferred extent allocation size for data
blocks mapped to this file, in units of filesystem blocks.
```

ofaaland (commented):
Thanks for your contribution @KoyamaSohei, I have one concern.

src/common/mfu_flist_copy.c
ofaaland: I would be more comfortable copying only specific flags or fields, such as `fsx_projid` and `fsx_xflags` bit `FS_XFLAG_PROJINHERIT`

ofaaland: > Is it safe to write all of attr like this? For example, are fsx_nextents and fsx_extesize guaranteed not to change, if src_path is on one type of FS, like xfs, and dest_path is on another, like nfs?

What I mean to say is, "Should fsx_nextents and fsx_extesize be changed depending on the type of the destination file system?  Are there flags in fsx_xflags that must not be set for some file system types?"

KoyamaSohei:
@ofaaland Could you please review again?

I fixed to copy only fsx_projid and FS_XFLAG_PROJINHERIT

src/common/mfu_flist_copy.c
ofaaland: @KoyamaSohei please move the entire new block of code you're adding to a helper function.  It looks to me like you just need to pass in the source and destination path, and return an `int` indicating success or failure.

ofaaland (changes requested):
@KoyamaSohei Thanks for this improvement.

Please do some testing to and report the performance impact of your code change, preferably on a parallel file system like Lustre which may have high throughput but higher latency than something local like xfs.

Please also modify your commit message to reflect that mfu is copying projectid ; the commit message still makes it sounds like it's copying many xattrs.

KoyamaSohei:
We compared the execution time when copying lustre-release 10 times on Lustre using dcp.

There was a significant decrease in performance ...

| without projid | with projid |
|----------|----------|
| 6.730 | 35.203 |
| 6.661 | 16.111 |
| 6.381 | 12.757 |
| 6.189 | 9.895 |
| 6.023 | 7.868 |
| 8.082 | 11.188 |
| 9.612 | 16.068 |
| 10.529 | 20.939 |
| 10.450 | 15.777 |
| 11.078 | 9.029 |

ofaaland:
> We compared the execution time when copying lustre-release 10 times on Lustre using dcp.
> 
> There was a significant decrease in performance ...
> without projid 	with projid
> 6.730 	35.203
> 6.661 	16.111
> 6.381 	12.757
> 6.189 	9.895
> 6.023 	7.868
> 8.082 	11.188
> 9.612 	16.068
> 10.529 	20.939
> 10.450 	15.777
> 11.078 	9.029

I think we should try and fix that before we merge this.  For our site, for example, we don't yet use projectid at all, so this would be a big performance hit for no benefit.

I wonder if this is because we are sort-of opening each file twice, once for `getxattr(2)` and once for `open(2)` in preparation for the `ioctl(FS_IOC_FSGETXATTR)`.  If that's true, refactoring the code to to call `open` once and then use that fd for both `fgetxattr` and `ioctl` might improve performance.

Before you do that refactoring work, though, maybe you can figure out with some careful experiments whether that's where the time is going, or whether we're spending it elsewhere.

PR #596, opened 2024-11-06
test_dsync: add test_expectfail.sh, test_existence.sh, and stubs for two other tests

ofaaland:
Add the following tests:
* test_expectfail.sh: add tests for missing source or destination parent
* test_existence.sh: verify set of files synced is correct

Add stubs for the following tests:
* test_metadata.sh: verify metadata of synced files is correct
* test_data.sh: verify data in synced files is correct

Add a small set of bash functions for performing set operations, for use with dsync tests.

Add a small test driver and two data files for testing the bash functions.


ofaaland:
@gonsie @carbonneau1 I intend this to be a template for other tests, so this is a good time to discuss or object to anything about how I implemented these tests - writing tests as shell scripts, the organization of the tests, the fact that they're single-node tests, etc.  Thanks!

test/tests/test_dsync/utility/one
carbonneau1, line 1: What is this file for?

ofaaland: > Add a small test driver and two data files for testing the bash functions.

carbonneau1 (approved):

ofaaland:
Hi @carbonneau1 please see the two new commits, running dsync under "mpirun", and copying a directory tree from /usr/include for use as test data.

test/tests/test_dsync/test_data.sh
```diff
@@ -0,0 +1,31 @@
...
+##############################################################################
+# Description:
+#
+#   Verify dsync handles data
+#     - file data is identical after a copy
+#     - file with differing data is copied if --contents arg is used
+#     - file with differing data is not copied if --contents arg is not used and metadata match
+#
+# Notes:
+#   - does not test whether data copes are spread across nodes/tasks evenly
```
daltonbohning, line 12:
```suggestion
#   - does not test whether data copies are spread across nodes/tasks evenly
```

daltonbohning (approved):
Mostly a cursory overview. I think bash here is fine, since that's the easiest interface.

test/tests/test_dsync/test_data.sh
ofaaland: Fixed

Issue #587, opened 2024-10-09
dsync deletes destination files after getting errors while walking the source


defaziogiancarlo:
I've seen multiple instances of dsync reporting errors while walking the source tree. In this case they are "transport endpoint shutdown" error on lustre due to clients being evicted from some of the OSTs. At the end of the source walk, the number of items reported is incorrect, and much lower than it should be due to the errors during the walk. Then, after the walking the destination, dsync determines that they are many excess items on the destination and deletes them. However, those items are not excess, they were just not found on the source due to the errors.

My suggestion for a solution is to do what rsync does, which is to not delete things on the destination if there were errors while reading the source, even if the --delete flag is specified.

v0.11.1
this is running on TOSS 4, which is based on RHEL using the 4.18.0-553.22.1 kernel
I'm actually not sure which mpi it was compiled with
I can't provide logs, but I'll try to reproduce this in an environment where I can provide the logs

ofaaland:
> I can't provide logs, but I'll try to reproduce this in an environment where I can provide the logs

No need.  Eric can do that himself :)

ofaaland:
@defaziogiancarlo please see https://github.com/hpc/mpifileutils/pull/599

PR #598, opened 2024-11-11
Draft: test_walkfail.sh: implement src walk fail test

ofaaland:
Draft only - *do not merge*

Implement a test where dsync --delete is run, and the walk of the
source tree is caused to fail by making a directory unreadable.

This test is currently expected to fail.

dsync currently has a bug described in
https://github.com/hpc/mpifileutils/issues/587

After the source walk fails, dsync continues with the dest walk
followed by the copies that the deletes.

Because the source walk failed, the list of files and directories
on the source is incomplete. Files or directories that are already
on both source and destination, but which were not walked on the
source, are missing from the source list.  Those files or directories
are then deleted from the destination tree.


ofaaland:
After https://github.com/hpc/mpifileutils/pull/596 is merged, I'll rebase this, and there will only be one commit - "test_walkfail.sh: implement src walk fail test"

PR #600, opened 2024-11-14
mfu_flist_walk.c: set WALK_RESULT = -1

ofaaland:
Fixes issues identified in the code review of #599

Since WALK_RESULT is an int and can only hold one error code, but any number of errors might be encountered during a walk, there is little value in recording errno.

In addition, the value of errno might have changed unless it is captured immediately after the library or system call that failed, which would require deeper code changes.

Instead, set to -1 as the indicator of a failure.


carbonneau1 (approved):
good catch!

PR #612, opened 2024-11-26
If an argument option returns an error we no longer try to load defa…

carbonneau1:
When passing a wrong expression/conjunction to the -o option a segmentation violation occured.
We were trying to free the conjunction twice.

Also once the error message was issued by the parser we did not check it before trying to set the default options causing an assert.

We now give the error messages an print the usage.

### DIFF
--- a/src/dsync/dsync.c
+++ b/src/dsync/dsync.c
@@ -2412,20 +2412,19 @@ static void dsync_disjunction_free(struct dsync_disjunction* disjunction)
     struct dsync_conjunction *conjunction;
     struct dsync_conjunction *n;
     
-    if (disjunction) {
-        assert(list_empty(&disjunction->linkage));
-        list_for_each_entry_safe(conjunction,
-                                 n,
-                                 &disjunction->conjunctions,
-                                 linkage) {
-            list_del_init(&conjunction->linkage);
-            dsync_conjunction_free(conjunction);
-        }
-        assert(list_empty(&disjunction->conjunctions));
-        mfu_free(&disjunction);
-    } else {
+    if (!disjunction) {
         return;
     }
+    assert(list_empty(&disjunction->linkage));
+    list_for_each_entry_safe(conjunction,
+                             n,
+                             &disjunction->conjunctions,
+                             linkage) {
+        list_del_init(&conjunction->linkage);
+        dsync_conjunction_free(conjunction);
+    }
+    assert(list_empty(&disjunction->conjunctions));
+    mfu_free(&disjunction);
 }
 
 static void dsync_disjunction_print(
@@ -3190,9 +3189,25 @@ int main(int argc, char **argv)
         usage = 1;
     }
     
-    if (usage) {
-        goto do_usage;
+    /* we should have two arguments left, source and dest paths */
+    int numargs = argc - optind;
+
+    /* if help flag was thrown, don't bother checking usage */
+    if (numargs != 2 && !help) {
+        MFU_LOG(MFU_LOG_ERR,
+            "You must specify a source and destination path.");
+        usage = 1;
     }
+
+    /* print usage and exit if necessary */
+    if (usage) {
+        if (rank == 0) {
+            print_usage();
+        }
+        rc = 1;
+        goto dsync_common_cleanup;
+    }
+    
     /* Generate default output */
     if (list_empty(&options.outputs) ) {
         /*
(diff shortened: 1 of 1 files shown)


src/dsync/dsync.c
```diff
@@ -2501,7 +2501,9 @@ static void dsync_output_init_disjunction(
 static void dsync_output_free(struct dsync_output* output)
 {
     assert(list_empty(&output->linkage));
-    dsync_disjunction_free(output->disjunction);
+    if (output->disjunction != NULL) {
```
ofaaland, line 2504: How about putting this check into `dsync_disjunction_free()` instead of doing it here?  That would make it the same as `free()`, where it's just a no-op if it's passed a NULL pointer.

src/dsync/dsync.c
```diff
@@ -3188,7 +3190,7 @@ int main(int argc, char **argv)
     }
 
     /* Generate default output */
-    if (list_empty(&options.outputs)) {
+    if (list_empty(&options.outputs) && ret == 0) {
```
ofaaland, line 3193: Can we just clean up and `exit()` up above, at
```
3147         case 'o':
3148             ret = dsync_option_output_parse(optarg, 0);
3149             if (ret) {
3150                 usage = 1;
3151             }
3152             break;
```

If we're going to show the usage, then apparently the user provided invalid options, and we shouldn't waste time doing anything else, IMHO.

Also, right now nothing else above uses the `ret` variable, but in the future someone might decide it's useful for processing another option and not realize it affects behavior down below.

ofaaland (changes requested):
Hi @carbonneau1 thanks for tracking this down and fixing it.  I have two code suggestions below.  Also, please change your commit message to conform to the same standard we use for ZFS and Lustre,

```
Very Short (72 chars or less) summary (including utility if only one changed)

A blank line after the summary before the longer description (critical)

More detailed explanatory text. Wrap it to 72 characters. Give the
motivation for the change (ie "fixes a seg fault triggered by a double-free
when dsync is run with options dsync -o fubar") and a summary
of the code change(s).
```

A longer description of some good things things to think about with
commit messages is at https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

src/dsync/dsync.c
carbonneau1: make sense. I'll make the change

src/dsync/dsync.c
carbonneau1: I think that as soon as a usage is set to 1 we should just break, cleanup and show usage. Are we on the same page?

ofaaland: Yes, I agree with you

src/dsync/dsync.c
```diff
@@ -2411,17 +2411,21 @@ static void dsync_disjunction_free(struct dsync_disjunction* disjunction)
...
-
-    assert(list_empty(&disjunction->linkage));
-    list_for_each_entry_safe(conjunction,
-                             n,
-                             &disjunction->conjunctions,
-                             linkage) {
-        list_del_init(&conjunction->linkage);
-        dsync_conjunction_free(conjunction);
+    
+    if (disjunction) {
```
ofaaland, line 2415: style: This could be simplified a bit with an early exit:

```suggestion
    if (!disjunction) {
        return;
        }
```

Then the code after that (`assert()`, `list_for_each_entry_safe()`, etc.) doesn't change at all.

src/dsync/dsync.c
```diff
@@ -3186,9 +3189,12 @@ int main(int argc, char **argv)
         }
         usage = 1;
     }
-
+    
+    if (usage) {
+        goto do_usage;
```
ofaaland, line 3194: How about if you just move the code from lines 3225-3229 (if rank==0 print_usage(), goto dsync_common_cleanup) here?  Then you don't need the label `do_usage` and don't need two separate `if (usage) {` blocks, I think.

src/dsync/dsync.c
carbonneau1: I did that first and took it out. I'll put it back

src/dsync/dsync.c
carbonneau1: there is still the part with 2 arguments left to check if the source and destination are there. Can't move it

carbonneau1: I moved a litle more so it yields the parser failures and the error cause we don't have source and destination argument if they're missing . that's even bettoer.

ofaaland (changes requested):
Looks great!

For the commit message(s), can you add "dsync: " to the beginning of the summary (first line) since these changes don't affect any other utility?

It just makes it easier, later, if someone is looking for a change that fixed or created a bug they care about.

Then I'll approve and merge.

ofaaland (approved):
Looks good, thanks!

carbonneau1:
yes 608 fixed by 612!

PR #613, opened 2024-12-05
dfilemaker: exit in a consistent manner

ofaaland:
In some cases MPI_Finalize() and similar teardown
functions were called before exit, and in other cases they were not.

In addition, in many cases the exit code was 0 even though we were exiting due to an error condition.

When an error is identified during option parsing or other tasks in main(), exit with a meaningful error code and use the same code for cleanup prior to exiting.

When an error is encountered elsewhere, exit with a useful error code.


carbonneau1 (approved):
Much cleaner.

levatin1:
I approve.

levatin1 (approved):

PR #617, opened 2024-12-13
test_dsync/test_existence.sh: use dfilemaker for test data

ofaaland:
Instead of using /usr/include/c++ as a source of test data, generate test data using dfilemaker.  There's no guarantee a machine being used for testing MFU will have /usr/include/c++ installed.

Remove TEST_TREE_DATA for test that does not use it.

Make first arg of the test_dsync tests, specify the MFU binary directory to use for testing, instead of specifying the path to dsync itself.

### DIFF
--- a/test/tests/test_dsync/test_data.sh
+++ b/test/tests/test_dsync/test_data.sh
@@ -16,12 +16,12 @@
 # Turn on verbose output
 #set -x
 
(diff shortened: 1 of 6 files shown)


daltonbohning (approved):
Nice :)

PR #619, opened 2025-01-03
test_data.sh: add test to verify file contents

ofaaland:
Implement a test to verify checksums of the destination match those of the source, after dsync runs.

The first test runs dsync without any options, copying to an empty destination directory. It then calculates checksums for the files in source and destination, and verifies they match.

The second test creates data in source and destination directories, including files whose metadata match but whose data differ.  It then runs dsync without any options to verify no copies occur.  The process is roughly this:

	- Create data in src
	- Copy it to dest
	- Change data in files in dest, without altering length
	- Set timestamps on all files so they are the same
	- Run dsync without --contents option

	For all files, metadata should match but data should not match,
	and no copies should occur.

The third test runs dsync with --contents and then verifies that the copies occurred and the data in destination and source match.

### DIFF
--- a/test/tests/test_dsync/test_data.sh
+++ b/test/tests/test_dsync/test_data.sh
@@ -17,15 +17,164 @@
 #set -x
 
 MFU_TEST_BIN=${MFU_TEST_BIN:-${1}}
-DSYNC_SRC_DIR=${DSYNC_SRC_DIR:-${2}}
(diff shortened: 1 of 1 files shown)


carbonneau1 (approved):

ofaaland:
No content difference between old commit, 2693b1d and new commit, ee00160.  The difference is that the latter commit is signed.

PR #622, opened 2025-01-17
Fix for issue dsync not updating times correctly #560.

carbonneau1:
Some file system do not support nanosecond part of the timespec for mtime (last modification time). We were comparing the mtime (tv_nsec nanosecond) regardless of the support to mtime nanosecond. This was causing dsync to add files to the copy to destination list even if the file were identical.

We added a check for source dir and destination dir to verify if the timepsec nanosecond (tv_nsec) was set to 0. In which case we do not compare the mtime nanosecond part of it (tv_nsec).

### DIFF
--- a/src/dsync/dsync.c
+++ b/src/dsync/dsync.c
@@ -48,6 +48,9 @@
 #include "mfu_daos.h"
 #endif
 
+/* True if both src and dest file systems support nsec field in mtime */
+static bool comp_mtime_nsec = true;
+
 /* Print a usage message */
 static void print_usage(void)
 {
@@ -1152,7 +1155,7 @@ static void dsync_strmap_compare_lite_link_dest(
 
         /* if size and mtime are the same, we assume the file contents are same */
         if ((src_size == dst_size) &&
-            (src_mtime == dst_mtime) && (src_mtime_nsec == dst_mtime_nsec))
+            (src_mtime == dst_mtime) && (comp_mtime_nsec ? (src_mtime_nsec == dst_mtime_nsec):true))
         {
             mfu_flist_file_copy(link_compare_list, idx, link_same_list);
         }
@@ -1200,12 +1203,12 @@ static int dsync_strmap_compare_lite(
 
         /* if size or mtime is different, we assume the file contents are different */
         if ((src_size != dst_size) ||
-            (src_mtime != dst_mtime) || (src_mtime_nsec != dst_mtime_nsec))
+            (src_mtime != dst_mtime) || (comp_mtime_nsec ? (src_mtime_nsec != dst_mtime_nsec):false))
         {
             /* update to say contents of the files were found to be different */
             dsync_strmap_item_update(src_map, name, DCMPF_CONTENT, DCMPS_DIFFER);
             dsync_strmap_item_update(dst_map, name, DCMPF_CONTENT, DCMPS_DIFFER);
-
+            
             /* mark file to be deleted from destination, copied from source */
             if (!options.dry_run || use_hardlinks) {
                 mfu_flist_file_copy(dst_compare_list, idx, dst_remove_list);
@@ -2968,6 +2971,20 @@ dsync_daos_out:
 }
 #endif
 
+static bool is_mtime_nsec_used(char *srcpath, char *dstpath)
+{
+    struct stat srcpath_stat, dstpath_stat;
+    if (stat(srcpath, &srcpath_stat) == 0 && stat(dstpath, &dstpath_stat) == 0) {
+        if (srcpath_stat.st_mtim.tv_nsec == 0 || dstpath_stat.st_mtim.tv_nsec == 0) {
+           return false; 
+        } else {
+            return true;
+        }
+    } else {
+        return false;
+    }
+}
+
 int main(int argc, char **argv)
 {
     int rc = 0;
@@ -3254,9 +3271,14 @@ int main(int argc, char **argv)
     mfu_param_path* srcpath  = &paths[0];
     mfu_param_path* destpath = &paths[1];
 
+    
     /* process each path */
     mfu_param_path_set((const char*)argpaths[0], srcpath,  mfu_src_file, true);
     mfu_param_path_set((const char*)argpaths[1], destpath, mfu_dst_file, false);
+    /* some filesystem don't support 
+     * nanosecond in mtime time stamp aavoid the comp if 
+     * it is the case */
+    comp_mtime_nsec = is_mtime_nsec_used(srcpath->path, destpath->path);
 
     /* advance to next set of options */
     optind += numargs;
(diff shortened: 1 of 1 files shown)


src/dsync/dsync.c
```diff
@@ -1197,14 +1206,20 @@ static int dsync_strmap_compare_lite(
         uint64_t src_mtime_nsec = mfu_flist_file_get_mtime_nsec(src_compare_list, idx);
         uint64_t dst_mtime      = mfu_flist_file_get_mtime(dst_compare_list, idx);
         uint64_t dst_mtime_nsec = mfu_flist_file_get_mtime_nsec(dst_compare_list, idx);
+        char srcmtimeString[100];
+        char dstmtimeString[100];
+        strftime(srcmtimeString, sizeof(srcmtimeString), "%c", localtime(&src_mtime));
```
ofaaland, line 1211: I think these two strftime() calls might affect performance since they are called unconditionally.  How about declaring and filling these string values only if you need them?

src/dsync/dsync.c
```diff
@@ -1197,14 +1206,20 @@ static int dsync_strmap_compare_lite(
...
         /* if size or mtime is different, we assume the file contents are different */
         if ((src_size != dst_size) ||
-            (src_mtime != dst_mtime) || (src_mtime_nsec != dst_mtime_nsec))
+            (src_mtime != dst_mtime) || (comp_mtime_nsec ? (src_mtime_nsec != dst_mtime_nsec):false))
         {
             /* update to say contents of the files were found to be different */
             dsync_strmap_item_update(src_map, name, DCMPF_CONTENT, DCMPS_DIFFER);
             dsync_strmap_item_update(dst_map, name, DCMPF_CONTENT, DCMPS_DIFFER);
+            
+            MFU_LOG(MFU_LOG_WARN, "file: %s src_mtime: %s dst_mtime: %s\n", name, srcmtimeString, dstmtimeString);
```
ofaaland, line 1222: "Warning" or MFU_LOG_WARN is for cases where something might be broken broken, for example data may not be copied correctly or files may be lost.    In this case, are you just telling the user why a file was copied?

If so, look for other uses of MFU_LOG();  MFU_LOG_DBG or MFU_LOG_INFO are probably more appropriate.

src/common/mfu_flist_copy.c
```diff
@@ -150,6 +150,33 @@ static void mkdir_progress_fn(const uint64_t* vals, int count, int complete, int
 static mfu_copy_file_cache_t mfu_copy_src_cache;
 static mfu_copy_file_cache_t mfu_copy_dst_cache;
 
+/* close a file that opened with mfu_copy_open_file */
+
+static int mfu_copy_close_file(
```
ofaaland, line 155: (style) Why move this?  It doesn't hurt anything, but it makes it much harder to see if there are changes and so harder to see if a bug was created accidentally.

src/common/mfu_flist_copy.c
```diff
@@ -170,8 +197,7 @@ static int mfu_copy_open_file(
         } else {
             /* the file we're trying to open is different,
              * close the old file and delete the name */
-            mfu_file_close(name, mfu_file);
-            mfu_free(&cache->name);
+            mfu_copy_close_file(cache, mfu_file);
```
ofaaland, line 200: Is this change part of fixing the nsec bug?

ofaaland (commented):
Looks good!  I think there's some unused code from earlier testing, and possibly fixes to other bugs (if so, they should go in separate commits).

src/dsync/dsync.c
```diff
@@ -48,6 +48,15 @@
 #include "mfu_daos.h"
 #endif
 
+/* for statfs call to check fs type */
+ #include <sys/vfs.h>
+
+/* the next macro is temporary until lustre can handle usec in timestamp */
+#define LUSTRE_SUPER_MAGIC  0xbd00bd0
```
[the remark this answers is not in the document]
carbonneau1: Not anymore I was detecting lustre with status for a while and no more.

src/dsync/dsync.c
```diff
@@ -48,6 +48,15 @@
 #include "mfu_daos.h"
 #endif
 
+/* for statfs call to check fs type */
```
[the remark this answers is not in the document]
carbonneau1: not anymore, gone.

src/dsync/dsync.c
carbonneau1: debug statement left out, taking it out. thanks for the findings.

src/dsync/dsync.c
carbonneau1: Debug statement left over. Taking it out.

src/common/mfu_flist_copy.c
carbonneau1: I'll move it back and add a prototype at the beginning of the file.

src/common/mfu_flist_copy.c
carbonneau1: no it was to make sure the sync is called with the close of file. It will be good for other FS where the sync is enabled on the backend.

carbonneau1: That change was from Adam but never merged.

src/common/mfu_flist_copy.c
carbonneau1: I'll add a new MR for that one as it is a different issue.

src/common/mfu_flist_copy.c
carbonneau1: I'll take it out and add a new MR as it is a different issue.

src/dsync/dsync.c
```diff
@@ -2968,6 +2971,20 @@ static int dsync_daos(
 }
 #endif
 
+static bool is_mtime_nsec_used(char *srcpath, char *dstpath)
```
ofaaland, line 2974: (nit) If you refresh the patch, it would be good to add a comment here that this is a heuristic - we're guessing, that if tv_nsec == 0, that the file system doesn't support nsec even though 0 is a valid value and sometimes our check might be wrong.

No need to re-do just for this though, since that's described in the issue.

src/dsync/dsync.c
```diff
@@ -2968,6 +2971,20 @@ static int dsync_daos(
...
+{
+    struct stat srcpath_stat, dstpath_stat;
+    if (stat(srcpath, &srcpath_stat) == 0 && stat(dstpath, &dstpath_stat) == 0) {
+        if (srcpath_stat.st_mtim.tv_nsec == 0 || dstpath_stat.st_mtim.tv_nsec == 0) {
+           return false; 
+        } else {
+            return true;
+        }
+    } else {
+        return false;
```
ofaaland, line 2984: If we can't stat() one of src and dest, then we don't know whether nsec are supported.  

Is the "safe" thing to return true, so that we compare nsec?  Then we might copy more files than we need to, but we can be sure the dest and the src match.

ofaaland (commented):
I'm not sure about the case where the stat() calls fail.  Looks good to me otherwise, thanks!

src/dsync/dsync.c
carbonneau1: ok, will return true. Make sense, I think!

PR #624, opened 2025-02-20
readthedocs configuration file now required

gonsie:
I will test that this actually builds successfully


ofaaland (approved):
Thanks!  LGTM

PR #632, opened 2025-03-06
daos: fix permissions with drm

daltonbohning:
It was assumed that "source" containers only needed read permission, but with drm the "source" needs write permission for deletion.


daltonbohning:
Doing more internal testing with this before review

src/common/mfu_daos.c
```diff
@@ -1362,6 +1361,10 @@ daos_args_t* daos_args_new(void)
     da->daos_preserve = false;
     da->daos_preserve_path = NULL;
 
+    /* in most cases source only needs RO and destination needs RW */
+    da->default_src_cont_open_flags = DAOS_COO_RO;
```
carbonneau1, line 1365: What was the side effect of not setting permissions on source and destination.
Does that mean the delete never worked in days.

carbonneau1 (approved):
Looks good. I was curious about the side effect of read only on source during the deletion.

src/common/mfu_daos.c
daltonbohning: Yep! We used to set RW on the source and destination, so remove worked. Then we trimmed just the source to be RO, which broke remove.

carbonneau1:
Hi Dalton,  have you signed your commit?

daltonbohning:
> Hi Dalton, have you signed your commit?

It should be, yes. Does it now show for you? (I'm new to signing this way)

carbonneau1:
I tried a rebase and merge, complained about commit had to be signed. I'm still adjusting to git ....

carbonneau1:
squash and merge did it.

daltonbohning:
> I tried a rebase and merge, complained about commit had to be signed. I'm still adjusting to git ....

Just guessing but I think a `Rebase and merge` would require YOU to sign the commit? Because that puts your name on the commit too.
But a `Squash and merge` would not because it doesn't include your name? Again, just guessing

daltonbohning:
Thanks @carbonneau1 !

daltonbohning:
Anyway, I think `Squash and merge` is probably want you want. That squashes all commits on the PR into one, then "merges" / pushes that commit to master. `Rebase and merge` applies each commit individually to master

PR #629, opened 2025-03-06
dtar: fix creation with symlink and --preserve-*

rezib:
Fix creation of corrupted archive with symlink when --preserve-{xattrs,acl,flags} command line option is used.

When it is provided a valid file descriptor, libarchive archive_read_disk_entry_from_file() calls fstat() which follows symlinks. This commit introduces an exception for symlinks to not provide a file descriptor to archive_read_disk_entry_from_file() which makes it call lstat() on entry path and create a valid symlink archive entry eventually.

fix #628

Issue #628, opened 2025-03-06
dtar produces corrupted archive with symlinks and `--preserve-{xattrs,acl,flags}`


rezib:
For example:

```console
$ ls -lh test/
total 4.0K
-rw-rw-r-- 1 remi remi 4 Mar  6 14:07 file1
lrwxrwxrwx 1 remi remi 5 Mar  6 14:07 symlink1 -> file1
$ mpirun --mca mpi_abort_print_stack 1 -N 4 /home/remi/Documents/git/cea/env/bin/dtar --preserve-xattrs -cf archive.tar test
[2025-03-06T14:09:23] Walking /path/to/test
[2025-03-06T14:09:23] Walked 3 items in 0.003 secs (935.879 items/sec) ...
[2025-03-06T14:09:23] Walked 3 items in 0.003 seconds (888.621 items/sec)
[2025-03-06T14:09:23] Writing archive to archive.tar
[2025-03-06T14:09:23] Items: 3
[2025-03-06T14:09:23]   Directories: 1
[2025-03-06T14:09:23]   Files: 1
[2025-03-06T14:09:23]   Links: 1
[2025-03-06T14:09:23] Data: 4.000 B (4.000 B per file)
[2025-03-06T14:09:23] Sorted 3 items in 0.000 seconds (9201.325 items/sec)
[2025-03-06T14:09:23] Truncating archive
[2025-03-06T14:09:23] Gathering index
[2025-03-06T14:09:23] Writing index as last entry in archive.tar
[2025-03-06T14:09:23] Writing entry headers
[2025-03-06T14:09:23] Copying file data
[2025-03-06T14:09:23] Started:   Mar-06-2025, 14:09:23
[2025-03-06T14:09:23] Completed: Mar-06-2025, 14:09:23
[2025-03-06T14:09:23] Seconds: 0.003
[2025-03-06T14:09:23] Archive size: 8.500 KiB
[2025-03-06T14:09:23] Rate: 3.040 MiB/s (8704 bytes in 0.003 seconds)
```

The generated archive is corrupted:

```console
$ tar -tf archive.tar 
test/
test/file1
test/symlink1
tar: Skipping to next header
archive.tar.dtaridx
tar: Exiting with failure status due to previous errors

```

carbonneau1:
Hi Remi, sorry for the late reply. I will be looking into that one shortly.

carbonneau1:
[...]

PR #625, opened 2025-02-25
dsync: fix -L,--dereference option ignored

rezib:
This commit fixes dsync ignoring -L,--dereference option because of a missing break in the case statement, causing walk/copy options to be overwritten by -P,--no-dereference values.

Issue #336, opened 2020-05-20
How to detect hardlinks?


timeu:
We would like to use dfind in combination with dwalk to find hardlinks.
The dfind command line tool does not provide this option as of now. 
I am wondering how difficult it would be to add this feature? 
As far as I can tell, dwalk uses stat to retrieve information about a file/directory which is used for some filters (gid, atime, etc). 
Is the entire stat struct stored in the binary output of dwalk ? (the text output does not contain it)

adammoody:
Unfortunately, we don't record the hard link information from stat in the binary file from dwalk either.  In fact, we don't even have an interface for hard links in our mfu_flist, which is a fundamental data structure that all of the tools are built around.  That means adding full and proper support would take time and require a new binary file format.

Can you describe in more detail the operation you'd like to do?

timeu:
Thanks for the quick reply. 
We are dealing with a storage migration where the source filesystem might have symlinks and hardlinks. Dsync will properly snyc the symlinks however the hardlinks are dereferenced on the target filesystem. 
Coreutils `cp` has the --preserve=link or -a flag to deal with this. 
My assumption was that even if mpifileutils cannot deal with hardlinks, we could at least filter out the hardlinks from the source and then copy them over using the regular coreutil `cp` tool or `rsync`.

adammoody:
Thinking about this more.  It would take a lot of work (time) to add full support for hard links into the library, but one might be able to create a one-off filtering tool relatively quickly.  It would require stat'ing each file twice.  One could do a full mfu_flist_walk to get an initial list, then each process could stat each of its items within its local list.  The process checks the hard link count of each item, and either copies the item into an output list or drops the item.  Finally, the resulting filtered list is written out.

timeu:
@adammoody 
That would be actually a good workaround. So basically one would write a new CLI tool that takes the output of dwalk and runs stat every file again and then outputs the hardlinks or the list without hardlinks.
I haven't done any C code for a long long time but I will have a look. Probably makes sense to take the code from dfind as a base and adapt it ?

adammoody:
I can likely hack up something to help get you started.  Will do that in a bit.

adammoody:
Here is a branch with a one-off tool:
https://github.com/hpc/mpifileutils/tree/hardlink

You can grab the source here and see the buildme script for the compile line:
https://github.com/hpc/mpifileutils/tree/hardlink/src/dhardlink

[...]

PR #633, opened 2025-03-21
Hardlinks support

### LINKED ISSUE #417: dsync and probably dcp doesn't handle hardlinks
Hardlinks are not handled and causes errors on the destination files:
source:
/a/b hardlinked to /a/c
dsync /a /target

[2020-10-31T05:44:52] [39] [/path/src/common/mfu_flist_copy.c:330] ERROR: Failed to change ownership on `/target/c' lchown() (errno=2 No such file or directory)
[2020-10-31T05:44:52] [39] [/path/src/common/mfu_flist_copy.c:356] ERROR: Failed to change permissions on `/target/c' chmod() (errno=2 No such file or directory)
And one more for utime

And /target/b and /target/c ends up as separate files, sometimes with one of them (c in the above case) having incorrect timestamps.

Currently running mpifileutils 0.10.1

rezib:
Dear mpifileutils developers,

This is my proposal to add support of hardlinks in many mpifileutils commands: `dwalk`, `dcp`, `dcmp`, `dsync` and `dtar`. 

During tree walk with details, regular files with more than one _nlink_ are temporarily placed in a hardlinks flist. This flist is then globally ordered by names and ranked to select one reference path per inode, and flag all other paths to this inodes as hardlinks. The sorted hardlinks flist is finally merged in global flist with all other items. The paths name ordering is performed to ensure reproducibility between two similar trees, thus minimizing the differences for `dcmp` and `dsync` eventually.

> [!NOTE]
> You may find more implementations details in respective commits messages.

The pull request introduces a cache format v5, to support encoding of files nlink and hardlinks references paths.

This pull request also includes a functional test suite that relies on Python standard `unittest` library. This suite is designed to be easy to execute:

* Set two environment variables to define respectively the path to mpifileutils binaries and arguments provided to mpirun, eg:

```console
$ export MFU_BIN=~/dev/bin
$ export MFU_MPIRUN_ARGS="--bind-to none --oversubscribe -N 4"
```

* And run all the tests:

```console
$ python3 -m unittest discover -v test
```

* Or:

```console
$ pytest  # require pytest
```

It is also designed to be easy to integrate in continuous integration systems. The pull request even provides a GitHub action workflow to execute this test suite on every pull requests and merges in _main_ branch ([example run](https://github.com/rezib/mpifileutils/actions/runs/13989540919/job/39170212335)).

For the record, this test suite has already helped detect and fix the following bugs:

* ~https://github.com/hpc/mpifileutils/pull/625~ (merged)
* ~https://github.com/hpc/mpifileutils/issues/628~ → ~https://github.com/hpc/mpifileutils/pull/629~ (merged)

Please let me know what you think! I can also remove the tests and GitHub actions workflow if you don't like the technical approach.

> [!IMPORTANT]
> Note this feature does not work properly without this fix for a bug in DTCMP: https://github.com/LLNL/dtcmp/pull/20

> [!IMPORTANT]
> There is one limitation with `dcp`/`dsync --dereference` when symlinks point to path with more than one link. In this specific case, mpifileutils will consider the symlink as one more additional path to the same inode and create one more hardlink on this inodes in destination directory. For reference, this case is coverered by test `test_dsync_symlink_dereference_target_nlinks`.

> [!NOTE]
> I would like to emphasize that this work is sponsored by @cea-hpc.

fix #417 #336

### DIFF
--- a/src/common/mfu_flist.c
+++ b/src/common/mfu_flist.c
@@ -125,7 +125,7 @@ static size_t list_elem_pack2_size(int detail, uint64_t chars, const elem_t* ele
 {
     size_t size;
     if (detail) {
-        size = 2 * 4 + chars + 0 * 4 + 10 * 8;
+        size = 2 * 4 + chars + 1 * 4 + 10 * 8 + 8 + chars;
     }
     else {
         size = 2 * 4 + chars + 1 * 4;
@@ -154,11 +154,7 @@ static size_t list_elem_pack2(void* buf, int detail, uint64_t chars, const elem_
     mfu_pack_uint32(&ptr, (uint32_t) chars);
 
     /* copy in file name */
-    char* file = elem->file;
-    if (file != NULL) {
-        strcpy(ptr, file);
-    }
-    ptr += chars;
+    mfu_pack_sized_str(&ptr, elem->file, chars);
 
 #ifdef DAOS_SUPPORT
     /* copy in values for obj ids */
@@ -166,6 +162,7 @@ static size_t list_elem_pack2(void* buf, int detail, uint64_t chars, const elem_
     mfu_pack_uint64(&ptr, elem->obj_id_hi);
 #endif
 
+    mfu_pack_uint32(&ptr, elem->type);
     if (detail) {
         /* copy in fields */
         mfu_pack_uint64(&ptr, elem->mode);
@@ -178,10 +175,8 @@ static size_t list_elem_pack2(void* buf, int detail, uint64_t chars, const elem_
         mfu_pack_uint64(&ptr, elem->ctime);
         mfu_pack_uint64(&ptr, elem->ctime_nsec);
         mfu_pack_uint64(&ptr, elem->size);
-    }
-    else {
-        /* just have the file type */
-        mfu_pack_uint32(&ptr, elem->type);
+        mfu_pack_uint64(&ptr, elem->nlink);
+        mfu_pack_sized_str(&ptr, elem->ref, chars);
     }
 
     size_t bytes = (size_t)(ptr - start);
@@ -203,15 +198,11 @@ static size_t list_elem_unpack2(const void* buf, elem_t* elem)
     uint32_t chars;
     mfu_unpack_uint32(&ptr, &chars);
 
-    /* get name and advance pointer */
-    const char* file = ptr;
-    ptr += chars;
-
-    /* copy path */
-    elem->file = MFU_STRDUP(file);
+    /* get name */
+    mfu_unpack_sized_str(&ptr, &elem->file, chars);
 
     /* set depth */
-    elem->depth = mfu_flist_compute_depth(file);
+    elem->depth = mfu_flist_compute_depth(elem->file);
 
     elem->detail = (int) detail;
 
@@ -220,6 +211,10 @@ static size_t list_elem_unpack2(const void* buf, elem_t* elem)
     mfu_unpack_uint64(&ptr, &elem->obj_id_lo);
     mfu_unpack_uint64(&ptr, &elem->obj_id_hi);
 #endif
+
+    uint32_t type;
+    mfu_unpack_uint32(&ptr, &type);
+    elem->type = (mfu_filetype) type;
 
     if (detail) {
         /* extract fields */
@@ -233,14 +228,8 @@ static size_t list_elem_unpack2(const void* buf, elem_t* elem)
         mfu_unpack_uint64(&ptr, &elem->ctime);
         mfu_unpack_uint64(&ptr, &elem->ctime_nsec);
         mfu_unpack_uint64(&ptr, &elem->size);
-        /* use mode to set file type */
-        elem->type = mfu_flist_mode_to_filetype((mode_t)elem->mode);
-    }
-    else {
-        /* only have type */
-        uint32_t type;
-        mfu_unpack_uint32(&ptr, &type);
-        elem->type = (mfu_filetype) type;
+        mfu_unpack_uint64(&ptr, &elem->nlink);
+        mfu_unpack_sized_str(&ptr, &elem->ref, chars);
     }
 
     size_t bytes = (size_t)(ptr - start);
@@ -346,6 +335,8 @@ static void list_insert_copy(flist_t* flist, elem_t* src)
     elem->ctime      = src->ctime;
     elem->ctime_nsec = src->ctime_nsec;
     elem->size       = src->size;
+    elem->nlink      = src->nlink;
+    elem->ref        = MFU_STRDUP(src->ref);
 
     /* append element to tail of linked list */
     mfu_flist_insert_elem(flist, elem);
@@ -368,6 +359,9 @@ void mfu_flist_insert_stat(flist_t* flist, const char* fpath, mode_t mode, const
     /* set file type */
     elem->type = mfu_flist_mode_to_filetype(mode);
 
+    /* hardlinks references are discovered afterwhile */
+    elem->ref = NULL;
+
     /* copy stat info */
     if (sb != NULL) {
         elem->detail = 1;
@@ -389,6 +383,7 @@ void mfu_flist_insert_stat(flist_t* flist, const char* fpath, mode_t mode, const
         elem->ctime_nsec = nsecs;
 
         elem->size  = (uint64_t) sb->st_size;
+        elem->nlink = (uint64_t) sb->st_nlink;
 
         /* TODO: link to user and group names? */
     }
@@ -409,6 +404,7 @@ static void list_delete(flist_t* flist)
     while (current != NULL) {
         elem_t* next = current->next;
         mfu_free(&current->file);
+        mfu_free(&current->ref);
         mfu_free(&current);
         current = next;
     }
@@ -987,6 +983,28 @@ uint64_t mfu_flist_file_get_size(mfu_flist bflist, uint64_t idx)
     return ret;
 }
 
+uint64_t mfu_flist_file_get_nlink(mfu_flist bflist, uint64_t idx)
+{
+    uint64_t ret = (uint64_t) - 1;
+    flist_t* flist = (flist_t*) bflist;
+    elem_t* elem = list_get_elem(flist, idx);
+    if (elem != NULL && flist->detail) {
+        ret = elem->nlink;
+    }
+    return ret;
+}
+
+const char* mfu_flist_file_get_ref(mfu_flist bflist, uint64_t idx)
+{
+    const char* ref = NULL;
+    flist_t* flist = (flist_t*) bflist;
+    elem_t* elem = list_get_elem(flist, idx);
+    if (elem != NULL) {
+        ref = elem->ref;
+    }
+    return ref;
+}
+
 const char* mfu_flist_file_get_username(mfu_flist bflist, uint64_t idx)
 {
     const char* ret = NULL;
@@ -1167,6 +1185,19 @@ void mfu_flist_file_set_size(mfu_flist bflist, uint64_t idx, uint64_t size)
     elem_t* elem = list_get_elem(flist, idx);
     if (elem != NULL) {
         elem->size = size;
+    }
+    return;
+}
+
+void mfu_flist_file_set_ref(mfu_flist bflist, uint64_t idx, const char* ref)
+{
+    flist_t* flist = (flist_t*) bflist;
+    elem_t* elem = list_get_elem(flist, idx);
+    if (elem != NULL) {
+        /* free existing name if there is one */
+        mfu_free(&elem->ref);
+        /* set new ref*/
+        elem->ref = MFU_STRDUP(ref);
     }
     return;
 }
@@ -1353,6 +1384,8 @@ uint64_t mfu_flist_file_create(mfu_flist bflist)
     elem->ctime      = 0;
     elem->ctime_nsec = 0;
     elem->size       = 0;
+    elem->nlink      = 0;
+    elem->ref        = NULL;
 
     /* for DAOS */
 #ifdef DAOS_SUPPORT
@@ -1817,11 +1850,12 @@ void mfu_flist_print_summary(mfu_flist flist)
     MPI_Comm_size(MPI_COMM_WORLD, &ranks);
 
     /* initlialize counters */
-    uint64_t total_dirs    = 0;
-    uint64_t total_files   = 0;
-    uint64_t total_links   = 0;
-    uint64_t total_unknown = 0;
-    uint64_t total_bytes   = 0;
+    uint64_t total_dirs      = 0;
+    uint64_t total_files     = 0;
+    uint64_t total_links     = 0;
+    uint64_t total_hardlinks = 0;
+    uint64_t total_unknown   = 0;
+    uint64_t total_bytes     = 0;
 
     /* step through and print data */
     uint64_t idx = 0;
@@ -1839,8 +1873,12 @@ void mfu_flist_print_summary(mfu_flist flist)
                 total_dirs++;
             }
             else if (S_ISREG(mode)) {
-                total_files++;
-                total_bytes += size;
+                if (mfu_flist_file_get_ref(flist, idx) != NULL) {
+                    total_hardlinks++;
+                } else {
+                    total_files++;
+                    total_bytes += size;
+                }
             }
             else if (S_ISLNK(mode)) {
                 total_links++;
@@ -1863,6 +1901,9 @@ void mfu_flist_print_summary(mfu_flist flist)
             else if (type == MFU_TYPE_LINK) {
                 total_links++;
             }
+            else if (type == MFU_TYPE_HARDLINK) {
+                total_hardlinks++;
+            }
             else {
                 /* unknown file type */
                 total_unknown++;
@@ -1874,13 +1915,14 @@ void mfu_flist_print_summary(mfu_flist flist)
     }
 
     /* get total directories, files, links, and bytes */
-    uint64_t all_dirs, all_files, all_links, all_unknown, all_bytes;
+    uint64_t all_dirs, all_files, all_links, all_hardlinks, all_unknown, all_bytes;
+    MPI_Allreduce(&total_dirs,      &all_dirs,      1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(&total_files,     &all_files,     1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(&total_links,     &all_links,     1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(&total_hardlinks, &all_hardlinks, 1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(&total_unknown,   &all_unknown,   1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(&total_bytes,     &all_bytes,     1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
     uint64_t all_count = mfu_flist_global_size(flist);
-    MPI_Allreduce(&total_dirs,    &all_dirs,    1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
-    MPI_Allreduce(&total_files,   &all_files,   1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
-    MPI_Allreduce(&total_links,   &all_links,   1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
-    MPI_Allreduce(&total_unknown, &all_unknown, 1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
-    MPI_Allreduce(&total_bytes,   &all_bytes,   1, MPI_UINT64_T, MPI_SUM, MPI_COMM_WORLD);
 
     /* convert total size to units */
     if (rank == 0) {
@@ -1888,6 +1930,7 @@ void mfu_flist_print_summary(mfu_flist flist)
         MFU_LOG(MFU_LOG_INFO, "  Directories: %llu", (unsigned long long) all_dirs);
         MFU_LOG(MFU_LOG_INFO, "  Files: %llu", (unsigned long long) all_files);
         MFU_LOG(MFU_LOG_INFO, "  Links: %llu", (unsigned long long) all_links);
+        MFU_LOG(MFU_LOG_INFO, "  Hardlinks: %llu", (unsigned long long) all_hardlinks);
         /* MFU_LOG(MFU_LOG_INFO, "  Unknown: %lu", (unsigned long long) all_unknown); */
 
         if (mfu_flist_have_detail(flist)) {
--- a/src/common/mfu_flist.c
+++ b/src/common/mfu_flist.c
@@ -404,7 +404,8 @@ static void list_delete(flist_t* flist)
     while (current != NULL) {
         elem_t* next = current->next;
         mfu_free(&current->file);
-        mfu_free(&current->ref);
+        if (current->detail)
+            mfu_free(&current->ref);
         mfu_free(&current);
         current = next;
     }
--- a/src/common/mfu_flist.h
+++ b/src/common/mfu_flist.h
@@ -94,11 +94,12 @@ extern "C" {
  * so changing their values can break backwards compatibility
  * in reading any such files */
 typedef enum mfu_filetypes_e {
-    MFU_TYPE_NULL    = 0, /* type not set */
-    MFU_TYPE_UNKNOWN = 1, /* type not known */
-    MFU_TYPE_FILE    = 2, /* regular file */
-    MFU_TYPE_DIR     = 3, /* directory */
-    MFU_TYPE_LINK    = 4, /* symlink */
+    MFU_TYPE_NULL     = 0, /* type not set */
+    MFU_TYPE_UNKNOWN  = 1, /* type not known */
+    MFU_TYPE_FILE     = 2, /* regular file */
+    MFU_TYPE_DIR      = 3, /* directory */
+    MFU_TYPE_LINK     = 4, /* symlink */
+    MFU_TYPE_HARDLINK = 5, /* hardlink */
 } mfu_filetype;
 
 /* define handle type to a file list */
@@ -395,6 +396,8 @@ uint64_t mfu_flist_file_get_mtime_nsec(mfu_flist flist, uint64_t index);
 uint64_t mfu_flist_file_get_ctime(mfu_flist flist, uint64_t index);
 uint64_t mfu_flist_file_get_ctime_nsec(mfu_flist flist, uint64_t index);
 uint64_t mfu_flist_file_get_size(mfu_flist flist, uint64_t index);
+uint64_t mfu_flist_file_get_nlink(mfu_flist flist, uint64_t idx);
+const char* mfu_flist_file_get_ref(mfu_flist flist, uint64_t idx);
 uint64_t mfu_flist_file_get_perm(mfu_flist flist, uint64_t index);
 #if DCOPY_USE_XATTRS
 void *mfu_flist_file_get_acl(mfu_flist bflist, uint64_t idx, ssize_t *acl_size, char *type);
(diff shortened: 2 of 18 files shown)


carbonneau1:
Going through your changes.

rezib:
> Important
> 
> There is one limitation with `dcp`/`dsync --dereference` when symlinks point to path with more than one link. In this specific case, mpifileutils will consider the symlink as one more additional path to the same inode and create one more hardlink on this inodes in destination directory. For reference, this case is coverered by test `test_dsync_symlink_dereference_target_nlinks`.

FWIW, I checked the behavior of GNU cp and rsync with similar options and it produces the same results.

For reference, here is my `setup-test.sh` script for this test:

```sh
#!/bin/sh

# Purge previous testing data
if [ -d orig ]; then
  rm orig/*
  rmdir orig
fi

if [ -d dest ]; then
  rm dest/*
  rmdir dest
fi

# Create fresh testing data
mkdir orig
echo foo > orig/foo
ln orig/foo orig/bar
ln -s foo orig/baz

#cp -R --dereference --preserve=all orig dest
#rsync -a --copy-links --hard-links orig/ dest
```

* With GNU `cp`:

```console
$ sh setup-test.sh
$ cp -R --dereference --preserve=all orig dest && ls -lirn orig dest
orig:
total 8
28201162 -rw-rw-r-- 2 1000 1000 4 Jun 13 16:53 foo
28201168 lrwxrwxrwx 1 1000 1000 3 Jun 13 16:53 baz -> foo
28201162 -rw-rw-r-- 2 1000 1000 4 Jun 13 16:53 bar

dest:
total 12
28201217 -rw-rw-r-- 3 1000 1000 4 Jun 13 16:53 foo
28201217 -rw-rw-r-- 3 1000 1000 4 Jun 13 16:53 baz
28201217 -rw-rw-r-- 3 1000 1000 4 Jun 13 16:53 bar
```

* With `rsync`:

```console
$ sh setup-test.sh
$ rsync -a --copy-links --hard-links orig/ dest && ls -lirn orig dest
orig:
total 8
28201162 -rw-rw-r-- 2 1000 1000 4 Jun 13 16:54 foo
28201168 lrwxrwxrwx 1 1000 1000 3 Jun 13 16:54 baz -> foo
28201162 -rw-rw-r-- 2 1000 1000 4 Jun 13 16:54 bar

dest:
total 12
28201217 -rw-rw-r-- 3 1000 1000 4 Jun 13 16:54 foo
28201217 -rw-rw-r-- 3 1000 1000 4 Jun 13 16:54 baz
28201217 -rw-rw-r-- 3 1000 1000 4 Jun 13 16:54 bar
```

In all cases, the dereferenced symlink `baz` is transformed into an additional hardlink on the inode pointed by `foo` and `bar`.

src/dsync/dsync.c
```diff
@@ -1581,6 +1581,166 @@ static int dsync_strmap_compare_link_dest(
...
+    }
+
+    MPI_Allgatherv(sendbuf, local_removed_refs * chars, MPI_CHAR,
+        recvbuf, recvcounts, recvdispls, MPI_CHAR, MPI_COMM_WORLD);
+
+    /* iterate of all reference names received */
+    uint64_t count = mfu_flist_size(dst_list);
+    char* recvptr = (char*) recvbuf;
+    for (int i = 0; i < (int) ranks; i++) {
+        for (int j = 0; j<recvcounts[i]; j++) {
```
ajmes: This line causes a segfault, when all the refs of a file are removed from the FS source.
```
(gdb) bt
#0  00007f77898a7332 in __strcmp_avx2 () from /lib64/libc.so.6
#1  000055dca3bc0109 in dsync_remove_hardlinks_with_removed_refs (mfu_src_file=..., mfu_dst_file=..., dst_map=..., dst_remove_list=...,
    dst_list=..., src_map=..., src_cp_list=..., src_list=...) at /usr/src/debug/mpifileutils-0.12-1.el8.x86_64/src/dsync/dsync.c:1697
#2  dsync_strmap_compare (src_list=..., src_map=..., dst_list=..., dst_map=..., link_list=..., link_map=..., strlen_prefix=...,
    copy_opts=..., src_path=..., dest_path=..., link_path=..., mfu_src_file=..., mfu_dst_file=...)
    at /usr/src/debug/mpifileutils-0.12-1.el8.x86_64/src/dsync/dsync.c:2057
#3  000055dca3bbc950 in main (argc=..., argv=...) at /usr/src/debug/mpifileutils-0.12-1.el8.x86_64/src/dsync/dsync.c:3649

(gdb) frame 1
#1  000055dca3bc0109 in dsync_remove_hardlinks_with_removed_refs (mfu_src_file=..., mfu_dst_file=..., dst_map=..., dst_remove_list=...,
    dst_list=..., src_map=..., src_cp_list=..., src_list=...) at /usr/src/debug/mpifileutils-0.12-1.el8.x86_64/src/dsync/dsync.c:1697
1697                    if(strcmp(ref, removed_ref))
(gdb) l
1692                    if(type != MFU_TYPE_HARDLINK)
1693                        continue;
1694
1695                    /* skip item if reference does not match */
1696                    const char *ref = mfu_flist_file_get_ref(dst_list, dst_index);
1697                    if(strcmp(ref, removed_ref))
1698                        continue;
1699
1700                    /* skip item if type already differs */
1701                    dsync_state state;
```
Here we iterate on the byte array instead of the reference array.

This should look like:
```
    for (int i = 0; i < (int) ranks; i++) {
        size_t ref_size = char_extent * chars;
        char *rank_next = recvptr + recvcounts[i];

        for (;recvptr < rank_next; recvptr+=ref_size) {
...
        }
        recvptr = rank_next;
```

Maybe we should add a regression test for this case.

src/dsync/dsync.c
```diff
@@ -1581,6 +1581,166 @@ static int dsync_strmap_compare_link_dest(
...
+                const char* key = strmap_node_key(node);
+
+                /* get index of destination file */
+                uint64_t dst_index;
+                int tmp_rc;
+                tmp_rc = dsync_strmap_item_index(dst_map, key, &dst_index);
+                if (tmp_rc) {
+                    rc = -1;
+                    MFU_LOG(MFU_LOG_ERR, "ERROR: Unable to find index of key `%s' in source map", key);
+                    continue;
```
ajmes, line 1692: This does not increment recvptr.

src/dsync/dsync.c
```diff
@@ -1581,6 +1581,166 @@ static int dsync_strmap_compare_link_dest(
...
+                }
+
+                /* skip item if not hardlink */
+                mfu_filetype type = mfu_flist_file_get_type(dst_list, dst_index);
+                if(type != MFU_TYPE_HARDLINK)
+                    continue;
```
ajmes, line 1697: same

src/dsync/dsync.c
```diff
@@ -1581,6 +1581,166 @@ static int dsync_strmap_compare_link_dest(
...
+
+                /* skip item if reference does not match */
+                const char *ref = mfu_flist_file_get_ref(dst_list, dst_index);
+                if(strcmp(ref, removed_ref))
+                    continue;
```
ajmes, line 1702: same

src/dsync/dsync.c
```diff
@@ -1581,6 +1581,166 @@ static int dsync_strmap_compare_link_dest(
...
+
+                /* skip item if type already differs */
+                dsync_state state;
+                tmp_rc = dsync_strmap_item_state(dst_map, key, DCMPF_TYPE, &state);
+                assert(tmp_rc == 0);
+                if (state == DCMPS_DIFFER)
+                    continue;
```
ajmes, line 1710: same

ajmes (commented):
The CEA uses this patch to migrate data between 2 Lustre FSs. The patch seems to work fine for the initial sync, but it can crash when updating hardlinks.

src/common/mfu_flist.c
```diff
@@ -409,6 +404,7 @@ static void list_delete(flist_t* flist)
     while (current != NULL) {
         elem_t* next = current->next;
         mfu_free(&current->file);
+        mfu_free(&current->ref);
```
ajmes: This segfault here during the tests with v5 cache format. If detail=0, ref pointer is never init (dwalk --input, dcp --input):
```
(gdb) bt
#0  0x00007fc31b08e675 in __GI___libc_free (mem=0x70616d6d206d6574) at malloc.c:3375
#1  0x00007fc31b88d374 in mfu_free (p=0xab73860) at /home/eaujames/mpifileutils/src/common/mfu_util.c:205
#2  0x00007fc31b85f4d1 in list_delete (flist=0xac5bce0) at /home/eaujames/mpifileutils/src/common/mfu_flist.c:410
#3  0x00007fc31b85f926 in mfu_flist_free (pbflist=0x7ffca0c8c1c0)
    at /home/eaujames/mpifileutils/src/common/mfu_flist.c:568
#4  0x0000000000402b0d in main (argc=3, argv=0x7ffca0c8c7a8) at /home/eaujames/mpifileutils/src/dwalk/dwalk.c:712
#2  0x00007fc31b85f4d1 in list_delete (flist=0xac5bce0) at /home/eaujames/mpifileutils/src/common/mfu_flist.c:410
410             mfu_free(&current->ref);
$13 = {file = 0x0, depth = 2, type = MFU_TYPE_DIR, detail = 0, mode = 8223706451981198701, 
  uid = 8386109825618634597, gid = 7379464839710535529, atime = 8390891584458421024, 
  atime_nsec = 7595413268515939429, mtime = 8247607283625304179, mtime_nsec = 8223706426295346546, 
  ctime = 2334102048954741093, ctime_nsec = 8316291872320417652, size = 2330684568022181236, 
  nlink = 8320808369521453414, 
  ref = 0x70616d6d206d6574 <error: Cannot access memory at address 0x70616d6d206d6574>, 
  next = 0xab720f0, obj_id_lo = 7378413688017811826, obj_id_hi = 2967907589870613857}
```
test:
```
test_dwalk.py:36: in run_dwalk                     
    return self.run_cmd(cmd)                       
...
E           AssertionError: Command error:                                                                                 
E            - command: mpirun --bind-to none --oversubscribe -N 8 /usr/local/bin/dwalk --input /tmp/tmpobbg88zs           
E            - exit code: 139                                                                                              
E            - stdout:                                                                                                     
E           [2025-09-11T11:42:18] Reading from input file: /tmp/tmpobbg88zs                                                
E           [2025-09-11T11:42:18] Read 9 files in 0.005 seconds (1762.452 files/sec)                                       
E           [2025-09-11T11:42:18] Items: 9                                                                                 
E           [2025-09-11T11:42:18]   Directories: 2                                                                         
E           [2025-09-11T11:42:18]   Files: 5                                                                               
E           [2025-09-11T11:42:18]   Links: 2                                                                               
E           [2025-09-11T11:42:18]   Hardlinks: 0                                                                           
E                                                                                                                          
E            - stderr:                                                                                                     
E           [eaujamesFR0130:322806:0:322806] Caught signal 11 (Segmentation fault: Sent by the kernel at address (nil))    
E           ==== backtrace (tid: 322806) ====                                                                              
E            0  /lib64/libucs.so.0(ucs_handle_error+0x2e4) [0x7fc31a7b5a74]                                                
E            1  /lib64/libucs.so.0(+0x1876d) [0x7fc31a7b776d]                                                              
E            2  /lib64/libucs.so.0(+0x1893d) [0x7fc31a7b793d]                                                              
E            3  /lib64/libc.so.6(+0x1a490) [0x7fc31b026490]                                                                
E            4  /lib64/libc.so.6(__libc_free+0x25) [0x7fc31b08e675]                                                        
E            5  /usr/local/lib64/libmfu.so.4.0.0(mfu_free+0x31) [0x7fc31b88d374]                                           
E            6  /usr/local/lib64/libmfu.so.4.0.0(+0xb4d1) [0x7fc31b85f4d1]                                                 
E            7  /usr/local/lib64/libmfu.so.4.0.0(mfu_flist_free+0x23) [0x7fc31b85f926]                                     
E            8  /usr/local/bin/dwalk(main+0xb98) [0x402b0d]                                                                
E            9  /lib64/libc.so.6(+0x3488) [0x7fc31b00f488]                                                                 
E           10  /lib64/libc.so.6(__libc_start_main+0x8b) [0x7fc31b00f54b]                                                  
E           11  /usr/local/bin/dwalk(_start+0x25) [0x401385]                                                               
E           =================================                                                                              
E           --------------------------------------------------------------------------                                     
E           prterun noticed that process rank 0 with PID 322806 on node eaujamesFR0130 exited on                           
E           signal 11 (Segmentation fault).                                                                                
E           --------------------------------------------------------------------------                                     
                                                                                                                           
lib.py:408: AssertionError                                                                                                 
```
With detail=0, ref is not set to NULL in list_elem_unpack(), so there is garbage in current->ref and mfu_free() try to free a random address.

src/dsync/dsync.c
rezib: Indeed, the pointer arithmetic over the the MPI_allgatherv() was wrong with chars. Fixed in [79b2219](https://github.com/hpc/mpifileutils/pull/633/commits/79b2219f72163d107c0577f1990d3275a28bc490).

src/dsync/dsync.c
rezib: Here continue jumps to next iteration of strmap_foreach, not the parent for loop, then there is no reason to increment recvptr here.

rezib:
Hello @ajmes, thanks for the review and the report!

The bug was actually covered by the tests but for reason I can't explain I have never seen it fail. That's why real world test is always better that small CI test cases :)

I added some test cases to increase probability of failure. I also fixed another bug discovered while adding these test cases where modification on source hardlink reference could lead to double copy (and error eventually) of hardlinks in destination.

src/dsync/dsync.c
ajmes: Yes, sorry, my bad.

src/common/mfu_flist.h
```diff
@@ -94,11 +94,12 @@ extern "C" {
...
-    MFU_TYPE_UNKNOWN = 1, /* type not known */
-    MFU_TYPE_FILE    = 2, /* regular file */
-    MFU_TYPE_DIR     = 3, /* directory */
-    MFU_TYPE_LINK    = 4, /* symlink */
+    MFU_TYPE_NULL     = 0, /* type not set */
+    MFU_TYPE_UNKNOWN  = 1, /* type not known */
+    MFU_TYPE_FILE     = 2, /* regular file */
+    MFU_TYPE_DIR      = 3, /* directory */
+    MFU_TYPE_LINK     = 4, /* symlink */
+    MFU_TYPE_HARDLINK = 5, /* hardlink */
```
ajmes, line 102: Setting a new type that does not reflect posix type can have side effect:
e.g:
```
$ touch test_tree0/file{0..9}
$ ln test_tree0/file0 test_tree0/file0_ln1
$ ln test_tree0/file0 test_tree0/file0_ln2
$ mpirun -np 8 --hostfile hosts.mpi dsync -S test_tree0/ test_tree1/
...
[2025-09-11T18:47:28]   Directories: 0
[2025-09-11T18:47:28]   Files: 10
[2025-09-11T18:47:28]   Links: 0
[2025-09-11T18:47:28]   Hardlinks: 2

# unlink a secondary ref (MFU_TYPE_HARDLINK)
# rm test_tree0/file0_ln2
$ mpirun -np 8 --hostfile hosts.mpi dcmp test_tree0/ test_tree1/
Number of items that exist in both directories: 12 (Src: 12 Dest: 12)
Number of items that exist only in one directory: N/A (Src: 0 Dest: 1)
Number of items that exist in both directories and have the same type: 12 (Src: 12 Dest: 12)
Number of items that exist in both directories and have different types: 0 (Src: 0 Dest: 0)
Number of items that exist in both directories and have the same content: 12 (Src: 12 Dest: 12)
Number of items that exist in both directories and have different contents: 0 (Src: 0 Dest: 0)

# unlink the primary ref (MFU_TYPE_FILE)
$ ln test_tree0/file0 test_tree0/file0_ln2
$ rm test_tree0/file0
$ mpirun -np 8 --hostfile hosts.mpi dcmp  --out EXIST=COMMON@TYPE=DIFFER:out --text test_tree0/ test_tree1/
Number of items that exist in both directories and have different types: 1 (Src: 1 Dest: 1), dumped to "out"
$ cat out
-rw-r--r-- eaujames eaujames   0.000   B Sep 11 2025 18:45 /tmp/test_tree0/file0_ln1
-rw-r--r-- eaujames eaujames   0.000   B Sep 11 2025 18:45 /tmp/test_tree1/file0_ln1
$ stat --printf "%n %i %F\n" test_tree*/*_ln*
test_tree0/file0_ln1 195619 regular empty file
test_tree0/file0_ln2 195619 regular empty file
test_tree1/file0_ln1 195758 regular empty file
test_tree1/file0_ln2 195758 regular empty file
```
Here we mark abitrary an entry as primary (MFU_TYPE_FILE) and the others as secondary (MFU_TYPE_HARDLINK). But that does not make sense for a Posix FS (all the entries point to the same inode), the type should be unchanged.

Can we save this information somewhere else? For example, we can use `elem_t::detail` to store detail flags.
e.g:
```
enum mfu_detail_e {
    MFU_DETAIL_NULL     = 0x0, /* detail not set */
    MFU_DETAIL_SET  = 0x1, /* element contains stat data */
    MFU_DETAIL_LINK = 0x2, /* element contains a valid link ref*/
};
```
This would enable to support hardlink for other types of inode (e.g: fifo).

ajmes (commented):
Thank you for the patch update. I’ve noticed some additional corner case issues

src/common/mfu_flist.c
rezib: We can definitely add:

```diff
diff --git a/src/common/mfu_flist_io.c b/src/common/mfu_flist_io.c
index 37039bf..13ebefb 100644
--- a/src/common/mfu_flist_io.c
+++ b/src/common/mfu_flist_io.c
@@ -356,7 +356,8 @@ static size_t list_elem_unpack(const void* buf, int detail, uint64_t chars, elem
         mfu_unpack_io_uint64(&ptr, &elem->size);
         mfu_unpack_io_uint64(&ptr, &elem->nlink);
         mfu_unpack_sized_str(&ptr, &elem->ref, chars);
-    }
+    } else
+        elem->ref = NULL;
 
     size_t bytes = (size_t)(ptr - start);
     return bytes;
```

But I don't see a code path leading to `list_elem_unpack()` being called with detail=0… It is called in `read_cache_v5()` → `list_insert_ptr(detail=1)` → `list_elem_unpack(detail=1)`. I can't find any other code path to this function.

How did you manage to produce this segfault with the tests? I'm not able to reproduce it :worried:

src/common/mfu_flist.h
rezib: > Setting a new type that does not reflect posix type can have side effect: e.g:
> 
> ```
> $ touch test_tree0/file{0..9}
> $ ln test_tree0/file0 test_tree0/file0_ln1
> $ ln test_tree0/file0 test_tree0/file0_ln2
> $ mpirun -np 8 --hostfile hosts.mpi dsync -S test_tree0/ test_tree1/
> ...
> [2025-09-11T18:47:28]   Directories: 0
> [2025-09-11T18:47:28]   Files: 10
> [2025-09-11T18:47:28]   Links: 0
> [2025-09-11T18:47:28]   Hardlinks: 2
> 
> # unlink a secondary ref (MFU_TYPE_HARDLINK)
> # rm test_tree0/file0_ln2
> $ mpirun -np 8 --hostfile hosts.mpi dcmp test_tree0/ test_tree1/
> Number of items that exist in both directories: 12 (Src: 12 Dest: 12)
> Number of items that exist only in one directory: N/A (Src: 0 Dest: 1)
> Number of items that exist in both directories and have the same type: 12 (Src: 12 Dest: 12)
> Number of items that exist in both directories and have different types: 0 (Src: 0 Dest: 0)
> Number of items that exist in both directories and have the same content: 12 (Src: 12 Dest: 12)
> Number of items that exist in both directories and have different contents: 0 (Src: 0 Dest: 0)
> ```

This output looks good to me so far, doesn't it?

> ```
> # unlink the primary ref (MFU_TYPE_FILE)
> $ ln test_tree0/file0 test_tree0/file0_ln2
> $ rm test_tree0/file0
> $ mpirun -np 8 --hostfile hosts.mpi dcmp  --out EXIST=COMMON@TYPE=DIFFER:out --text test_tree0/ test_tree1/
> Number of items that exist in both directories and have different types: 1 (Src: 1 Dest: 1), dumped to "out"
> $ cat out
> -rw-r--r-- eaujames eaujames   0.000   B Sep 11 2025 18:45 /tmp/test_tree0/file0_ln1
> -rw-r--r-- eaujames eaujames   0.000   B Sep 11 2025 18:45 /tmp/test_tree1/file0_ln1
> $ stat --printf "%n %i %F\n" test_tree*/*_ln*
> test_tree0/file0_ln1 195619 regular empty file
> test_tree0/file0_ln2 195619 regular empty file
> test_tree1/file0_ln1 195758 regular empty file
> test_tree1/file0_ln2 195758 regular empty file
> ```

Yes, with the current logic:

* `test_tree0/file0_ln1` is considered a `MFU_TYPE_FILE` in source while `test_tree1/file0_ln1` is considered `MFU_TYPE_HARDLINK` in destination → type difference
* `test_tree0/file0_ln2` has a reference to `test_tree0/file0_ln1` in source while `test_tree1/file0_ln2` has a reference to `test_tree1/file0` → content difference

Sure, this leads to false positive differences in this case but hey, how could we best handle this case with a distributed algorithm?

> Here we mark abitrary an entry as primary (MFU_TYPE_FILE) and the others as secondary (MFU_TYPE_HARDLINK). But that does not make sense for a Posix FS (all the entries point to the same inode), the type should be unchanged.
> 
> Can we save this information somewhere else? For example, we can use `elem_t::detail` to store detail flags. e.g:
> 
> ```
> enum mfu_detail_e {
>     MFU_DETAIL_NULL     = 0x0, /* detail not set */
>     MFU_DETAIL_SET  = 0x1, /* element contains stat data */
>     MFU_DETAIL_LINK = 0x2, /* element contains a valid link ref*/
> };
> ```
> 
> This would enable to support hardlink for other types of inode (e.g: fifo).

You're right in theory, except mpifileutils doesn't support fifo yet 😅 What's opinion of mpifileutils maintainers (@ofaaland, @carbonneau1) on this point?

src/common/mfu_flist.c
ajmes: Yes, I don't think this came from `read_cache_v5()`.
```python
    def test_walk_input_lite(self):                                            
        with tempfile.NamedTemporaryFile(mode="w+") as fh:                     
            cache = Path(fh.name)                                              
            self.run_dwalk(output=cache, lite=True)                            
            proc = self.run_dwalk(input=cache)                                 
            # lite cache do not contain hardlinks, then dwalk misses it.       
            self.assertInProcStdout(                                           
```
This test uses "--lite" option (without stat), so `detail` is set to 0 in the cache file:
```c
int mfu_flist_walk_paths(uint64_t num_paths, const char** paths,            
                          mfu_walk_opts_t* walk_opts, mfu_flist bflist,     
                          mfu_file_t* mfu_file)                             
{                                                                           
...
    flist->detail = 0;                             
    if (walk_opts->use_stat) {                     
```
Then we use `write_cache_readdir_variable` to write the cache file and `read_cache_variable()` to read it. `read_cache_variable` does not set ref: 
`read_cache_variable` ->`list_insert_decode()` -> `list_elem_decode()`.

So, I am more in favor of:
```diff
@@ -407,7 +407,8 @@ static void list_delete(flist_t* flist)          
     while (current != NULL) {                                       
         elem_t* next = current->next;                               
         mfu_free(&current->file);                                   
-        mfu_free(&current->ref);                                    
+       if (current->detail)                                         
+               mfu_free(&current->ref);                             
         mfu_free(&current);                                         
         current = next;                                             
     }                                                               
```

rezib: I just pushed:

* [530594f](https://github.com/hpc/mpifileutils/pull/633/commits/530594f5bb6a4244bc0d9828f49c413adce6ceee): Set ref to NULL w/o detail
* [a1bd23d](https://github.com/hpc/mpifileutils/pull/633/commits/a1bd23ddcd261cd944f6f198e1bc3a150fd75cd6): Skip free() w/o detail

src/dsync/dsync.c
```diff
@@ -1791,6 +1948,24 @@ static int dsync_strmap_compare(
...
+            const char* src_ref = mfu_flist_file_get_ref(src_list, src_index) + strlen_prefix;
+            const char* dst_ref = mfu_flist_file_get_ref(dst_list, dst_index) + strlen(dest_path->path);
+
+            if(strcmp(src_ref, dst_ref)) {
+                /* take them as differ content */
+                dsync_strmap_item_update(src_map, key, DCMPF_CONTENT, DCMPS_DIFFER);
+                dsync_strmap_item_update(dst_map, key, DCMPF_CONTENT, DCMPS_DIFFER);
+
+                if (!options.dry_run) {
+                    mfu_flist_file_copy(src_list, src_index, src_cp_list);
```
ajmes, line 1968: A segmentation fault happens in here.

ajmes (commented):
We hit another SEGFAULT in dsync_remove_hardlinks_with_removed_refs().

The crash occured at:
```c
1715                 dsync_strmap_item_update(src_map, key, DCMPF_CONTENT, DCMPS_DIFFER);
```

```c
static void dsync_strmap_item_update(
    strmap* map,
    const char *key,
    dsync_field field,
    dsync_state state)
{
    /* Should be long enough for 64 bit number and DCMPF_MAX */
    char new_val[21 + DCMPF_MAX];
    /* lookup item from map */
    const char* val = strmap_get(map, key);
    /* copy existing index over */
    assert(field < DCMPF_MAX);
assert(strlen(val) + 1 <= sizeof(new_val));
```
`strmap_get()`  returned a NULL pointer. Then `strlen(NULL)` crashed dsync.

I have applied the following hot-fix for our migrations:
```diff
diff --git a/src/dsync/dsync.c b/src/dsync/dsync.c
index a84aa92..7f33ae2 100644
--- a/src/dsync/dsync.c
+++ b/src/dsync/dsync.c
@@ -418,6 +418,8 @@ static void dsync_strmap_item_update(
 
     /* lookup item from map */
     const char* val = strmap_get(map, key);
+    if (!val)
+           return;
 
     /* copy existing index over */
     assert(field < DCMPF_MAX);
```

This occcured on a "trash" entry: we use a robinhood policy to automaticaly put file in a trash folder. This policy use "lfs fid2path" to find all the hardlinks in Lustre and then it executes "mv" for each entry. I think the crash can be caused by racy entry renames.

src/common/mfu_flist.h
rezib: > > ```
> > # unlink the primary ref (MFU_TYPE_FILE)
> > $ ln test_tree0/file0 test_tree0/file0_ln2
> > $ rm test_tree0/file0
> > $ mpirun -np 8 --hostfile hosts.mpi dcmp  --out EXIST=COMMON@TYPE=DIFFER:out --text test_tree0/ test_tree1/
> > Number of items that exist in both directories and have different types: 1 (Src: 1 Dest: 1), dumped to "out"
> > $ cat out
> > -rw-r--r-- eaujames eaujames   0.000   B Sep 11 2025 18:45 /tmp/test_tree0/file0_ln1
> > -rw-r--r-- eaujames eaujames   0.000   B Sep 11 2025 18:45 /tmp/test_tree1/file0_ln1
> > $ stat --printf "%n %i %F\n" test_tree*/*_ln*
> > test_tree0/file0_ln1 195619 regular empty file
> > test_tree0/file0_ln2 195619 regular empty file
> > test_tree1/file0_ln1 195758 regular empty file
> > test_tree1/file0_ln2 195758 regular empty file
> > ```
> 
> Yes, with the current logic:
> 
>     * `test_tree0/file0_ln1` is considered a `MFU_TYPE_FILE` in source while `test_tree1/file0_ln1` is considered `MFU_TYPE_HARDLINK` in destination → type difference
> 
>     * `test_tree0/file0_ln2` has a reference to `test_tree0/file0_ln1` in source while `test_tree1/file0_ln2` has a reference to `test_tree1/file0` → content difference
> 
> 
> Sure, this leads to false positive differences in this case but hey, how could we best handle this case with a distributed algorithm?

FYI, I added in a test in [ca9896f](https://github.com/hpc/mpifileutils/pull/633/commits/ca9896f6e787d44451014a401a342b55e27cf4b3) to document the behavior in this case.

rezib:
> We hit another SEGFAULT in dsync_remove_hardlinks_with_removed_refs().
> 
> The crash occured at:
> 
> ```c
> 1715                 dsync_strmap_item_update(src_map, key, DCMPF_CONTENT, DCMPS_DIFFER);
> ```
> […]

Thanks @ajmes for reporting it!

Indeed, this segfault is triggered when a reference is different between source and destination and a new hardlink is present on this reference in destination. I was able to reproduce it with tests introduced in [90ff407](https://github.com/hpc/mpifileutils/pull/633/commits/90ff407bd9e877861b47cd7f1256636d50748eff).

This is fixed in [d9d0cde](https://github.com/hpc/mpifileutils/pull/633/commits/d9d0cde60492353b93c1c2ab934e28893339b4a9).

PR #642, opened 2025-05-12
dcp: avoid truncate when not needed

mchaarawi:
ftruncate is expensive on many file systems. Avoid calling truncate in unecessary situation like when using O_DIRECT but the file size is what is expected, or when not using O_DIRECT. The logic is implemented as follows with dcp:
create file
if EEXIST:
    # only need to truncate if it existed
    stat file
    if sparse or file size > bytes we will write:
    # ... and if using sparse or the size is larger than what the new size will be
        truncate file to 0
write data
if O_DIRECT:
    # only need to truncate after if O_DIRECT and the file size is greater than what we want to truncate to
    stat file
    if file size > truncate offset:
        truncate file to truncate offset

### DIFF
--- a/src/common/mfu_flist_copy.c
+++ b/src/common/mfu_flist_copy.c
@@ -1242,6 +1242,7 @@ static int mfu_create_file(
      * for regular files, dev argument is supposed to be ignored,
      * see makedev() to create valid dev */
     dev_t dev;
+    bool file_exists = false;
     memset(&dev, 0, sizeof(dev_t));
     int mknod_rc = mfu_file_mknod(dest_path, DCOPY_DEF_PERMS_FILE | S_IFREG, dev, mfu_dst_file);
     if(mknod_rc < 0) {
@@ -1249,6 +1250,7 @@ static int mfu_create_file(
             /* destination already exists, no big deal, but print warning */
             MFU_LOG(MFU_LOG_WARN, "Original file exists, skip the creation: `%s' (errno=%d %s)",
                     dest_path, errno, strerror(errno));
+	    file_exists = true;
         } else {
             /* failed to create inode, that's a problem */
             MFU_LOG(MFU_LOG_ERR, "File `%s' mknod() failed (errno=%d %s)",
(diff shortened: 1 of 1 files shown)


src/common/mfu_flist_copy.c
```diff
@@ -1847,18 +1864,30 @@ static int mfu_copy_file_normal(
     }
 #endif
 
-    /* if we wrote the last chunk, truncate the file */
+    /* if sparse or using O_DIRECT, we already truncated the file at the beginning, we do not need
+     * to do it again. */
+    if (copy_opts->sparse || !copy_opts->direct)
+	    return 0;
```
daltonbohning, line 1870: What if using sparse and O_DIRECT? Should this instead be?
```
    if (!copy_opts->direct)
	    return 0;
```

src/common/mfu_flist_copy.c
```diff
@@ -1847,18 +1864,30 @@ static int mfu_copy_file_normal(
...
+	 * of Lustre would cause block to truncate() since the fd is different
+	 * from the out_fd. */
+	struct stat st;
+        int status = mfu_file_lstat(dest, &st, mfu_dst_file);
+	if (status == 0 && st.st_size > file_size_offt) {
+            if (mfu_file_ftruncate(mfu_dst_file, file_size_offt) < 0) {
+	        MFU_LOG(MFU_LOG_ERR, "Failed to truncate destination file: %s (errno=%d %s)",
+			dest, errno, strerror(errno));
+		return -1;
+	    }
```
daltonbohning, line 1886: I'm not 100% confident, so don't want to suggest it, but I think we _**might**_ already know whether we wrote past the end because of
https://github.com/hpc/mpifileutils/blob/92ca762531f8136e3008708b3539f9b377a46256/src/common/mfu_flist_copy.c#L1795-L1811

src/common/mfu_flist_copy.c
mchaarawi: yes indeed.. not sure why i had sparse there

src/common/mfu_flist_copy.c
mchaarawi: are you suggesting to add something like:
if (remainder > 0) { need_truncate = true} and we avoid doing the stat at the end?

daltonbohning: Yeah, and I also see now that we DO need to truncate for sparse if we skipped writing the last chunk. Since this relies on the truncate to _**extend**_ the file size
https://github.com/hpc/mpifileutils/blob/92ca762531f8136e3008708b3539f9b377a46256/src/common/mfu_flist_copy.c#L1813-L1820

mchaarawi: i updated  the PR with this

daltonbohning (approved):
LGTM. I'll do some manual testing to sanity check this

daltonbohning:
I tested with --direct and files that are aligned and not aligned, and testing with --sparse. I didn't find any issues with the code paths or file sizes.

carbonneau1 (approved):

carbonneau1:
@mchaarawi Could you sign your commit and resubmit please so we can merge.

thanks

-Eric

mchaarawi:
done.

PR #648, opened 2025-07-08
dsync: use SEEK_DATA for sparse detection if possible

ofaaland:
In the past, when flag --sparse is given, the fiemap ioctl is used to detect holes and data extents.

lseek() with SEEK_HOLE and SEEK_DATA works on Lustre 2.15.7 when fiemap does not (e.g. files striped via PFL), and works on tmpfs which appears not support fiemap, at least on RHEL 8.

Tools in wider use, such as GNU Fileutils and tar, prefer lseek() over fiemap.

Since MFU doesn't use the other information returned by fiemap anyway, and use SEEK_HOLE and SEEK_DATA first, and fall back on fiemap only if that fails.

### DIFF
--- a/src/common/mfu_flist_copy.c
+++ b/src/common/mfu_flist_copy.c
@@ -1888,7 +1888,45 @@ static int mfu_copy_file_normal(
     return 0;
 }
 
-static int mfu_copy_file_fiemap(
+struct mfu_extent {
+    __u64 me_logical;  /* logical offset in bytes for the start of the extent */
+    __u64 me_length;   /* length in bytes for this extent */
+};
+
+struct mfu_extent_list {
+    __u32 mel_mapped_extents;/* number of extents that were mapped */
+    __u32 mel_extent_count;  /* size of fm_extents array */
+    struct mfu_extent mel_extents[0]; /* array of mapped extents */
+};
+
+/*
+ * If we are resizing an existing allocation, the new extent_count must be
+ * larger than the old one.
+ */
+struct mfu_extent_list *mfu_extent_list_realloc(struct mfu_extent_list *orig, size_t extent_count)
+{
+    size_t extent_list_size = sizeof(struct mfu_extent_list) + sizeof(struct mfu_extent) * extent_count;
+    size_t old_size = 0;
+
+    if (orig) {
+        if (extent_count <= orig->mel_extent_count) {
+            return NULL;
+        }
+
+        old_size = sizeof(struct mfu_extent_list) + sizeof(struct mfu_extent) * orig->mel_extent_count;
+    }
+
+    struct mfu_extent_list* extent_list = (struct mfu_extent_list*) realloc(orig, extent_list_size);
+    if (extent_list == NULL) {
+        return NULL;
+    }
+
+    memset((char *)extent_list + old_size, 0, extent_list_size - old_size);
+    extent_list->mel_extent_count = extent_count;
+    return extent_list;
+}
+
+static struct mfu_extent_list * mfu_fiemap_get_extents(
     const char* src,
     const char* dest,
     uint64_t offset,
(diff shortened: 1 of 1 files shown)


src/common/mfu_flist_copy.c
```diff
@@ -1981,8 +1999,150 @@ static int mfu_copy_file_fiemap(
...
+    struct mfu_extent_list* extent_list = mfu_extent_list_realloc(NULL, extent_count);
+    if (extent_list == NULL) {
+        MFU_LOG(MFU_LOG_ERR, "Out of memory allocating extent_list with count %d", extent_count);
+        goto fail_no_extent_list;
+    }
+
+    op = SEEK_DATA;
+    idx = 0;
+
+    MFU_LOG(MFU_LOG_INFO, "map from lseek():");
```
daltonbohning, line 2050: Is this meant to be removed?

ofaaland: Oops, I meant for that to be MFU_LOG_DBG.  Thanks

ofaaland:
> I think the preference upstream (GNU Fileutils, tar, etc.) is to prefer SEEK_HOLE, SEEK_DATA over FIEMAP, unless MFU can use some of the extra information from the FIEMAP data vs. just knowing where the data is?

That's good to know, thank you.  At the moment, MFU doesn't use any additional information from FIEMAP.  I'll change the order.

src/common/mfu_flist_copy.c
```diff
@@ -1981,8 +1999,150 @@ static int mfu_copy_file_fiemap(
...
+    op = SEEK_DATA;
+    idx = 0;
+
+    MFU_LOG(MFU_LOG_DBG, "map from lseek():");
+    start = offset;
+    while (offset < last_byte) {
+        offset = lseek(mfu_src_file->fd, offset, op);
+
+        if (op == SEEK_HOLE) {
+            MFU_LOG(MFU_LOG_DBG, "%llu - %llu type %s\n", start, offset, (op == SEEK_HOLE ? "data" : "hole"));
```
defaziogiancarlo, line 2056: the ternary operator condition is always true here, "op == SEEK_HOLE" is always true inside this if block.

src/common/mfu_flist_copy.c
```diff
@@ -1888,7 +1888,37 @@ static int mfu_copy_file_normal(
...
+    if (orig) {
+        old_size = sizeof(struct mfu_extent_list) + sizeof(struct mfu_extent) * orig->mel_extent_count;
+    }
+
+    struct mfu_extent_list* extent_list = (struct mfu_extent_list*) realloc(orig, extent_list_size);
+    if (extent_list == NULL) {
+        return NULL;
+    }
+
+    memset((char *)extent_list + old_size, 0, extent_list_size - old_size);
```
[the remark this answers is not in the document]
ofaaland: Good catch.  I'll check for that.

src/common/mfu_flist_copy.c
ofaaland: I'll just remove that message.

src/common/mfu_flist_copy.c
ofaaland: Fixed in the update I just pushed

src/common/mfu_flist_copy.c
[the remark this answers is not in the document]
ofaaland: Fixed in the update I just pushed

defaziogiancarlo (approved):

ofaaland:
There are wrappers for lseek, such as `mfu_file_lseek()` which calls `daos_lseek()` for daos file systems.  I'm not using those, so I need to refactor.

Switching back to draft temporarily.

ofaaland:
Switched to use mfu_file_lseek() so the right variant of lseek is called for source files on DAOS, and so we get retries for certain conditions.

Fixed a bug which was triggered when a file was mostly sparse and multiple threads were used, which resulted in no data being copied.

I think it's now ready.

src/common/mfu_flist_copy.c
```diff
@@ -1981,8 +2007,184 @@ static int mfu_copy_file_fiemap(
...
+    *normal_copy_required = true;
+    uint64_t logical_off = offset;
+    uint64_t last_byte = offset + length;
+    int op;
+    int idx;
+    uint64_t start;
+
+    /* Verify that SEEK_HOLE and SEEK_DATA are supported */
+    if ((mfu_file_lseek(src, mfu_src_file, offset, SEEK_HOLE) == (off_t)-1) ||
+        (mfu_file_lseek(src, mfu_src_file, offset, SEEK_DATA) == (off_t)-1)) {
```
daltonbohning, line 2055: Thanks for using these wrappers. FWIW, there is no support for this yet in daos

daltonbohning (approved):
I'm not familiar with these sparse functions but conceptually LGTM

defaziogiancarlo (approved):

PR #649, opened 2025-07-15
Add a --print0 option to print null terminated strings

jprorama:
This option is useful for working with files that have special characters like newlines embedded in them.  It makes it it possible to pass this output to other downstream commands that can process null terminated strings, e.g. "parallel --null ..."

Added option string, new function MFU_PRED_PRINT0() and argument parsing to select this feature.


jbd:
This would be a nice addition. I hacked something similar, but more intrusive than this patch which is much cleaner. 

I actually need to use dwalk and dfind to find sockets and pipes files and use rsync to sync them. Using rsync from0 is less error prone considering you can have path with weird characters in them, like a newline.

PR #681, opened 2026-07-23
ddup: remove openssl dependency (#535)

### LINKED ISSUE #535: ddup hashing and openssl 3.x
Was looking through the code to see if there was a less computationally expensive way to handle duplicate files (SHA256 is unnecessary IMO, but that's a separate issue/PR) and as it turns out openssl 3.x drops support for the low level interface to hash functions.

Considering openssl 1.1.1 is EOL in under a year, this may be worth looking at. From what I'm reading, the correct functions are in evp.h per https://stackoverflow.com/questions/69806220/advice-needed-for-migration-of-low-level-openssl-api-to-high-level-openssl-apis

jeking3:
The `ddup` command uses SHA256 to identify files with identical content and prints each duplicate's digest to stdout. Because that digest is user-visible output, changing the hash algorithm could be a breaking change. Instead, we leverage a permissively-licensed (public-domain), portable SHA256 implementation to sever the dependency on openssl. This lets us build on newer operating systems that have dropped the openssl 1.1 low-level SHA API.  This builds under ubuntu 26.04 inside a container, for example.

The SHA256 implementation is from Brad Conte's public domain solution and a known-answer regression test was added.

In the future, if a higher-performing content comparison is desired, it can be added via a new command-line option and/or by embedding the hash choice into any future output or cache format, defaulting to SHA256. That default could change in a later major release.

This fixes #535.


samuelkgutierrez (approved):
Looks good to me.

Note that I'm moving this forward because it seems that this project isn't being actively maintained.

PR #683, opened 2026-09-09
Use the fetch io map for EC extents in DAOS object copy and serialize

mchaarawi:
daos_obj_list_recx() returns a superset of the data for EC objects: the extents are derived from the parity and are stripe granular. dcp, dsync and daos-serialize copied whole listed extents, so punched and unwritten regions were written out as zeros, or the copy failed outright on the sg_nr_out sanity check. Drive the copy from the fetch io map instead.

Source reads also ran at DAOS_TX_NONE while the object list came from a snapshot, so enumeration and fetch could land on different epochs. Open a snapshot transaction per rank and use it for all source reads. The epoch is now broadcast, since only rank 0 took the snapshot.

Also fix the 8x over-allocation in alloc_iov_buf, add overflow checks on the extent/record-size products, and free the serialize buffer on error.


src/common/mfu_daos.c
```diff
@@ -2025,143 +1988,246 @@ static int mfu_daos_obj_sync_recx_array(
...
+            uint64_t     off = 0;
+            uint64_t     map_len = 0;
+            uint64_t     map_end = 0;
+            uint32_t     k;
+
+            for (k = 0; k < number; k++) {
+                /* subtract rather than add, rx_idx + rx_nr can wrap */
+                if (map->rx_idx >= recxs[k].rx_idx &&
+                    map->rx_idx - recxs[k].rx_idx < recxs[k].rx_nr) {
+                    off += (map->rx_idx - recxs[k].rx_idx) * size;
```
daltonbohning, line 2124: But we don't need to worry about the multiplication here overflowing?

src/common/mfu_daos.c
```diff
@@ -2025,143 +1988,246 @@ static int mfu_daos_obj_sync_recx_array(
...
+                mfu_free(&dst_buf);
+                dst_buf = (char*) MFU_CALLOC(map_bytes, 1);
+                dst_buf_alloc = map_bytes;
             }
 
-            rc = daos_obj_fetch(*dst_oh, DAOS_TX_NONE, 0, dkey, 1, iod,
-                                &dst_sgl, NULL, NULL);
+            /* the fetch leaves holes in the dst untouched, so they have to read as zeros
+             * rather than as whatever the previous extent left behind */
+            memset(dst_buf, 0, map_bytes);
```
daltonbohning: Maybe it's negligible, but the memset seems not necessary if we just did a free + calloc above. So maybe `... else memset`?

src/common/mfu_daos.c
```diff
@@ -2531,10 +2602,27 @@ int mfu_daos_flist_sync(
...
+        /* copy object ids listed in newflist to destination in daos args */
+        rc = mfu_daos_flist_obj_sync(da, newflist, da->src_coh, da->dst_coh, th,
                                      compare_dst, write_dst, &stats);
 
+        int tmp_rc = daos_tx_close(th, NULL);
+        if (tmp_rc != 0) {
+            MFU_LOG(MFU_LOG_ERR, "DAOS failed to close snapshot transaction "DF_RC, DP_RC(tmp_rc));
+            if (rc != -1) {
+                rc = -1;
+            }
```
daltonbohning: Simplify?
```suggestion
            rc = -1;
```

daltonbohning (approved):

src/common/mfu_daos.c
mchaarawi: bounded by the rec_nr * size above

src/common/mfu_daos.c
mchaarawi: sure

mchaarawi:
sorry for the force push, but just changed the 3 places you pointed out

daltonbohning:
> sorry for the force push, but just changed the 3 places you pointed out

Huh. GitHub actually handled that well. I guess it saves a reference to the previous commit now
https://github.com/hpc/mpifileutils/compare/352686ac75e6ab215856ba6b9e834e2280a7fea9..2a249402f26110e48ebea786b433779d73a5cdb9

daltonbohning (approved):

